PR #3, opened 2017-09-13
Added the paradox documentation

bogdanromanx:


README.md
```diff
@@ -33,67 +33,67 @@ Now let's take a look at these components individually.
 ### Nexus KnowledgeGraph
 This service is the heart of the Blue Brain Nexus platform. It allows users to define their domain, populate the knowledge graph with data, attach data files to entities. It also provides semantic search facilities to discover similar and relevant data in the platform.
 
-**Status**: active development
-**Source code**: https://github.com/BlueBrain/nexus-kg
+**Status**: active development  
+**Source code**: https://github.com/BlueBrain/nexus-kg
```
wwajerowicz: Is there a reason this whitespace was added here?

bogdanromanx: It was intended, to get a line break.

bogdanromanx:
Good catch, however I wouldn't add build instructions to the project readme... I would very much prefer to fix the issue (by publishing the artifacts to a public repo).

src/main/paradox/setup/resources/index-domains.properties
```diff
@@ -0,0 +1,11 @@
+com.bigdata.rdf.sail.namespace=kg-domains
+com.bigdata.rdf.sail.isolatableIndices=false
+com.bigdata.rdf.store.AbstractTripleStore.textIndex=true
+com.bigdata.rdf.store.AbstractTripleStore.quads=true
+com.bigdata.rdf.sail.truthMaintenance=false
+com.bigdata.namespace.test-ns.lex.com.bigdata.btree.BTree.branchingFactor=400
+com.bigdata.rdf.store.AbstractTripleStore.justify=false
+com.bigdata.rdf.store.AbstractTripleStore.geoSpatial=false
+com.bigdata.namespace.test-ns.spo.com.bigdata.btree.BTree.branchingFactor=1024
```
samuel-kerrien: test-ns ? is that something we want to put out int he default config ?

src/main/paradox/setup/resources/index-domains.properties
```diff
@@ -0,0 +1,11 @@
+com.bigdata.rdf.sail.namespace=kg-domains
+com.bigdata.rdf.sail.isolatableIndices=false
+com.bigdata.rdf.store.AbstractTripleStore.textIndex=true
+com.bigdata.rdf.store.AbstractTripleStore.quads=true
+com.bigdata.rdf.sail.truthMaintenance=false
+com.bigdata.namespace.test-ns.lex.com.bigdata.btree.BTree.branchingFactor=400
```
samuel-kerrien: test-ns ?

src/main/paradox/setup/resources/index-domains.properties
MFSY: For now, I guess these namespace configs need to be (manually) applied  prior to start Nexus KG service in order to benefit from indexing and search through blazegraph. 

But the right approach would be to read and applied (user provided or default) namespace configs automatically at boot time.

bogdanromanx: Very nice catch; I've just used the index definitions that were present in the docs. I'll regenerate the indices.

bogdanromanx: Corrected the issue.

src/main/paradox/setup/resources/index-domains.properties
bogdanromanx: Corrected the issue.

bogdanromanx:
Is this GTM?

wwajerowicz:
looks fine to me

PR #43, opened 2018-05-04
Added jupyter notebook tutorial for basic operations with nexus v0

annakristinkaufmann:


pafonta:
It would be better to use the URL pointing to the rendered version instead of one to the raw file:
https://github.com/BlueBrain/nexus/blob/master/tutorial/basic_operations_nexus_v0.ipynb. People can thus read the notebook right away.

PR #44, opened 2018-05-07
Add github pages static content folder

kenjinp:
This will create a folder called /docs. In this folder, static content can be served from the project's github pages domain.  We can use that github pages domain as a landing / marketing page for the nexus project.


mdupont:
I would love to have the sources also committed publicly, and building into docs/. Here we only have the compiled/minified version.

kenjinp:
As discussed with @bogdanromanx we'll use the `/docs` folder to host the source code and push the built dist files to `gh-pages` branch for deployment.

PR #46, opened 2018-05-24
remove email catches

kenjinp:
This removes the email catches and fixes a few UI bugs
- fix broken github link at top right
- adds a new way for scrolling text
- adds css vendor prefixes to fix some minor style bugs
- removes all instances of the email catches


kenjinp:
It's commented out, I'll upload the src to `nexus-gh-pages` branch under docs so you can confirm

bogdanromanx:
I couldn't find any `<!--` or `--> in the html source, that's why I assumed the form is hidden from JS.

kenjinp:
Here's the new [source](https://github.com/BlueBrain/nexus/tree/nexus-github-pages).

They're removed in the minification process but you're right the modal form is still there, just invisible.

PR #71, opened 2018-09-05
Add docker-compose template for 32 GB of RAM

hygt:
* Add docker-compose template for 32 GB of RAM

* Use 3rd party Docker image for Blazegraph

### DIFF
--- a/src/main/paradox/running_locally/docker/docker-compose-32GB.yaml
+++ b/src/main/paradox/running_locally/docker/docker-compose-32GB.yaml
@@ -2,10 +2,6 @@ version: "3.3"
 services:
   kg:
     image: "bluebrain/kg-nexus:latest"
-    deploy:
-      resources:
-        limits:
-          memory: 6G
     environment:
       - "CASSANDRA_CONTACT_POINT1=cassandra:9042"
       - "BIND_INTERFACE=0.0.0.0"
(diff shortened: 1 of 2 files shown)


src/main/paradox/running_locally/docker/docker-compose-32GB.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - "JAVA_OPTS=-Xms1G -Xmx3G"
+
+  elasticsearch:
+    image: "elasticsearch:latest"
+    deploy:
+      resources:
+        limits:
+          memory: 4G
+    environment:
+      - "ES_JAVA_OPTS=-Xms2G -Xmx3G"
```
bogdanromanx: I think ES complains if the min and max heap sizes don't match

src/main/paradox/running_locally/docker/docker-compose-32GB.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - "SPARQL_BASE_URI=http://blazegraph:9999/blazegraph"
+      - "SPARQL_ENDPOINT=http://blazegraph:9999/blazegraph/sparql"
+      - "ELASTIC_BASE_URI=http://elasticsearch:9200"
+      - "JAVA_OPTS=-Xms2G -Xmx4G"
+
+  iam:
+    image: "bluebrain/iam-nexus:latest"
+    deploy:
+      resources:
+        limits:
```
bogdanromanx: could we remove the limits across the entire file to avoid cases where the process gets killed because of off-heap memory usage? limiting the heap should suffice

PR #73, opened 2018-09-26
Add link to comprehensive example as tutorial

pafonta:
Hi,

I suggest to add links to the IPython notebook I created for the Literature Annotation domain and to the **productivity toolkit** I designed to support this **common workflow** of users of Nexus wanting to publish their data in a Knowledge Graph backed by it.

I think it would be of great help for users to see **all in one place**:
- the SHACL schemas (Neuroshapes),
- the original data (NeuroCurator / NAT),
- the transformed data (samples in the notebook),
- the transformation logic to make source data correspond to the schemas,
- the general workflow to follow to achieve **from 0 to Nexus**,
- the common issues a user have to face and how to solve them with the productivity toolkit.

To the best of my knowledge, there is indeed no **public resource** achieving this **unique view** on all these axes yet. Adding this reference in the 'Tutorials' section will **guide new users** and will make **existing users more productive**. It will also **decrease the need of support** for common mistakes and operations.


julienmachon:
This PR is now out of date.

PR #372, opened 2019-01-24
Nexus v1 product page update

kenjinp:
# Nexus v1 product page update

kapture 2019-01-24 at 19 09 54

Add all output assets compiled to the base /src/main/paradox main path, exactly as we have done in the [neuroshapes repo](https://github.com/INCF/neuroshapes/tree/gh-pages). 

## New stuff
- links to documentation at header
- introducing nexus v1 link
- link to tutorial 
- link to python sdk Doc
- link to js sdk Repo
- re-enabled basic email catch service using google docs
- style updates

The index.html references its assets as `./', so the assets should be good-to-go. 

I'm not sure how we previously copied the assets there, I suspect we might have done it by hand. I did not remove the previous assets.  We should reference the neuroshapes repo when we have time in the future as an example of how to re-implement the build process with gh-pages and node.


mdupont:
Replaced by #374

PR #436, opened 2019-02-08
Feedback Quick Start page

clifle:
OVERVIEW :
    Create a clear separation in between BBN aims / properties, 
    and the example of MoviLens dataset afterwards (under the voice 'Example').
    In Ingest data section as well. 
    
    LIST DATA: 
"The full payload of the resources are not retrieved when listing them: only identifier, type as well as Nexus added metadata are. But the result list can be scrolled and each resource fetched by identifier." ?
        JSON-LD link goes back to the beginning of the page

    QUERY DATA:
in the tab = ElasticSearch is linked to the beginning of the page

    https://nexus-sandbox.io/v1/views/tutorialnexus/$PROJECTLABEL/documents/_search and
    https://nexus-sandbox.io/v1/views/tutorialnexus/$PROJECTLABEL/graph/sparql
    not working:
    {"@context":"https://bluebrain.github.io/nexus/contexts/error.json","@type":"AuthorizationFailed","reason":"The supplied authentication is not authorized to access this resource."}


src/main/paradox/docs/tutorial/getting-started/quick-start/index.md
```diff
@@ -27,7 +27,7 @@ Let's get started.
 
 ### Install and setup the Nexus CLI
 
-Since the CLI is written in python, you may want to create a virtual env for a clean set up. To do so, Conda can be used. If you don't have it installed follow the instructions [here](https://conda.io/projects/conda/en/latest/user-guide/install/index.html).
+Since the CLI is written in python, you may want to create a virtualenv for a clean set up. To do so, Conda can be used. If you don't have it installed follow the instructions [here](https://conda.io/projects/conda/en/latest/user-guide/install/index.html).
```
MFSY: I'll use "virtual environment" instead as virtualenv is a specific tool.

MFSY: Fixed

clifle:
some more additional comments for a basic user friendly system :)

in LOGIN SECTION:
add the command for create the variable TOKEN 
export TOKEN=[copy here your token]

***load datasets takes around 40 mins***

with the command 
nexus resources fetch https://nexus-sandbox.io/v1/resources/tutorialnexus/$PROJECTLABEL/_/Movie_1
i always crash into 404 error

Should be mentioned how to connect ElasticSearchView\SparqlView to Nexus?

PR #437, opened 2019-02-08
thinking-in-graph images

clifle:
About the images, i can improve the resolution of the sections 'It is semantic' and 'It is alive' : 
make higher resolution of the pics or create original ones. **let me know**


bogdanromanx:
Sounds good (about the pics ;))

PR #600, opened 2019-05-01
update inferred relation image

clifle:


src/main/paradox/docs/tutorial/knowledge-graph/thinking-in-graph.md
```diff
@@ -38,12 +38,12 @@ and it has a property **Release date** for this the value is **March 29th, 1995*
  Since all this information is represented in an ontology, it enables us to run a tool called an inference engine on
 the data and derive new information from explicitly asserted data. This inference can make information easier to
 discover.
-
+git push
```
umbreak: That's probably not the place where you wanted to write this :)

PR #606, opened 2019-05-08
Update faq.md

clifle:
updated links in the updated path


umbreak:
I've made the changes myself against the @clifle branch. Does it look good to you now, @hygt ?

PR #630, opened 2019-05-16
docu impro avoiding repetitions

clifle:
Getting Started -> Getting Started Guide
for differentiate from tutorial one

Kg section/Toy Story image: was in both 'Understanding' and "Querying' pages


clifle:
> Maybe we can call it `Quick Start guide` instead? It is the way it is called from this page: https://bluebrainnexus.io/docs/tutorial/index.html

We could have 'Getting started' at the beginning,
and then
'Quick start guide' containing 'Quick Start' and 'Movielens'
otherwise

PR #669, opened 2019-06-17
Added resources, permissions, realms, orgs and projects SSE

umbreak:


src/main/paradox/docs/api/1.1/admin/admin-orgs-api.md
```diff
@@ -150,4 +150,37 @@ Request
...
\ No newline at end of file
+:   @@snip [organization-list.json](../assets/organization-list.json)
+
+
+## Organization Server Sent Events
+
+This endpoint allows clients to receive automatic updates from the organizations in a streaming fashion.
+
+```
+GET /v1/orgs/events?Last-Event-Id={last}
```
bogdanromanx: Last-Event-Id is passed as a header, not a query param.

src/main/paradox/docs/api/1.1/admin/admin-orgs-api.md
```diff
@@ -150,4 +150,37 @@ Request
...
+data:{payload}
+event:{type}
+id:{id}
+```
+
+where...
+
+- `{payload}`: Json - is the actual payload of the current organization
+- `{type}`: String - is a type identifier for the current organization. Possible types are: OrganizationCreated, OrganizationUpdated and OrganizationDeprecated
+- `{id}`: String - is the identifier of the organization event. It can be used in the `Last-Event-Id` query parameter
```
bogdanromanx: header

PR #1024, opened 2020-02-06
Small Fixes on views page

umbreak:


src/main/paradox/docs/api/current/kg/views/sparql-view-api.md
```diff
@@ -104,7 +105,7 @@ Request
 Response
 :   @@snip [view-fetched.json](../../assets/views/view-statistics.json)
 
-where:
+where...
```
wwajerowicz: I think `:` makes more sense here than `...`.

umbreak: Yeah, we can change then everywhere, cause on the other places we have 3 dots

PR #1056, opened 2020-02-27
Move docs to docs/ submodule

wwajerowicz:


.github/workflows/ci-docs.yml
```diff
@@ -1,8 +1,10 @@
-name: Continuous Integration
+name: Continuous Integration Docs
 on:
   push:
     branches:
       - master
+    paths:
+      - 'modules/docs/**'
```
bogdanromanx: this is wrong

PR #1064, opened 2020-03-02
Added clients implementation

umbreak:


cli/shared/src/test/scala/ch/epfl/bluebrain/nexus/cli/SparqlClientSpec.scala
```diff
@@ -1,6 +1,85 @@
...
 
-class SparqlClientSpec extends AnyWordSpecLike with Matchers {}
+class SparqlClientSpec extends AnyWordSpecLike with Matchers with Fixtures with OptionValues {
+
+  private val organization  = Label("myorg")
+  private val project       = Label("mylabel")
+  private val query         = "SELECT * {?s ?p ?o} LIMIT 10"
+  private val randomViewUri = nxv / genString()
+
+  private val mockedHttpApp = HttpApp[IO] {
```
wwajerowicz: maybe using `mockito` would be a bit cleaner here?

cli/shared/src/main/scala/ch/epfl/bluebrain/nexus/cli/types/EventEnvelope.scala
```diff
@@ -1,6 +1,6 @@
 package ch.epfl.bluebrain.nexus.cli.types
 
 /**
- * Event wrapper adding the offset metadata.
+  * Event wrapper adding the offset metadata.
  **/
-final case class EventEnvelope(offset: Offset, event: Event)
+final case class EventEnvelope(offsetOpt: Option[Offset], event: Event)
```
wwajerowicz: is there a case where an event wouldn't have an offset?

cli/shared/src/test/scala/ch/epfl/bluebrain/nexus/cli/SparqlClientSpec.scala
umbreak: Yes, maybe. I tried to do it without mockito from what we discussed lately. But this case is simple enough to be able to do it in both ways.

Let's see what @bogdanromanx thinks.

cli/shared/src/main/scala/ch/epfl/bluebrain/nexus/cli/types/EventEnvelope.scala
umbreak: In our case probably not, but the standard, the API and the `ServerSent` model of `fs2` all allow for this field to be optional.

cli/shared/src/test/scala/ch/epfl/bluebrain/nexus/cli/SparqlClientSpec.scala
bogdanromanx: I very much prefer this implementation over a mockito impl.

cli/shared/src/main/scala/ch/epfl/bluebrain/nexus/cli/types/EventEnvelope.scala
bogdanromanx: I'm happy to merge this as is, but we should be able to rely on having an offset. And fail on the first event that doesn't have one. It's part of the api.

PR #1076, opened 2020-03-06
Added Live and Test classes to Clients

umbreak:


cli/shared/src/main/scala/ch/epfl/bluebrain/nexus/cli/ClientError.scala
```diff
@@ -13,6 +15,31 @@ sealed trait ClientError extends Product with Serializable {
...
+
+  /**
+    * Construct a [[ClientError]] from the passed HTTP status code and message
+    *
+    * @throws IllegalArgumentException if the code is a Successful HTTP Status code is 2xx
+    */
+  def unsafe(code: Status, message: String): ClientError =
+    code.responseClass match {
+      case Status.Successful =>
+        "".toBoolean
```
bogdanromanx: ?

umbreak: Hahaha I don't know what that was doing there. Removed

PR #1079, opened 2020-03-09
Added support for read/write a Config

umbreak:
- Support to write a configuration to disk
- Support to read a configuration from disk
- Added a way to fetch the current offset from a `EventStream`. Once having this, the saving of that value in NexusConfig is up to the client: It could be done using.

If the client would want an offset to be stored every X minutes, one could do:
```scala
      val eventStream: EventStream[IO] = ??? // an EventStream
      fs2.Stream.awakeEvery[IO](X.minutes).evalMap{ _ =>
        eventStream.currentEventId().flatMap(offset => OffsetConfig(Some(offset)).write())
      }.compile.last.unsafeRunSync()
```

Or maybe just want to store the `NexusConfig` prior to shutdown:
```scala
scala.sys.addShutdownHook {
  eventStream.currentEventId().flatMap(offset => OffsetConfig(Some(offset)).write()).unsafeRunSync() 
  actorSystem.terminate()
  Await.result(actorSystem.whenTerminated, 30 seconds)
}
```

### DIFF
--- a/cli/shared/src/main/resources/reference.conf
+++ b/cli/shared/src/main/resources/reference.conf
@@ -20,4 +20,9 @@ app {
       condition = "on-server-error"
     }
   }
+  # the SSE configuration
+  sse {
+    # the optionally latest consumed lastEventId (a Sequence or a TimeBased UUID)
+    # last-event-id = "bd62f6d0-5c75-11ea-beb1-a5eb66b44d1c"
+  }
 }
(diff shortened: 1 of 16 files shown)


wwajerowicz:
Maybe it would be better to store the offset in a separate file? For me, the offset is not really a config value, it's more the state of the program, which I don't think should belong in the config file. The path to the offset file could be in the config.

bogdanromanx:
That's a very good point. It would be a lot easier to read/write the offset with a separate file.

PR #1081, opened 2020-03-09
CLI Postgres Module

bogdanromanx:


cli/postgres/src/test/scala/ch/epfl/bluebrain/nexus/cli/postgres/PostgresDocker.scala
```diff
@@ -0,0 +1,48 @@
...
+      PostgresConfig(knownAddress.hostV4, knownAddress.port)
+    }
+
+    // add docker dependencies and override default configuration
+    include(new DockerContainerModule[F] overridenBy new ModuleDef {
+      make[Docker.ClientConfig].from {
+        Docker.ClientConfig(
+          readTimeoutMs = 500,
+          connectTimeoutMs = 500,
+          allowReuse = true,
```
umbreak: shouldn't that be false? Since we are running it through a CI...

cli/postgres/src/test/scala/ch/epfl/bluebrain/nexus/cli/postgres/ProjectionSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+  )
+
+  "A DB" should {
+    "select 1" in {
+      import doobie.implicits._
+      (xa: Transactor[IO], eff: Effect[IO]) => {
+        implicit val e = eff
+        for {
+          result <- sql"select 1;".query[Int].unique.transact(xa)
+          _ = assert(result == 1)
```
umbreak: do we have to use this least expressive way of evaluating things, or we can still use
```scala
result shouldEqual 1
```

cli/postgres/src/test/scala/ch/epfl/bluebrain/nexus/cli/postgres/PostgresDocker.scala
bogdanromanx: it doesn't make a difference for the CI if the container remains active

cli/postgres/src/test/scala/ch/epfl/bluebrain/nexus/cli/postgres/ProjectionSpec.scala
bogdanromanx: I couldn't find one so far, but I will look into it.

cli/src/main/scala/ch/epfl/bluebrain/nexus/cli/sse/Offset.scala
```diff
@@ -0,0 +1,25 @@
...
+  lazy val asString: String = value.toString
+}
+
+object Offset {
+
+  /**
+    * Attempts to create an [[Offset]] from the passed string value.
+    */
+  final def apply(string: String): Option[Offset] =
+    Try(UUID.fromString(string)).map(Offset.apply).toOption
```
umbreak, line 20:
```scala
Try(UUID.fromString(string)) match {
   case Success(uuid) if uuid.version == 1 => Some(Offset(uuid))
   case _ => None
```

cli/src/main/scala/ch/epfl/bluebrain/nexus/cli/sse/Offset.scala
```diff
@@ -0,0 +1,25 @@
...
+
+  implicit final val offsetOrdering: Ordering[Offset] = (x: Offset, y: Offset) =>
+    x.value.timestamp() compareTo y.value.timestamp()
```
umbreak, line 23: If UUID is not version 1, `timestamp()` method will throw an error at runtime

PR #1146, opened 2020-04-30
Remove unused cli modules

bogdanromanx:
The PR also reconfigures the build to no longer depend on the sbt-nexus sbt plugin.


build.sbt
```diff
@@ -203,8 +161,119 @@ lazy val root = project
...
+  publishArtifact in (Test, packageSrc)    := false,
+  publishMavenStyle                        := true,
+  pomIncludeRepository                     := Function.const(false),
+  // removes compile time only dependencies from the resulting pom
+  pomPostProcess := { node =>
+    XmlTransformer.transformer(moduleFilter("org.scoverage") | moduleFilter("com.sksamuel.scapegoat")).transform(node).head
+  }
+)
+
+lazy val servicePackaging = {
```
umbreak, line 232: can't we move this somewhere out of the build.sbt?

It has way to many things.

bogdanromanx: not really, no...

PR #1147, opened 2020-04-30
Addressed previous postgres review comments

bogdanromanx:


cli/src/main/scala/ch/epfl/bluebrain/nexus/cli/sse/Offset.scala
```diff
@@ -2,7 +2,7 @@ package ch.epfl.bluebrain.nexus.cli.sse
 
 import java.util.UUID
 
-import scala.util.Try
+import scala.util.{Success, Try}
 
 /**
   * An offset for events.
```
umbreak, line 8: should the case class constructor then private, like `final case class private Offset` or at least `final case class private[sse] Offset`

PR #1151, opened 2020-05-04
Added postgres projection retry logic

bogdanromanx:


bogdanromanx:
> Those changes on `ClientRetryCondition` achieve something different in practice than just increasing the amount of code?
> 
> I see the classes using it not being affected...so I'm wondering why this is needed at all

It allows defining a ConfigConvert instance for the ClientRetryCondition and thus to be properly used as a type param. Classes like ClientConfig don't have to embed logic to serialize the config.

cli/src/main/scala/ch/epfl/bluebrain/nexus/cli/modules/postgres/PostgresProjection.scala
```diff
@@ -92,22 +100,28 @@ class PostgresProjection[F[_]: ContextShift](
...
         val elems = results.results.bindings.map { map =>
           map.toList.sortBy(_._1).map(_._2).map(binding => toElem(binding))
         }
-        val connections = Fragment.const(delete).update.run :: elems.map(row => Fragment(insert, row).update.run)
-        connections.sequence.transact(xa).map(_ => Right(()))
+        val statements = Fragment.const(delete).update :: elems.map(row => Fragment(insert, row).update)
+
+        implicit val log: (Throwable, RetryDetails) => F[Unit] = logFnForSqlStatements(statements)
+
+        statements.map(_.run).sequence.transact(xa).retryingOnAllErrors.map(_ => Right(()))
```
wwajerowicz, line 124: is there a way we could only retrying on server errors/timeouts etc and ignore client errors?

bogdanromanx: All pg errors are reported in the same way as far as I noticed. I'm not sure what you mean by "server errors/timeouts".

umbreak: wojtek is talking about the fact that `retryingOnAllErrors` will retry on ny error, even it is a client error, which is useless.

As we did on `SparqlClient`, something like that should be used instead:

```scala
private val retry                                = env.httpClient.retry
    private val successCondition                     = retry.retryCondition.notRetryFromEither[SparqlResults] _
    implicit private val retryPolicy: RetryPolicy[F] = retry.retryPolicy
    implicit private val logOnError: (ClientErrOr[SparqlResults], RetryDetails) => F[Unit] =
      (eitherErr, details) => Logger[F].info(s"Sparql client error '$eitherErr'. Retry details: '$details'")

  whateverFyouWantToRetry.retryingM(successCondition)
```

bogdanromanx: I don't see how it applies here, please explain.

umbreak: this line: `statements.map(_.run).sequence.transact(xa)`
is running the SQL INSERT queries, right?

So, we don't want to retry if, for example, the INSERT query is wrong (client error). Or if there is a gaateway timeout (since gateways will have a constant timeout threshold). That's why we introduced the `onServerError` ClientRetryCondition.

If that's not what this line is doing, then is my misunderstanding

wwajerowicz: I meant that it would be useful if we could skip retrying errors which will always fail(e.g. incorrect SQL statements). It might be tricky if all the pg errors are reported the same way.

bogdanromanx: 1. there's no gateway: direct tcp connection. 
2. we don't have additional information at this point about the cause of the errors

this is something that would be very interesting to invest in, but let's prepare the initial release first.

PR #1154, opened 2020-05-05
Improved HTTP client retry and replaced Logger with Console

umbreak:


cli/src/main/scala/ch/epfl/bluebrain/nexus/cli/clients/EventStreamClient.scala
```diff
@@ -70,7 +72,13 @@ object EventStreamClient {
           val req          = Request[F](uri = uri, headers = Headers(lastEventIdH.toList ++ env.authorizationHeader.toList))
           client
             .stream(req)
-            // TODO: handle client errors
+            .map {
```
bogdanromanx, line 75: evalMap with F.raiseError is preferred over throwing exceptions

cli/src/main/scala/ch/epfl/bluebrain/nexus/cli/modules/postgres/PostgresProjection.scala
```diff
@@ -48,34 +49,40 @@ class PostgresProjection[F[_]: ContextShift](
...
-      }
-      .through(printEvaluatedProjection(console))
-      .compile
-      .drain
+              .sequence
+              .map(_.sequence)
+        }
+        .through(printEvaluatedProjection(console))
+        .attempt
+        .map(_.leftMap(err => Unexpected(err.getMessage.take(30))).map(_ => ()))
```
bogdanromanx, line 82: err.getMessage.take can throw null pointer in some cases; it would be safer to wrap in an Option and getOrElse.

PR #1170, opened 2020-05-13
Prevent triggering unnecessary docker containers

umbreak:


cli/src/test/scala/ch/epfl/bluebrain/nexus/cli/AbstractCliSpec.scala
```diff
@@ -64,7 +66,11 @@ abstract class AbstractCliSpec
     pluginConfig = PluginConfig.empty,
     activation = StandardAxis.testDummyActivation,
     moduleOverrides = overrides,
```
neko-kai, line 68: Note: most likely you'll need to move the `defaultModules` into an object for memoization to be global - because most of the modules accept implicit parameters and close over these implicit parameters inside lambdas passed to `.fromEffect` - this makes lambdas non-constants and make each new instantiation of the module not `equal` to the previous one.

Each non-equal `TestConfig` will be put into a distinct memoization group, see [TestConfig scaladoc](https://github.com/7mind/izumi/blob/a7ffca059dcbdff65129ab7bb2dae8e5370341d5/distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/TestConfig.scala#L30) - so to ensure global memoization it's best for lambdas inside `ModuleDef` to not close over external parameters or to instantiate modules that do use closures once and put them in an object.

PR #1176, opened 2020-05-14
Remove recursion on LocatorRef

neko-kai:
* In Config/Influx/Postgres subcommands, do not create unrelated components (using GCMode to select one component for which dependencies should be created)
* remove axis tags from components with only a single possible implementation
* Move out test IO modules into an object to help reducing amount of different memoization envs (see https://github.com/BlueBrain/nexus/pull/1170/files#r424561485)
* ^ Whitelist `DistageTestRunner.runEnvs` log - this helps diagnose having too many memoization envs, each line corresponds to a new memoization env


bogdanromanx:
Many thanks for the PR @neko-kai, much appreciated. We'd like to merge the PR, but first we need to look into the process for accepting external contributions. This may take a bit of time, but we'll post back as soon as we have something.

Cheers!

bogdanromanx:
Hi @neko-kai, we've decided to drop this module completely, pending a new implementation that follows your advice on how to deal with CLI args evaluation and testing. We therefore need to close this PR without merging it. We thank you very much for your contribution and for the really nice framework!

PR #1181, opened 2020-05-19
Initial porting of IAM service

umbreak:
Ported as it was on the `nexus-iam` repo with the following changes:

- Use Akka Http `Uri` implementation
- Changed `HttpClient`
- Small changes on the package structure in some cases

TODO:
- Change the ACLs to constrain the possible ACLs to the root, the org or the project level (that will remove some complexity)
- Use F[_+, +_]
- Add minimal support for JsonLD (for now we only need to inject the context Uris into the Json document for the response).
- Have a different ADT for Events stored in the primary store and for HTTP responses (to be debated) ??


.github/workflows/ci-service.yml
```diff
@@ -0,0 +1,63 @@
+name: Continuous Integration Service
+on:
+  push:
+    branches:
+      - master
+    paths:
```
bogdanromanx, line 6: also needs to include _sourcing_

build.sbt
```diff
@@ -208,11 +226,58 @@ lazy val sourcing = project
...
+      akkaHttpCirce,
+      akkaHttpCors,
+      catsCore,
+      catsEffectRetry,
+      catsEffect,
+      kryo,
+      monixEval,
+      nimbusJoseJwt,
+      splitBrainLithium,
+      "net.bytebuddy" % "byte-buddy-agent" % "1.10.10",
```
bogdanromanx, line 253: I would recommend extracting these in a separate block:
```
lazy val kanelaAgent = ...

lazy val kamonSettings = Seq(
  libraryDependencies ++= Seq(
    kanelaAgent,
    ...
  )
)
```

and adding them to the service project under using the setting clause.

Issue #1068, opened 2020-03-04
Unify RDF library in Nexus repository


umbreak:
1. Create a submodule in nexus repo for the rdf library.
2. Port the required functionality, while refactoring when needed/desired.

PR #1207, opened 2020-05-27
ported RDF from nexus-rdf repo

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1068


.github/workflows/ci-service.yml
```diff
@@ -5,12 +5,14 @@ on:
...
     paths:
       - 'sourcing/**'
+      - 'rdf/**'
       - 'service/**'
       - 'build.sbt'
       - 'project/**'
   pull_request:
     paths:
       - 'sourcing/**'
+      - 'df/**'
```
bogdanromanx, line 15: typo here

build.sbt
```diff
@@ -13,10 +13,11 @@ scalafmt: {
 
 val javaSpecificationVersion = "11"
 val scalacSilencerVersion    = "1.6.0"
+val scalaCompileVersion      = "2.13.1"
```
bogdanromanx, line 16: typo: `scalaCompilerVersion` instead of `scalaCompileVersion`; I would recommend using `scalaVersion` instead if it doesn't create a conflict with an SBT key.

build.sbt
```diff
@@ -198,10 +203,47 @@ lazy val sourcing = project
...
   )
+lazy val rdf = project
+  .in(file("rdf"))
+  .settings(shared, compilation, coverage, release)
+  .settings(
+    name            := "rdf",
+    moduleName      := "rdf"
+  )
+  .settings(
+    libraryDependencies ++= Seq(
```
bogdanromanx, line 214: there are a lot of dependencies here are don't seem to be justified, like: kryo, akkaPersistence etc. please remove them.

build.sbt
umbreak: it does

PR #1235, opened 2020-06-11
Adapted IAM/ADMIN communication

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1218


service/src/main/scala/ch/epfl/bluebrain/nexus/admin/organizations/OrganizationCommand.scala
```diff
@@ -3,7 +3,8 @@ package ch.epfl.bluebrain.nexus.admin.organizations
 import java.time.Instant
 import java.util.UUID
 
-import ch.epfl.bluebrain.nexus.iam.client.types.Identity.Subject
+import ch.epfl.bluebrain.nexus.iam.types.Identity.Subject
+import ch.epfl.bluebrain.nexus.iam.types.Identity.Subject
```
bogdanromanx, line 7: duplicate import

service/src/test/scala/ch/epfl/bluebrain/nexus/admin/routes/ProjectRoutesSpec.scala
```diff
@@ -178,8 +209,7 @@ class ProjectRoutesSpec
     }
 
     "create a project without optional fields" in new Context {
-      iamClient.hasPermission(Path("/org").rightValue, create)(any[Option[AuthToken]]) shouldReturn Task.pure(true)
-      iamClient.identities shouldReturn Task(caller)
+//      iamClient.hasPermission(Path("/org").rightValue, create)(any[Option[AuthToken]]) shouldReturn Task.pure(true)
```
bogdanromanx, line 212: why do we need to keep these commented lines?

service/src/main/scala/ch/epfl/bluebrain/nexus/service/exceptions/ServiceError.scala
```diff
@@ -36,7 +36,6 @@ object ServiceError {
     case i: IamError   => IamError.iamErrorEncoder(i)
     case a: AdminError => AdminError.adminErrorEncoder(a)
     case r =>
-      println("dupa dupa")
```
bogdanromanx, line 39: nice catch :)

wwajerowicz: oops :)

service/src/main/scala/ch/epfl/bluebrain/nexus/admin/config/AdminConfig.scala
```diff
@@ -32,7 +30,7 @@ final case class AdminConfig(
 object AdminConfig {
 
   final case class ServiceAccountConfig(token: Option[String]) {
-    def credentials: Option[AuthToken] = token.map(AuthToken)
+    def credentials: Option[AccessToken] = token.map(AccessToken)
```
wwajerowicz, line 33: It's a bit weird  to be  HTTP Bearer token when not doing interacting with a service via HTTP.  We can probably refactor that later.

umbreak: Yeah. The thing is we still a `Subject` to make calls to the `Realms` and `Acls` API.
We could have instead something like
```scala
final case class ServiceAccount(realm: String, name: String){
   def subject: Subject = User(realm, name)
}
```

But that's more prone to errors than having the actual saToken from where the subject is going to be extracted by the JWT automatically

PR #1313, opened 2020-07-10
Indexed labels on ElasticSearch when present

### LINKED ISSUE #1255: Index schema:name or rdfs:label to ElasticSearch, return field in listings
Related to #913 .

umbreak:
Fixes #1255


service/src/test/resources/view/elasticview-source.json
```diff
@@ -27,6 +27,18 @@
         "type": "keyword",
         "copy_to": "_all_fields"
       },
+      "prefLabel": {
```
imsdu, line 30: Other fields are prefixed by _, why not those ?
I don't know if there is any norm so just asking.
Anf maybe they could be included in an object so that we have:
container  { prefLabel, label, name }

service/src/main/scala/ch/epfl/bluebrain/nexus/service/config/Vocabulary.scala
```diff
@@ -12,6 +12,11 @@ import ch.epfl.bluebrain.nexus.rdf.implicits._
   */
 object Vocabulary {
 
+  object skos {
+    private val baseStr = "http://www.w3.org/2004/02/skos/core#"
+    val prefLabel       = IriNode(url"${baseStr}prefLabel")
+  }
+
```
wwajerowicz, line 19: couldn't this go to rdf `Vocabulary` together with `rdfs`?

PR #1325, opened 2020-07-16
Changed S3 location from Path-Styled to Virtual Host Styled

### LINKED ISSUE #1251: Port the KG S3StorageOperationsSpec using an s3 compatible service within a container

umbreak:
Fixes #1251 

This fix is necessary due to the [deprecation of path-styles on S3](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/).

It uses Minio in a docker container which is triggered during Unit Tests. Notice that the to run this locally you'll have to run the following commands:
```
          echo "127.0.0.1 bucket.my-domain.com" | sudo tee -a /etc/hosts
          echo "127.0.0.1 bucket2.my-domain.com" | sudo tee -a /etc/hosts
```

### DIFF
--- a/.github/workflows/ci-delta.yml
+++ b/.github/workflows/ci-delta.yml
@@ -36,7 +36,10 @@ jobs:
       - name: StaticAnalysis
         run: sbt -Dsbt.color=always -Dsbt.supershell=false "project delta" clean scalafmtCheck test:scalafmtCheck scalafmtSbtCheck scapegoat
       - name: Tests
-        run: sbt -Dsbt.color=always -Dsbt.supershell=false "project delta" clean coverage test coverageReport coverageAggregate
+        run: |
+          echo "127.0.0.1 bucket.my-domain.com" | sudo tee -a /etc/hosts
+          echo "127.0.0.1 bucket2.my-domain.com" | sudo tee -a /etc/hosts
+          sbt -Dsbt.color=always -Dsbt.supershell=false "project delta" clean coverage test coverageReport coverageAggregate
   publish:
     if: github.event_name == 'push'
     runs-on: ubuntu-latest
--- a/.github/workflows/ci-delta.yml
+++ b/.github/workflows/ci-delta.yml
@@ -37,8 +37,8 @@ jobs:
         run: sbt -Dsbt.color=always -Dsbt.supershell=false "project delta" clean scalafmtCheck test:scalafmtCheck scalafmtSbtCheck scapegoat
       - name: Tests
         run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/ci-delta.yml
```diff
@@ -36,7 +36,10 @@ jobs:
       - name: StaticAnalysis
         run: sbt -Dsbt.color=always -Dsbt.supershell=false "project delta" clean scalafmtCheck test:scalafmtCheck scalafmtSbtCheck scapegoat
       - name: Tests
-        run: sbt -Dsbt.color=always -Dsbt.supershell=false "project delta" clean coverage test coverageReport coverageAggregate
+        run: |
+          echo "127.0.0.1 bucket.my-domain.com" | sudo tee -a /etc/hosts
```
imsdu, line 40: It will add an entry in /etc/hosts everytime

delta/src/test/scala/ch/epfl/bluebrain/nexus/kg/storage/S3StorageOperationsSpec.scala
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  override protected def config: TestConfig =
+    TestConfig(
+      pluginConfig = PluginConfig.empty,
+      activation = StandardAxis.testDummyActivation,
+      parallelTests = ParallelLevel.Sequential,
+      moduleOverrides = new ModuleDef {
+        // add docker dependencies and override default configuration
+        include(new DockerSupportModule[IO] overridenBy new ModuleDef {
```
imsdu, line 106: Bogdan did something similar maybe we could put some stuff in common ?

delta/src/test/scala/ch/epfl/bluebrain/nexus/util/MinioDocker.scala
```diff
@@ -0,0 +1,38 @@
...
+
+object MinioDocker extends ContainerDef with Randomness {
+  val primaryPort: DockerPort = DockerPort.TCP(9000)
+  val accessKey               = "my_key"
+  val secretKey               = "my_secret_key"
+  val virtualHost             = "my-domain.com"
+
+  override def config: Config = {
+    Config(
+      image = "minio/minio:latest",
```
imsdu, line 16: A release would be better than the latest

.github/workflows/ci-delta.yml
umbreak: Yes and that's fine

PR #1332, opened 2020-07-21
CI publish and release to DockerHub

umbreak:


build.sbt
```diff
@@ -287,7 +287,7 @@ lazy val storage = project
     coverageFailOnMinimum    := true,
     buildInfoKeys            := Seq[BuildInfoKey](version),
     buildInfoPackage         := "ch.epfl.bluebrain.nexus.storage.config",
-    Docker / packageName     := "storage",
+    Docker / packageName     := "nexus-storage",
     javaSpecificationVersion := "1.8",
```
imsdu, line 291: We have to change java version here too

umbreak: Actually I want 1.8 for this service. This flag should allow to build with java8, however it is being ignored or it is not enough. We will have to investigate further later, that's why I commented the other 2.

PR #1335, opened 2020-07-22
Add the service account caller explicitly instead of its token

umbreak:


delta/src/main/resources/app.conf
```diff
@@ -40,11 +40,12 @@ app {
     default-timeout = 30 seconds
   }
 
-  # Service account settings
-  service-account {
-    # the service account token, if needed (when not required, anonymous subject will be used)
-    token = null
-  }
+  # caller settings for when the service wants perform service calls. Do not asign a value if you want the caller to be anonymous
```
imsdu, line 43: To delete or to uncomment I guess

delta/src/main/scala/ch/epfl/bluebrain/nexus/delta/Main.scala
```diff
@@ -169,7 +166,8 @@ object Main {
         ArchiveCache[Task].runSyncUnsafe()(Scheduler.global, pm)
       )
     implicit val pc: ProjectCache[Task]           = cache.project
-    implicit val projectResolution                = ProjectResolution.task(repo, cache.resolver, cache.project, acls, saCaller)
+    implicit val projectResolution                =
```
imsdu, line 169: Formatting is weird here

delta/src/main/resources/app.conf
umbreak: No. It is left as a documented example of which parameters needs to be set when defining a caller

delta/src/main/scala/ch/epfl/bluebrain/nexus/delta/Main.scala
umbreak: Yeah. It is scalafmt formatting. Not a big deal

PR #1351, opened 2020-07-24
Added release notes for 1.4

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1201

### DIFF
--- a/docs/src/main/paradox/docs/releases/v1.3-to-v1.4-migration.md
+++ b/docs/src/main/paradox/docs/releases/v1.3-to-v1.4-migration.md
@@ -8,7 +8,7 @@ The migration steps are as follows:
 
 2. [Backup](https://docs.datastax.com/en/archived/cassandra/3.0/cassandra/operations/opsBackupRestore.html) the
      cassandra store.
-     
+          
 3. Delete all ElasticSearch indices:
 
 ```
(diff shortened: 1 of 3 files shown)


docs/src/main/paradox/docs/releases/v1.3-to-v1.4-migration.md
```diff
@@ -0,0 +1,190 @@
...
+    'bucket_high' : 1.5,
+    'bucket_low' : 0.5,
+    'max_threshold' : 32,
+    'min_threshold' : 4,
+    'min_sstable_size' : 50
+    };
+```      
+
+6. Deploy the new service images (tag `1.4`) for [delta](https://hub.docker.com/r/bluebrain/nexus-delta) with the `MIGRATE_V13_TO_V14=true` and `REPAIR_FROM_MESSAGES=true` environment variables.
+This will instruct the service to migrate all the messages table from iam, admin and kg into the delta keyspace. Afterwards the `tag_views` table and their related tables will be properly initialized.
```
imsdu, line 189: - It is not clear if the old tables are deleted afterwards.
- Maybe we should warn that we need twice the disk space used by Cassandra during the migration ?
- Redeploy the service afterwards without the flags ?

PR #1377, opened 2020-07-29
Update running instructions

wwajerowicz:


docs/src/main/paradox/docs/getting-started/running-nexus/index.md
```diff
@@ -542,133 +501,9 @@ $ kubectl get configmap/config -o yaml | grep public.ip:
...
-  "_results": []
-}
-$
-```
-
-#### Deploy KG
+#### Deploy Delta
 
-KG is the service that manages user defined resources, their schemas and configuration like resolvers, views etc.
+Delta is the service providing Nexus REST API/
```
carolili, line 506: Should there be a slash at the end of this sentence + "the" is missing? 🤔
E.g. "Delta is the service providing the Nexus REST API."

wwajerowicz: Fixed.

PR #1381, opened 2020-08-03
Use default timeout from configuration file

umbreak:


delta/src/main/scala/ch/epfl/bluebrain/nexus/delta/config/AppConfig.scala
```diff
@@ -73,6 +74,13 @@ object AppConfig {
 
   }
 
+  /**
+    * The runtime configuration
+    *
+   * @param defaultTimeout arbitrary Future completion timeout when
```
imsdu, line 80: The end is missing

delta/src/main/scala/ch/epfl/bluebrain/nexus/delta/config/AppConfig.scala
```diff
@@ -73,6 +74,13 @@ object AppConfig {
...
+    */
+  final case class RuntimeConfig(defaultTimeout: FiniteDuration)
```
imsdu, line 82: It looks more like a shutdown timeout no ?

PR #1390, opened 2020-08-06
Refactored Elasticsearch client tests to use docker

### LINKED ISSUE #1217: Adapt kg unit tests to use containers

bogdanromanx:
Also memoized container checks for BlazeGraph, Minio and Cassandra.

Fixes #1217


delta/src/test/scala/ch/epfl/bluebrain/nexus/util/ElasticSearchDocker.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import distage.{ModuleDef, TagK}
+import izumi.distage.docker.ContainerDef
+import izumi.distage.docker.Docker.DockerPort
+import izumi.distage.docker.healthcheck.ContainerHealthCheck
+
+object ElasticSearchDocker extends ContainerDef {
+  val primaryPort: DockerPort = DockerPort.TCP(9200)
+
+  override def config: Config = {
```
imsdu, line 11: I don't know where it can fit but we could add the fact that all indices have only one shard and no replica by default.
Config or template I guess

bogdanromanx: I'll look into it separately.

PR #1392, opened 2020-08-06
Added notes for REPAIR_FROM_MESSAGES migration

umbreak:


docs/src/main/paradox/docs/releases/v1.3-to-v1.4-migration.md
```diff
@@ -178,12 +173,13 @@ The migration steps are as follows:
...
+8. Deploy the new service images (tag `1.4`) for [delta](https://hub.docker.com/r/bluebrain/nexus-delta) and run it with the `MIGRATE_V13_TO_V14=true` and `REPAIR_FROM_MESSAGES=true` environment variables.
 This will instruct the service to copy all the messages table from iam, admin and kg into the delta keyspace. Afterwards the `tag_views` table and their related tables will be properly initialized.
+When populating the `tag_views` table consider increasing the `batch_size_fail_threshold_in_kb` in your cassandra.yaml configuration file if you see TagWriter error messages. Decreasing the value of the JVM property `akka.persistence.cassandra.events-by-tag.max-message-batch-size` and `akka.persistence.cassandra.journal.max-message-batch-size` to `1` can also help mitigate the issue.
 
-10. Once the service is up and running, and you have verified the resources are accessible, you can proceed to delete the previous keyspaces:
+9. Once the service is up and running, and you have verified the resources are accessible, you can proceed to delete the previous keyspaces:
```
imsdu, line 182: Maybe some examples of queries to run to make sure it is ok ?

umbreak: This is going to depend a lot on the scale of the deployment. counts would time out for big scale deployments. There they would have to do other things like keep some api responses before the service migration and after and compare them to see they are the same

Issue #1405, opened 2020-08-10
Unify duplicated and conflicting global SSE routes


umbreak:
the SSE endpoint `/{prefix}/events` is duplicated in the following:
- iam.routes.EventRoutes
- admin.routes.EventRoutes
- kg.routes.GlobalEventRoutes

This needs to be unified and handled at once

PR #1410, opened 2020-08-11
Fixed global SSE

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1405


delta/src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/KgRoutes.scala
```diff
@@ -155,12 +155,6 @@ class KgRoutes(
         extractCaller { implicit caller =>
           implicit val subject: Subject = caller.subject
           concat(
-            (get & pathPrefix(config.http.prefix / "events") & pathEndOrSingleSlash) {
-              new GlobalEventRoutes(acls, realms, caller).routes
-            },
-            (get & pathPrefix(config.http.prefix / "resources" / "events") & pathEndOrSingleSlash) {
```
bogdanromanx, line 161: we should still have this endpoint, right? or is it handled somewhere else?

delta/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/GlobalEventRoutes.scala
```diff
@@ -38,32 +48,44 @@ import scala.util.{Failure, Success, Try}
...
 
+  // format: off
   def routes: Route =
     concat(
-      routesFor("orgs" / "events", OrganizationTag, orgs.read, typedEventToSse[OrganizationEvent]),
-      routesFor("projects" / "events", ProjectTag, projects.read, typedEventToSse[ProjectEvent]),
-      routesFor("events", EventTag, events.read, eventToSse)
+      routesFor("events", EventTag, s"/$prefix/events", events.read, eventToSse),
+      routesFor("resources" / "events", EventTag, "/resources/events", resRead, typedEventToSse[Event]),
+      routesFor("resources" / "events", EventTag, "/resources/events", resRead, typedEventToSse[Event]),
```
wwajerowicz, line 67: Duplicate `/resources/events` route.

PR #1431, opened 2020-08-14
Remove IAM references from storage service.

wwajerowicz:


wwajerowicz:
> Can you remove all the environment variables? We don't need them

Done.

PR #1438, opened 2020-08-14
Merging to master

umbreak:


.github/workflows/ci-product-page.yml
```diff
@@ -0,0 +1,65 @@
+name: Continuous Integration Docs
```
bogdanromanx, line 1: name must be different

PR #1440, opened 2020-08-17
Post release fixes

adulbrich:


docs/src/main/product-page-src/src/pages/index.tsx
```diff
@@ -65,7 +65,7 @@ export default function Home() {
         id="why"
         title="Why Blue Brain Nexus?"
         subtitle="Nexus was born out of a need for better data management in the
-            simulation neuroscience field, but built to cover any use case.
+            simulation neuroscience field, but built to cover any use cases.
```
kenjinp, line 68:
```suggestion
            simulation neuroscience field, but built to cover any other use case.
```
grammatically I don't think this change is necessary

PR #1449, opened 2020-08-17
Added error handling for message table migration

bogdanromanx:
Also added two env variables that control the parallelism of the
migration (defaulting to 1):
*   MIGRATE_V13_TO_V14_PARALLELISM
*   REPAIR_FROM_MESSAGES_PARALLELISM

### DIFF
--- a/delta/src/main/scala/ch/epfl/bluebrain/nexus/delta/MigrateV13ToV14.scala
+++ b/delta/src/main/scala/ch/epfl/bluebrain/nexus/delta/MigrateV13ToV14.scala
@@ -5,22 +5,28 @@ import java.util.{UUID, Set => JSet}
 
 import akka.NotUsed
 import akka.actor.ActorSystem
+import akka.stream.alpakka.cassandra.CassandraSessionSettings
 import akka.stream.alpakka.cassandra.scaladsl._
-import akka.stream.alpakka.cassandra.{CassandraSessionSettings, CassandraWriteSettings}
 import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
 import ch.epfl.bluebrain.nexus.delta.MigrateV13ToV14._
 import ch.epfl.bluebrain.nexus.delta.config.AppConfig
 import ch.epfl.bluebrain.nexus.sourcing.projections.Projections.cassandraDefaultConfigPath
+import com.datastax.oss.driver.api.core.cql.{BoundStatement, PreparedStatement}
 import com.typesafe.scalalogging.Logger
 import monix.eval.Task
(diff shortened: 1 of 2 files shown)


delta/src/main/scala/ch/epfl/bluebrain/nexus/delta/RepairFromMessages.scala
```diff
@@ -18,6 +25,18 @@ object RepairFromMessages {
 
   private val log = Logger[RepairFromMessages.type]
 
+  private val parallelism = sys.env.getOrElse("REPAIR_FROM_MESSAGES_PARALLELISM", "1").toIntOption.getOrElse(1)
```
umbreak, line 28: You could have use a property from `migration.conf` (MigrationConfig in AppConfig), since this already exists and it is used for this purpose.

delta/src/main/scala/ch/epfl/bluebrain/nexus/delta/RepairFromMessagesClassic.scala
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+  * Repair tool to rebuild the dependent tables in case tag_views is out of sync with the messages tables.
+  */
+object RepairFromMessagesClassic {
+  // $COVERAGE-OFF$
+
+  private val log = Logger[RepairFromMessagesClassic.type]
+
+  private val parallelism = sys.env.getOrElse("REPAIR_FROM_MESSAGES_PARALLELISM", "1").toIntOption.getOrElse(1)
```
umbreak, line 39: You could have use a property from `migration.conf` (MigrationConfig in AppConfig), since this already exists and it is used for this purpose.

delta/src/main/scala/ch/epfl/bluebrain/nexus/delta/MigrateV13ToV14.scala
```diff
@@ -5,22 +5,28 @@ import java.util.{UUID, Set => JSet}
...
-import scala.concurrent.Future
 import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.control.NonFatal
+import scala.util.matching.Regex
 
 // $COVERAGE-OFF$
 class MigrateV13ToV14(implicit config: AppConfig, session: CassandraSession, as: ActorSystem) {
+
+  private val parallelism = sys.env.getOrElse("MIGRATE_V13_TO_V14_PARALLELISM", "1").toIntOption.getOrElse(1)
```
umbreak, line 28: Is this even needed? Does it speed up thing or just make it slower?

PR #1464, opened 2020-08-19
added 2 references for RDF and LD

adulbrich:


bogdanromanx (approved):

carolili (approved):
The `Authors` could be `Author` to match the rest of the document if you wish to update that. (The reason I put it in singular even if some documents have multiple authors is that I'm [used to doing that from LaTeX](https://www.overleaf.com/learn/latex/bibliography_management_with_bibtex#The_bibliography_file).

But not a big deal, feel free to merge @adulbrich 😃

Issue #1222, opened 2020-06-08
Resource resolution fails when specifying uuids in the schema and resource segments


bogdanromanx:

imsdu:
- assuming a resource constrained by a schema (other than unconstrained.json)
- if the schema is identified by a generated uuid
- if the resource is identified by a generated uuid
- when trying to fetch the resource using: /v1/resources/{org}/{proj}/{schema_uuid}/{resource_uuid} it returns 404
- when trying to fetch the resource using a wildcard for the schema: /v1/resources/{org}/{proj}/_/{resource_uuid} it returns the resource
- when trying to fetch the schema using its uuid: /v1/resources/{org}/{proj}/_/{schema_uuid} it returns the schema

And we expect exactly /v1/resources/{org}/{proj}/{schema_uuid}/{resource_uuid} == /v1/resources/{org}/{proj}/_/{resource_uuid}

PR #1474, opened 2020-08-19
Add some tests for already resolved issue #1222

imsdu:
#1222 seems already fixed

I just completed the test to avoid a regression which may occur.


delta/src/test/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourceRoutesSpec.scala
```diff
@@ -251,8 +259,10 @@ class ResourceRoutesSpec
 
       val endpoints = List(
         s"/v1/resources/${projectMeta.value.organizationUuid}/${projectMeta.uuid}/_/$urlEncodedId",
+        s"/v1/resources/${projectMeta.value.organizationUuid}/${projectMeta.uuid}/$schemaEncodedId/$urlEncodedId",
```
bogdanromanx, line 262: please add tests with variations where the schema id segment is just the schema uuid like for example `1988e0be-587e-4a49-a0d0-ea5e63bcc85b` and the resource id segment is just the resource uuid.

bogdanromanx (changes requested):

PR #1478, opened 2020-08-20
Add some tests for already resolved issue #1222

imsdu:
#1222 seems already fixed

I just completed the test to avoid a regression which may occur.


umbreak (approved):

bogdanromanx (changes requested):
This still doesn't test the very specific case of retrieving a resource based on the schema uuid and resource uuid as path segments. (just uuids, no prefixes, no names, no url encodings)

bogdanromanx (dismissed):

bogdanromanx (approved):

PR #1579, opened 2020-09-01
Added minimal JSON-LD high level API support

umbreak:
Provides a way to:
- Expand JSON-LDs
- Compact JSON-LDs
- Convert JSON-LDs to Jena Model (toRdf)
- Convert Jena Model to Json-LDs (fromRdf)
- Basic operations on an expanded Json-LD (add/fetch properties)
- Basic operations on an compacted Json-LD (add/fetch properties)
- Abstraction over the underlying Json-LD high level API and JsonLdOptions
- Resolve remote contexts
- It does not use any "magic" or anything out of the Json-LD spec
- Graph abstraction over Jena Model for basic filtering

TODO:
- Graph conversions to dot and ntriples


delta/rdf/src/test/scala/ch/epfl/bluebrain/nexus/delta/rdf/Fixtures.scala
```diff
@@ -0,0 +1,52 @@
...
+import org.scalatest.OptionValues
+
+trait Fixtures extends TestHelpers with CirceLiteral with EitherValuable with OptionValues {
+
+  val iri = iri"http://nexus.example.com/john-doé"
+
+  // format: off
+  val remoteContexts: Map[Uri, Json] =
+    Map(
+      Uri("http://example.com/context/0")  -> json"""{"@context": {"deprecated": {"@id": "http://schema.org/deprecated", "@type": "http://www.w3.org/2001/XMLSchema#boolean"} }}""",
```
imsdu, line 23: We have an interpolator for iri and json, mayne for Uri, it makes sense too ?

delta/rdf/src/test/resources/logback.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
imsdu, line 1: We could have only one logback-test.xml in the testkit resources

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/JsonLdError.scala
```diff
@@ -0,0 +1,46 @@
+package ch.epfl.bluebrain.nexus.delta.rdf.jsonld
+
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolutionError
+
+@SuppressWarnings(Array("IncorrectlyNamedExceptions"))
+sealed abstract class JsonLdError(reason: String) extends Exception {
```
imsdu, line 6: We could have here directly the underlying exception

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/api/JsonLdJavaApi.scala
```diff
@@ -0,0 +1,150 @@
...
+  // $COVERAGE-OFF$
+  override def compact[Ctx <: JsonLdContext](input: Json, ctx: Json, f: ContextFields[Ctx])(implicit
+      opts: JsonLdOptions,
+      resolution: RemoteContextResolution
+  ): IOErrorOr[(Json, Ctx)] =
+    for {
+      obj           <- catchTryErr(JsonUtils.fromString(input.noSpaces), "building input")
+      ctxObj        <- catchTryErr(JsonUtils.fromString(ctx.noSpaces), "building context")
+      options       <- remoteContextLoader(input, ctx).map(toOpts)
+      compacted     <- catchTryErr(JsonUtils.toString(JsonLdProcessor.compact(obj, ctxObj, options)), "compacting")
```
imsdu, line 37: We only want to manipulate json so maybe we could parse it in the same pass

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/api/JsonLdJavaApi.scala
```diff
@@ -0,0 +1,150 @@
...
+      input: Json
+  )(implicit options: JsonLdOptions, resolution: RemoteContextResolution): IOErrorOr[Json] =
+    for {
+      obj          <- catchTryErr(JsonUtils.fromString(input.noSpaces), "building input")
+      options      <- remoteContextLoader(input).map(toOpts)
+      expanded     <- catchTryErr(JsonUtils.toString(JsonLdProcessor.expand(obj, options)), "expanding")
+      expandedJson <- IO.fromEither(parser.parse(expanded).leftMap(err => UnexpectedJsonLd(err.getMessage())))
+    } yield expandedJson
+
+  override def frame[Ctx <: JsonLdContext](input: Json, frame: Json, f: ContextFields[Ctx])(implicit
```
imsdu, line 53: frame, expand and compact have common steps, maybe we can put do something with it

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/context/ExtendedJsonLdContext.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  lazy val aliasesInv: Map[IRI, String] = aliases.map { case (prefix, iri) => iri -> prefix }
+
+  lazy val prefixMappingsInv: Map[IRI, String] = prefixMappings.map { case (prefix, iri) => iri -> prefix }
+
+  override def addPrefix(prefix: String, iri: IRI): This                                  =
+    copy(value = add(prefix, iri.asJson), prefixMappings = prefixMappings + (prefix -> iri))
+
+  protected def addAlias(prefix: String, iri: IRI, dataType: Option[String] = None): This =
+    dataType match {
```
imsdu, line 34: Maybe more something like:
`copy(value = add(prefix, 
                              dataType.fold(iri.asJson)(expandedTermDefinition(dt, iri))), 
                              aliases = aliases + (prefix -> iri))
`

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/context/JsonLdContext.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  private def arrOrObj(arr: Seq[Json]): Json =
+    arr.take(2).toList match {
+      case Nil         => Json.obj()
+      case head :: Nil => head
+      case _ :: _      => Json.arr(arr: _*)
+    }
+
+  private def merge(json: Json, that: Json): Json =
+    (json.asArray, that.asArray, json.asString, that.asString) match {
```
imsdu, line 149: Why the asString part ?

umbreak: To check the case of context being a string, like 
```json
{
   "@context": "http://schema.org/"
}
```

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/api/JsonLdJavaApi.scala
umbreak: It is better understood like this than having a obscure refactoring. Repetition is not always a bad thing.

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/api/JsonLdJavaApi.scala
umbreak: Every step issues different errors, so I find it cleaner as is.

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/JsonLdError.scala
umbreak: This is the way we have been defining errors on KG, CLI and everywhere else

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/api/JsonLdApi.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.IOErrorOr
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.{ContextFields, JsonLdContext, RemoteContextResolution}
+import io.circe.Json
+import org.apache.jena.rdf.model.Model
+
+/**
+  * Json-LD high level API as defined in https://www.w3.org/TR/json-ld11-api/
+  */
+trait JsonLdApi {
```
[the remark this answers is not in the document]
umbreak: We need a `IO` because part of the JsonLdApi task is to resolve remote contexts (`Uri => IO[RemoteContextResolutionError, Json]`)

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/ExpandedJsonLd.scala
```diff
@@ -0,0 +1,95 @@
...
+  def add(key: IRI, literal: Int): This =
+    add(key, literal, includeDataType = false)
+
+  def add(key: IRI, literal: Long): This =
+    add(key, literal, includeDataType = false)
+
+  def add(key: IRI, literal: Double): This =
+    add(key, literal, includeDataType = false)
+
+  def add(key: IRI, literal: Int, includeDataType: Boolean): This =
```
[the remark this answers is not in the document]
umbreak: I find it quite clear for clients as is:
```scala
jsonld.add(age, 1, includeDataType = true)
```

What is your alternative suggestion?

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/api/JsonLdJavaApi.scala
imsdu: Agreed.

What about something like:
`private def process(input: Json, stage: Stage)(f: Any => Any) = // Stages could typed too
    for {
      obj             <- fromString(input.noSpaces, "building input")
      options      <- remoteContextLoader(input).map(toOpts)
      result         <- toJson(f(obj, options)), stage)
    } yield result

def expand(input: Json) = process(input, Stage.Expanding)(JsonLdProcessor.expand) 
`
JsonLdProcessor makes the thing a little meh

umbreak: I've try few things like that, but they introduce more code and more jumping from here to there.

Since this is the Json-LD Java implementation part of the code I think it is clear and straightforward as is. If it evolves to something else in the future, we can consider refactoring.

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/ExpandedJsonLd.scala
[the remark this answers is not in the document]
imsdu: I think I read a little fast, it is okay this way :)

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/context/ExtendedJsonLdContext.scala
```diff
@@ -0,0 +1,38 @@
...
+final case class ExtendedJsonLdContext(
+    value: Json,
+    base: Option[IRI] = None,
+    vocab: Option[IRI] = None,
+    aliases: Map[String, IRI] = Map.empty,
+    prefixMappings: Map[String, IRI] = Map.empty
+) extends JsonLdContext {
+  type This = ExtendedJsonLdContext
+
+  lazy val aliasesInv: Map[IRI, String] = aliases.map { case (prefix, iri) => iri -> prefix }
```
bogdanromanx, line 26: this could hide duplicates, is that acceptable?

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/package.scala
```diff
@@ -0,0 +1,7 @@
+package ch.epfl.bluebrain.nexus.delta.rdf
+
+import monix.bio.IO
+
+package object jsonld {
+  type IOErrorOr[A] = IO[JsonLdError, A]
```
bogdanromanx, line 6: There are a lot of aliases, maybe it would be better to use the actual type to avoid having to look up their meaning. Thoughts?

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/syntax/JsonSyntax.scala
```diff
@@ -0,0 +1,29 @@
...
+import io.circe.Json
+import org.apache.jena.iri.IRI
+
+trait JsonSyntax {
+  implicit final def contextSyntax(json: Json): JsonOps = new JsonOps(json)
+}
+
+final class JsonOps(private val json: Json) extends AnyVal {
+
+  def topContextValue: Option[Json] = JsonLdContext.topContextValue(json)
```
bogdanromanx, line 14: if this is the intended use of the api, maybe add some docs?

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/utils/SeqUtils.scala
```diff
@@ -0,0 +1,12 @@
+package ch.epfl.bluebrain.nexus.delta.rdf.utils
+
+trait SeqUtils {
+
+  def headOnlyOption[A](seq: Seq[A]): Option[A] =
```
bogdanromanx, line 5: add docs, it may be difficult to understand its purpose just from the name.

build.sbt
```diff
@@ -446,7 +476,7 @@ lazy val storage = project
 
 lazy val delta = project
```
bogdanromanx, line 477: project should aggregate on all submodules, like rdf for example?

bogdanromanx (changes requested):

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/api/JsonLdJavaApi.scala
```diff
@@ -0,0 +1,150 @@
...
+      ctxValue   <- context(frame, f, options)
+    } yield (framedJson, ctxValue)
+
+  // $COVERAGE-OFF$
+  // TODO: create the DocumentLoader and pass it to Jena: https://issues.apache.org/jira/browse/JENA-1959
+  override def toRdf(
+      input: Json
+  )(implicit opts: JsonLdOptions, resolution: RemoteContextResolution): IOErrorOr[Model] = {
+    val model       = ModelFactory.createDefaultModel()
+    val initBuilder = RDFParser.create.fromString(input.noSpaces).lang(Lang.JSONLD)
```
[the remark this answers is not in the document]
umbreak: No. The modelfactory does not throw and the second line is just a builder, at that point nothing is executed

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/context/ExtendedJsonLdContext.scala
umbreak: It could. In case of duplicates it will choose the first sorted alphabetically. So I'll change the implementation

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/context/RemoteContextResolution.scala
```diff
@@ -0,0 +1,54 @@
...
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution.Result
+
+trait RemoteContextResolution {
+
+  /**
+    * Resolve a passed ''uri''.
+    *
+   * @return the expected Json payload response from the passed ''uri''
+    */
+  def resolve(uri: Uri): Result[Json]
```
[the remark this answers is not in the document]
umbreak: Context resolution happens over the wire, over HTTP, which does not support IRIs. 
In the JSON-LD spec the meontions to remote context are saying it is a Uri and also in Json-LD Java implementation and titanium

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/package.scala
umbreak: Most of the times this one is used `IOErrorOr[A]`, since in most parts of the code I deal with `JsonLdError`.

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/syntax/JsonSyntax.scala
umbreak: All the docs are in the `JsonLdContext`. Do I have to duplicate the docs, or remove them from the `JsonLdContext` and put them on the syntax. 

Since we always said the syntax should not contain logic, why the docs should be there? (note there aren't docs for syntax usually in cats, etc)

build.sbt
umbreak: you mean the `lazy val root` ?

I don't understand

bogdanromanx: shouldn't the delta project have an aggregate statement (`.aggregate(rdf, sourcing)`)?

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/context/RemoteContextResolution.scala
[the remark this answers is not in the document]
bogdanromanx: But we don't allow that and instead we use resource identifiers which can be iris, right? We could transform the resource ids into Uris and back during the resolution, but does it make sense?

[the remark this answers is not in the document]
umbreak: Ok. It seems to work with IRIs as well as with Uris. I changed the signature to IRI since it is true that the Nexus resources can be IRIs, not just Uris

bogdanromanx (approved):

Issue #1291, opened 2020-07-01
Get rid of the rdf module and use Jena instead


umbreak:
Get rid of our own Iri and Graph implementation. 
Use akka `Uri` for Uris and when we need Iri support, use Jena `Iri`.
Use the JenaModel and the Jena graph representation to deal with graphs.

PR #1586, opened 2020-09-04
Added conversions: Graph → DOT, Graph → N-Triples and Graph ↔ Json

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1291

All the necessary conversions are now in place:

- CompactedJsonLd ↔ ExpandedJsonLd
- CompactedJsonLd ↔ Graph
- ExpandedJsonLd ↔ Graph
- Graph → Dot
- Graph → N-Triples


delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/graph/Graph.scala
```diff
@@ -85,8 +97,54 @@ final case class Graph private (root: IRI, model: Model) { self =>
...
+  )(implicit api: JsonLdApi, resolution: RemoteContextResolution, opts: JsonLdOptions): IO[GraphError, Dot] =
+    toDot(Json.obj(keywords.context -> context.value))
+
+  /**
+    * Attempts to convert the current Graph with the passed ''context'' as Json to the DOT format: https://graphviz.org/doc/info/lang.html
+    * If the context is empty the DOT output will have the expanded form
+    */
+  def toDot(context: ExtendedJsonLdContext): Either[GraphError, Dot]                                        =
+    Try {
+      RDFWriter.create().lang(DOT).source(model).set(ROOT_ID, rootResource).set(JSONLD_CONTEXT, context).asString()
```
imsdu, line 143: RDFWriter was also in your previous PR, maybe we could have an utility method which would take care about wrapping this in an Either ?

imsdu (changes requested):
Sometimes, it seems that we are losing the information that we have a json object and not only generic Json and then we have to check it again.
Maybe in JsonLdApi, we may return more refined types than Json to avoid this trouble ?

It just needs some documentation at some points to better understand what is done.
And I guess I will also need an explanation of the rdf part at some point :)

Otherwise it looks good to me.

PR #1602, opened 2020-09-14
Defined Acls events, states, rejections and contract

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1488

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/AclState.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/AclState.scala
@@ -1,0 +1,98 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.model.acls
+
+import java.time.Instant
+
+import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.{nxv, schemas}
+import ch.epfl.bluebrain.nexus.delta.sdk.AclTargetResource
+import ch.epfl.bluebrain.nexus.delta.sdk.model.Identity.Subject
+import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceRef.Latest
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{Identity, ResourceF}
+import org.apache.jena.iri.IRI
+
+/**
+  * Enumeration of ACLs states.
+  */
+sealed trait AclState extends Product with Serializable {
+
+  /**
+    * @return the current state revision
+    */
+  def rev: Long
+
+  /**
+    * @return the current deprecation status (always false for acls)
+    */
+  def deprecated: Boolean = false
+
+  /**
+    * Converts the state into a resource representation.
+    *
+    * @param id the resource identifier
+    */
+  def toResource(id: IRI): AclTargetResource
+}
+
+object AclState {
+
+  /**
+    * Initial state type.
+    */
+  type Initial = Initial.type
+
+  /**
+    * Initial state for the permission set.
+    */
+  final case object Initial extends AclState {
+    override val rev: Long = 0L
+
+    override def toResource(id: IRI): AclTargetResource =
+      ResourceF(
+        id = id,
+        rev = rev,
+        types = Set(nxv.AccessControlList),
+        deprecated = deprecated,
+        createdAt = Instant.EPOCH,
+        createdBy = Identity.Anonymous,
+        updatedAt = Instant.EPOCH,
+        updatedBy = Identity.Anonymous,
+        schema = Latest(schemas.acls),
+        value = Target.Root -> Acl.empty
+      )
+  }
+
+  /**
+    * An existing ACLs state.
+    *
+    * @param target    the target location for the ACL
+    * @param acl       the Access Control List
+    * @param rev       the ACLs revision
+    * @param createdAt the instant when the resource was created
+    * @param createdBy the identity that created the resource
+    * @param updatedAt the instant when the resource was last updated
+    * @param updatedBy the identity that last updated the resource
+    */
+  final case class Current(
+      target: Target,
+      acl: Acl,
+      rev: Long,
+      createdAt: Instant,
+      createdBy: Subject,
+      updatedAt: Instant,
+      updatedBy: Subject
+  ) extends AclState {
+    override def toResource(id: IRI): AclTargetResource =
+      ResourceF(
+        id = id,
+        rev = rev,
+        types = Set(nxv.AccessControlList),
+        deprecated = deprecated,
+        createdAt = Instant.EPOCH,
+        createdBy = Identity.Anonymous,
+        updatedAt = Instant.EPOCH,
+        updatedBy = Identity.Anonymous,
+        schema = Latest(schemas.acls),
+        value = target -> acl
+      )
+  }
+
+}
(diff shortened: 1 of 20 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/AclState.scala
```diff
@@ -0,0 +1,98 @@
...
+      updatedAt: Instant,
+      updatedBy: Subject
+  ) extends AclState {
+    override def toResource(id: IRI): AclTargetResource =
+      ResourceF(
+        id = id,
+        rev = rev,
+        types = Set(nxv.AccessControlList),
+        deprecated = deprecated,
+        createdAt = Instant.EPOCH,
```
imsdu, line 89: Why don't we use values from the current state for this field and the following ones ?

umbreak: You are right!

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Target.scala
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+  * Enumeration of possible ACL targets. A target is the location where a certain ACL is anchored.
+  */
+sealed trait Target extends Product with Serializable {
+
+  /**
+    * the string representation of the target location
+    */
+  def string: String
```
wwajerowicz, line 13: why not use regular `toString` method for this or `Show` from `cats` here?

imsdu (dismissed):

wwajerowicz (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Target.scala
umbreak: the regular `toString` is already used. However I also wanted to save it in a variable

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/AclState.scala
```diff
@@ -0,0 +1,108 @@
...
+    */
+  type Initial = Initial.type
+
+  /**
+    * Initial state for the permission set.
+    */
+  final case object Initial extends AclState {
+    override val rev: Long = 0L
+
+    override def toResource(id: IRI): AclTargetResource =
```
bogdanromanx, line 58: I think the id type of an ACL resource can be the target itself. The toResource won't need a param anymore. I was having the same dilemma with the realms, but since the ResourceF takes an Id type param, setting a different id makes the problem go away.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Acls.scala
```diff
@@ -0,0 +1,109 @@
...
+    */
+  def fetch(target: Target): Task[Option[AclResource]]
+
+  /**
+    * Fetches the ACL resource for a ''target'' on the passed revision.
+    *
+    * @param target the target location for the ACL
+    * @param rev    the revision to fetch
+    */
+  def fetchAt(target: Target, rev: Long): IO[RevisionNotFound, Option[AclResource]]
```
bogdanromanx, line 26: Do we need an option here in the result type? I understand that it allows to discriminate between the "resource is not found" and "resource exists but the revision doesn't ", but in terms of responses you'd still return a 404.
I went the other way, but if you think this is better, I will adjust the contract for realms.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Acls.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+    * Fetches the ACL resource for a ''target'' on the current revision.
+    *
+    * @param target the target location for the ACL
+    * @param self   flag to decide whether or not ACL of other identities than the provided ones should be included in the response.
+    *               This is constrained by the current caller having ''acls/read'' permissions on the provided ''target'' or it's parents
+    */
+  final def fetch(target: Target, self: Boolean)(implicit caller: Caller): Task[Option[AclResource]] =
```
bogdanromanx, line 35: I would make this function explicit: `def fetchSelf` since the routes should discriminate between the intent. This way we can explicitly ask for a caller argument and we can drop the self argument.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Acls.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+    * Fetches the ACL resource for a ''target'' on the passed revision.
+    *
+    * @param target the target location for the ACL
+    * @param self   flag to decide whether or not ACL of other identities than the provided ones should be included in the response.
+    *               This is constrained by the current caller having ''acls/read'' permissions on the provided ''target'' or it's parents
+    * @param rev    the revision to fetch
+    */
+  final def fetchAt(target: Target, self: Boolean, rev: Long)(implicit
```
bogdanromanx, line 46: same as above

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/permissions/PermissionsState.scala
```diff
@@ -22,14 +22,23 @@ sealed trait PermissionsState extends Product with Serializable {
   /**
     * @return the current deprecation status (always false for permissions)
     */
-  def deprecated: Boolean = false
+  final def deprecated: Boolean = false
+
+  /**
+    * @return the schema reference that permissions conforms to
+    */
+  final def schema: ResourceRef = Latest(schemas.permissions)
```
bogdanromanx, line 30: This way of encoding the value will not work for automatic derivation of typeclasses like Encoder. Same applies for Acls. Maybe that's ok, but it will generate more code.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/package.scala
```diff
@@ -11,4 +12,14 @@ package object sdk {
     */
   type PermissionsResource = ResourceF[IRI, Set[Permission]]
 
+  /**
+    * Type alias for a acl with its target location specific resource.
+    */
+  type AclTargetResource = ResourceF[IRI, (Target, Acl)]
```
bogdanromanx, line 18: I think these could be joined into a single: `ResourceF[Target, Acl]`

bogdanromanx (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Acls.scala
umbreak: Since the response payload is going to be different, we should distinguish them

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Acls.scala
umbreak: Sure. Even though that's gonna be more code on the routes side, since `self` is a query parameter, we'll have to do:
```scala
parameters("self" ? false) { self =>
   if(self) fetchSelf(target)
   else fetch(target)
}
```


```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/permissions/PermissionsState.scala
umbreak: We will see when we get to the Encoding and Decoding, right?

bogdanromanx (approved):

PR #1605, opened 2020-09-14
Add Plugin API.

### LINKED ISSUE #1513: Provide a plugin interface for bootstrapping a plugin
The plugin should be able to be notified when the system has started and when the system has successfully joined the cluster.

The plugin should have the opportunity to augment the route definitions.

The plugin should have access to the entire event log.

The plugin should have access to the application configuration and actor system.

The plugin should be able to store information on disk.

The plugin should be able to store information on the primary store.

The plugin should have access to instances of all contracts in the api module.

wwajerowicz:
`PluginDependencies` should be expanded as we add more interfaces to the `sdk`.

Fixes #1513.


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/Plugin.scala
```diff
@@ -0,0 +1,31 @@
...
+  */
+trait Plugin {
+
+  /**
+    * Start the plugin.
+    *
+    * @param deps  plugin dependencies
+    * @return optional Akka Http [[Route]], if the plugin defines one.
+    */
+  def start(deps: PluginDependencies): Task[Option[Route]]
```
imsdu, line 17: I don't know which library is best suited to inject dependencies to the plugin but I think we will have to use one.
For the return type, besides the route, we could return the plugin id, the plugin version, a healthcheck function, maybe a retry strategy and more

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/Plugin.scala
```diff
@@ -0,0 +1,31 @@
...
+    * Stop the plugin. This should allow the plugin to terminate gracefully.
+    */
+  def stop(): Task[Unit]
+
+  /**
+    * Retrieve plugin status.
+    *
+    * @return current status of the plugin
+    */
+  def status(): Task[PluginStatus]
```
imsdu, line 29: How about include it as a function after running so that we could maybe more easily handle grace periods and stuff like that ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/PluginStatus.scala
```diff
@@ -0,0 +1,33 @@
...
+  */
+final case class PluginStatus(name: String, version: String, status: Status) {}
+
+object PluginStatus {
+
+  /**
+    * Plugin status.
+    */
+  sealed trait Status extends Product with Serializable
+  object Status {
```
imsdu, line 20: Starting and Stopping could be also interesting statuses too.

And as we can stop them maybe splitting down in 2, one for errors and one when stopping has been asked

imsdu: And if the plugin failed, it would be nice to have a reason

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/Plugin.scala
bogdanromanx, line 17: I think we need the opportunity for a plugin to provide (register) contract definitions and instances during startup (such that plugins could possibly depend on other plugins).
A plugin also needs to be able to lookup instances provided by other plugins.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/Plugin.scala
```diff
@@ -0,0 +1,31 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.plugin
+
+import akka.http.scaladsl.server.Route
+import monix.bio.Task
+
+/**
+  * Plugin API.
+  */
+trait Plugin {
```
bogdanromanx, line 9: A plugin needs to expose a name and/or id and its version.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/PluginStatus.scala
bogdanromanx: starting and stopping would be interesting, but I think it complicates things... do you foresee a situation where a plugin has a long initialization?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/Plugin.scala
bogdanromanx: I'm not sure what you mean by this...

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/PluginStatus.scala
```diff
@@ -0,0 +1,33 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.plugin.PluginStatus.Status
+
+/**
+  * Representation of a [[Plugin]] status.
+  *
+  * @param name    name of the [[Plugin]]
+  * @param version version of the [[Plugin]]
+  * @param status  status of the [[Plugin]]
+  */
+final case class PluginStatus(name: String, version: String, status: Status) {}
```
bogdanromanx, line 12: the realms PR introduces a new type: Name; I would suggest we use it here instead of a string.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/PluginDependencies.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  /**
+    * Permissions API
+    */
+  def permissions: Permissions
+
+  /**
+    * Identities API
+    */
+  def identities: Identities
```
bogdanromanx, line 18: also add realms and acls

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/Plugin.scala
bogdanromanx: So maybe the plugin interface should be split in two, like so:

```
trait PluginDef {
  def name: Name
  def version: String
  def dependencies: Set[PluginDef]
  def initialise(registry: Registry): Task[Plugin]
}
trait Plugin {
  def name: Name
  def version: String
  def dependencies: Set[PluginDef]

  def route: Route
  def stop(): Task[Unit]
}
trait Registry {
  def register[A](value: A): Task[Unit]
  def lookup[A]: IO[ReferenceNotFound, A]
}
```
It would allow declaring dependencies before initialisation, an opportunity to register instances during initialisation, route registration after initialisation and a stop hint for coordinated shutdown.

We could simplify things by dropping the status for now... we could simply expect that plugins manage their own lifecycle; during the service shutdown, we can notify plugins to stop whatever they're doing.

bogdanromanx (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/PluginStatus.scala
imsdu: If communication with an external component takes time because this one is down for whatever reason.
Because most of the stuff should already be running.

But adding them could be added later when we will have more time or need to refine lifecycle for monitoring purposes for example.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/error/PluginError.scala
```diff
@@ -0,0 +1,14 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.error
+
+trait PluginError extends SDKError
```
bogdanromanx, line 3: make this a `sealed abstract class PluginError(reason: String, details: Option[String]) extends SDKError`
... and transform the `DependencyNotFound` into a case class with the appropriate information about what hasn't been found, like: `final case class DependencyNotFound(dependencies: Set[(Name, String)]) extends ...`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/Plugin.scala
```diff
@@ -0,0 +1,43 @@
...
+    * Stop the plugin. This should allow the plugin to terminate gracefully.
+    */
+  def stop(): Task[Unit]
+
+  /**
+    * Retrieve plugin status.
+    *
+    * @return current status of the plugin
+    */
+  def status(): Task[PluginStatus]
```
bogdanromanx, line 42: I would drop the status for now

bogdanromanx (changes requested):

umbreak (approved):

bogdanromanx (approved):

PR #1613, opened 2020-09-16
Added delta-sdk-testkit, delta-service and delta-all modules

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1499
Fixes https://github.com/BlueBrain/nexus/issues/1529
Fixes https://github.com/BlueBrain/nexus/issues/1528

### DIFF
--- a/.github/workflows/ci-app.yml
+++ b/.github/workflows/ci-app.yml
@@ -29,9 +29,9 @@ jobs:
     strategy:
       fail-fast: true
     steps:
-      - uses: actions/checkout@v2.1.0
-      - name: Cache Coursier
-        uses: actions/cache@v1.1.2
+      - name: Checkout
+      - uses: actions/checkout@v2
+      - uses: coursier/cache-action@v3
         with:
           path: ~/.cache/coursier
           key: ${{ runner.os }}-coursier-scala-${{ hashFiles('**/*.sbt') }}
@@ -53,9 +53,9 @@ jobs:
       BINTRAY_USER: ${{ secrets.BINTRAY_USER }}
       BINTRAY_PASS: ${{ secrets.BINTRAY_PASS }}
     steps:
-      - uses: actions/checkout@v2.1.0
-      - name: Cache Coursier
-        uses: actions/cache@v1.1.2
+      - name: Checkout
+      - uses: actions/checkout@v2
+      - uses: coursier/cache-action@v3
         with:
           path: ~/.cache/coursier
           key: ${{ runner.os }}-coursier-scala-${{ hashFiles('**/*.sbt') }}
--- a/.github/workflows/ci-app.yml
+++ b/.github/workflows/ci-app.yml
@@ -31,11 +31,7 @@ jobs:
     steps:
       - name: Checkout
       - uses: actions/checkout@v2
-      - uses: coursier/cache-action@v3
-        with:
-          path: ~/.cache/coursier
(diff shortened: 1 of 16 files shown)


.github/workflows/ci-app.yml
```diff
@@ -0,0 +1,77 @@
...
+      - uses: olafurpg/setup-scala@v5
+        with:
+          java-version: adopt@1.11
+    - name: PublishToBintray
+      run: sbt -Dsbt.color=always -Dsbt.supershell=false "project app" releaseEarly
+    - name: PublishToDockerHub
+      run: |
+        sbt -Dsbt.color=always -Dsbt.supershell=false "project app" docker:publishLocal
+        echo ${{ secrets.DOCKER_PASS }} | docker login --username ${{ secrets.DOCKER_USER }} --password-stdin
+        docker push bluebrain/nexus-delta-app:latest
```
bogdanromanx, line 72: We should have `bluebrain/nexus-delta` here and to modify the build.sbt to set the correct docker image name.

.github/workflows/ci-app.yml
```diff
@@ -0,0 +1,77 @@
+name: Continuous Integration Delta App
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'delta/sourcing/**'
+      - 'delta/rdf/**'
+      - 'delta/sdk/**'
+      - 'delta/sdk-testkit/**'
```
bogdanromanx, line 10: if we also include the sdk-testkit we could just have `delta/**`

.github/workflows/ci-sdk-testkit.yml
```diff
@@ -0,0 +1,66 @@
...
+      - 'project/**'
+jobs:
+  review:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+    steps:
+      - uses: actions/checkout@v2.1.0
+      - name: Cache Coursier
```
bogdanromanx, line 27: We could possibly replace these across all jobs to use coursier cache action: https://github.com/marketplace/actions/coursier-cache-action

bogdanromanx (changes requested):

wwajerowicz (dismissed):

bogdanromanx (dismissed):

.github/workflows/ci-app.yml
umbreak: Not yet, since we have a lot of things in `delta/src/**` which we don't care about on the `app` module

.github/workflows/ci-app.yml
umbreak: For now I will comment this part. Let's deal with it afterwards

.github/workflows/ci-app.yml
```diff
@@ -29,9 +29,9 @@ jobs:
     strategy:
       fail-fast: true
     steps:
-      - uses: actions/checkout@v2.1.0
-      - name: Cache Coursier
-        uses: actions/cache@v1.1.2
+      - name: Checkout
+      - uses: actions/checkout@v2
+      - uses: coursier/cache-action@v3
```
bogdanromanx, line 34: I think it should be v4 and should not have the with part.

bogdanromanx (changes requested):

bogdanromanx (dismissed):

PR #1615, opened 2020-09-16
Defined Projects events, states, rejections and contract

### LINKED ISSUE #1490: Define Projects events, states, rejections and contract in the 'api' module

bogdanromanx:
Fixes #1490

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/Project.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/Project.scala
@@ -1,0 +1,36 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.model.projects
+
+import java.util.UUID
+
+import ch.epfl.bluebrain.nexus.delta.sdk.model.Label
+import org.apache.jena.iri.IRI
+
+/**
+  * A project representation.
+  *
+  * @param label             the project label
+  * @param uuid              the project unique identifier
+  * @param organizationLabel the parent organization label
+  * @param organizationUuid  the parent organization unique identifier
+  * @param description       an optional description
+  * @param apiMappings       the API mappings
+  * @param base              the base IRI for generated resource IDs
+  * @param vocab             an optional vocabulary for resources with no context
+  */
+final case class Project(
+    label: Label,
+    uuid: UUID,
+    organizationLabel: Label,
+    organizationUuid: UUID,
+    description: Option[String],
+    apiMappings: Map[String, IRI],
+    base: IRI,
+    vocab: IRI
+) {
+
+  /**
+    * @return a project label reference containing the parent organization label
+    */
+  def projectRef: ProjectRef =
+    ProjectRef(organizationLabel, label)
+}
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/Project.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/Project.scala
@@ -31,6 +31,6 @@ final case class Project(
   /**
     * @return a project label reference containing the parent organization label
     */
-  def projectRef: ProjectRef =
+  def ref: ProjectRef =
     ProjectRef(organizationLabel, label)
 }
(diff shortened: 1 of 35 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/Project.scala
```diff
@@ -0,0 +1,36 @@
...
+    description: Option[String],
+    apiMappings: Map[String, IRI],
+    base: IRI,
+    vocab: IRI
+) {
+
+  /**
+    * @return a project label reference containing the parent organization label
+    */
+  def projectRef: ProjectRef =
```
umbreak, line 34: isn't `def ref: ProjectRef` enough?
since most of the times the code would look like
```
val project = ...
project.ref
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectDescription.scala
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+  * Type that represents a project payload for creation and update requests.
+  *
+  * @param description an optional description
+  * @param apiMappings the API mappings
+  * @param base        an optional base IRI for generated resource IDs
+  * @param vocab       an optional vocabulary for resources with no context
+  */
+final case class ProjectDescription(
```
umbreak, line 13: Maybe `ProjectData` or `ProjectFields` instead?
there is a `description` field here so it seems duplicated.

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/Project.scala
bogdanromanx: yep

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectDescription.scala
bogdanromanx: yep

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/Project.scala
imsdu: We could have the ref directly in the case class no ?
Most of the time, we use label and organizationLabel together no ?
This kind of composite key could make some signatures simpler

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/Project.scala
bogdanromanx: This class is mostly meant to encapsulate project information for serialization. I think it would generate some noise if we wrap those fields. Also, the ref doesn't contain the uuids so it would the class asymmetric... I prefer we keep it as is.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectRef.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import ch.epfl.bluebrain.nexus.delta.sdk.model.Label
+
+/**
+  * A project label along with its parent organization label.
+  *
+  * @param organization the parent organization label
+  * @param project      the project label
+  */
+final case class ProjectRef(organization: Label, project: Label)
```
umbreak, line 11:
```scala
override def toString(): String = s"$organization/$project"
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectState.scala
```diff
@@ -0,0 +1,132 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceRef.Latest
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{Label, ResourceF, ResourceRef}
+import org.apache.jena.iri.IRI
+
+/**
+  * Enumeration of Project state types.
+  */
+sealed trait ProjectState extends Product with Serializable {
+
+  /**
```
umbreak, line 18: Don't we add here the `rev` as well, like for the rest of states?

```scala
  /**
    * @return the current state revision
    */
  def rev: Long
```

umbreak (changes requested):

umbreak (approved):

PR #1616, opened 2020-09-16
Implemented Permissions state machine and PermissionsDummy

### LINKED ISSUE #1501: Provide a dummy implementation for the Permissions module

### LINKED ISSUE #1516: Extract the Permissions module, refactor it to implement the new contract and use the new RDF api

bogdanromanx:
Fixes #1501

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
@@ -1,7 +1,15 @@
 package ch.epfl.bluebrain.nexus.delta.sdk
 
-import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions.PermissionsRejection.RevisionNotFound
-import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions.{Permission, PermissionsRejection}
+import java.time.Instant
+import java.util.concurrent.TimeUnit
+
+import cats.effect.Clock
+import ch.epfl.bluebrain.nexus.delta.sdk.model.Identity.Subject
+import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions.PermissionsCommand._
+import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions.PermissionsEvent._
+import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions.PermissionsRejection._
+import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions.PermissionsState.{Current, Initial}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions._
 import monix.bio.{IO, UIO}
 
 /**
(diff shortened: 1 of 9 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+      case e: PermissionsAppended   => appended(e)
+      case e: PermissionsReplaced   => replaced(e)
+      case e: PermissionsSubtracted => subtracted(e)
+      case e: PermissionsDeleted    => deleted(e)
+    }
+  }
+
+  private[delta] def evaluate(minimum: Set[Permission])(state: PermissionsState, cmd: PermissionsCommand)(implicit
+      clock: Clock[UIO] = IO.timer.clock
+  ): IO[PermissionsRejection, PermissionsEvent] = {
```
umbreak, line 144: Do we need all these return types? I mean they are already on the `def evaluate` method. It seems repetitive.

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/PermissionsSpec.scala
```diff
@@ -0,0 +1,149 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.model
```
umbreak, line 1: Tests are way too compacted, I cannot understand much. Wouldn't it be easier to separate it in several scalatest blocks?

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+    def accept(f: Instant => PermissionsEvent): UIO[PermissionsEvent]            =
+      clock.realTime(TimeUnit.MILLISECONDS).map(rt => f(Instant.ofEpochMilli(rt)))
+    def reject[A <: PermissionsRejection](rejection: A): IO[A, PermissionsEvent] =
+      IO.raiseError(rejection)
+
+    def replace(c: ReplacePermissions): IO[PermissionsRejection, PermissionsEvent]   =
```
umbreak, line 150: Do we need all these return types? I mean they are already on the def evaluate method. It seems repetitive.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+        case Initial                    =>
+          val appended = c.permissions -- minimum
+          if (appended.isEmpty) reject(CannotAppendEmptyCollection)
+          else accept(PermissionsAppended(1L, c.permissions, _, c.subject))
+        case s: Current                 =>
+          val appended = c.permissions -- s.permissions -- minimum
+          if (appended.isEmpty) reject(CannotAppendEmptyCollection)
+          else accept(PermissionsAppended(c.rev + 1, c.permissions, _, c.subject))
+      }
+    def subtract(c: SubtractPermissions): IO[PermissionsRejection, PermissionsEvent] =
```
umbreak, line 168: same

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+        case Initial                    => reject(CannotSubtractFromMinimumCollection(minimum))
+        case s: Current                 =>
+          val intendedDelta = c.permissions -- s.permissions
+          val delta         = c.permissions & s.permissions
+          val subtracted    = delta -- minimum
+          if (intendedDelta.nonEmpty) reject(CannotSubtractUndefinedPermissions(intendedDelta))
+          else if (subtracted.isEmpty) reject(CannotSubtractFromMinimumCollection(minimum))
+          else accept(PermissionsSubtracted(c.rev + 1, subtracted, _, c.subject))
+      }
+    def delete(c: DeletePermissions): IO[PermissionsRejection, PermissionsEvent]     =
```
umbreak, line 181: same

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+      clock.realTime(TimeUnit.MILLISECONDS).map(rt => f(Instant.ofEpochMilli(rt)))
+    def reject[A <: PermissionsRejection](rejection: A): IO[A, PermissionsEvent] =
+      IO.raiseError(rejection)
+
+    def replace(c: ReplacePermissions): IO[PermissionsRejection, PermissionsEvent]   =
+      if (c.rev != state.rev) reject(IncorrectRev(c.rev, state.rev))
+      else if (c.permissions.isEmpty) reject(CannotReplaceWithEmptyCollection)
+      else if ((c.permissions -- minimum).isEmpty) reject(CannotReplaceWithEmptyCollection)
+      else accept(PermissionsReplaced(c.rev + 1, c.permissions, _, c.subject))
+    def append(c: AppendPermissions): IO[PermissionsRejection, PermissionsEvent]     =
```
umbreak, line 155: same

umbreak (changes requested):

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/PermissionsSpec.scala
bogdanromanx: It's a straightforward (state, event) => state and (state, command) => event test series that cover all cases. I could add more comments if that helps to show what's tested.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+      case e: PermissionsReplaced   => replaced(e)
+      case e: PermissionsSubtracted => subtracted(e)
+      case e: PermissionsDeleted    => deleted(e)
+    }
+  }
+
+  private[delta] def evaluate(minimum: Set[Permission])(state: PermissionsState, cmd: PermissionsCommand)(implicit
+      clock: Clock[UIO] = IO.timer.clock
+  ): IO[PermissionsRejection, PermissionsEvent] = {
+    def accept(f: Instant => PermissionsEvent): UIO[PermissionsEvent]            =
```
umbreak, line 145: you don't need this, see my PR, you can make use of: https://github.com/BlueBrain/nexus/blob/2a488ecf464bb23180ba00bc61f9108ff23ab4e3/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/utils/IOUtils.scala#L14

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+      clock.realTime(TimeUnit.MILLISECONDS).map(rt => f(Instant.ofEpochMilli(rt)))
+    def reject[A <: PermissionsRejection](rejection: A): IO[A, PermissionsEvent] =
```
umbreak, line 147: This method just creates noise here. Just add directly in the pattern match the `IO.raiseError(...)`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
bogdanromanx: I like return types... it's a general good practice and helps inference, but I could remove them if you feel strongly about it.

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
umbreak: For a method inside a method, which always returns the same as the parent method....seems quite overkilling

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/PermissionsSpec.scala
umbreak: Is the same to add comments as `//` than scalatest blocks. 

They are there to group tests and give some extra context, otherwise we would always do that, and clutter everything in one.

But in anyway is done is fine :)

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+  }
+
+  private[delta] def next(
+      minimum: Set[Permission]
+  )(state: PermissionsState, event: PermissionsEvent): PermissionsState = {
+    val apply                                                  = withPermissions(state, minimum) _
+    def appended(e: PermissionsAppended): PermissionsState     =
+      state match {
+        case Initial if e.rev == 1L           => apply(e.permissions, e.instant, e.subject)
+        case s: Current if s.rev + 1 == e.rev => apply(s.permissions ++ e.permissions, e.instant, e.subject)
```
imsdu, line 116: We will find this condition on revision for every state/event, we may find a way to abstract it ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+  private[delta] def evaluate(minimum: Set[Permission])(state: PermissionsState, cmd: PermissionsCommand)(implicit
+      clock: Clock[UIO] = IO.timer.clock
+  ): IO[PermissionsRejection, PermissionsEvent] = {
+    def accept(f: Instant => PermissionsEvent): UIO[PermissionsEvent]            =
+      clock.realTime(TimeUnit.MILLISECONDS).map(rt => f(Instant.ofEpochMilli(rt)))
+    def reject[A <: PermissionsRejection](rejection: A): IO[A, PermissionsEvent] =
+      IO.raiseError(rejection)
+
+    def replace(c: ReplacePermissions): IO[PermissionsRejection, PermissionsEvent]   =
+      if (c.rev != state.rev) reject(IncorrectRev(c.rev, state.rev))
```
imsdu, line 151: It may be out of scope for now but we could accumulate rejections.
With https://typelevel.org/cats/datatypes/validated.html it is quite simple to handle
It is not very interesting for permissions but for more complicated cases, it could be very nice.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
umbreak: The state machine is quite self containing and explicit. I don't think abstracting in this situation is helping to understand the code

imsdu: In your PR, there is no check on revision when applying the event and it is true that an incorrect revision must be caught there. So maybe, we just don't need them at all at this point ?

umbreak: You are right. The check on rev is done on the evaluate function. The next function should not do that, it is not its responsibility.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+    val apply                                                  = withPermissions(state, minimum) _
+    def appended(e: PermissionsAppended): PermissionsState     =
+      state match {
+        case Initial if e.rev == 1L           => apply(e.permissions, e.instant, e.subject)
+        case s: Current if s.rev + 1 == e.rev => apply(s.permissions ++ e.permissions, e.instant, e.subject)
+        case other                            => other
+      }
+    def replaced(e: PermissionsReplaced): PermissionsState     =
+      state match {
+        case s if s.rev + 1 == e.rev => apply(e.permissions, e.instant, e.subject)
```
umbreak, line 121: unnecessary check on rev

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+        case other                   => other
+      }
+    def subtracted(e: PermissionsSubtracted): PermissionsState =
+      state match {
+        case s: Current if s.rev + 1 == e.rev => apply(s.permissions -- e.permissions, e.instant, e.subject)
```
umbreak, line 126: unnecessary check on rev

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -71,3 +79,118 @@ trait Permissions {
...
+        case other                            => other
+      }
+    def deleted(e: PermissionsDeleted): PermissionsState       =
+      state match {
+        case s: Current if s.rev + 1 == e.rev => apply(Set.empty, e.instant, e.subject)
```
umbreak, line 131: unnecessary check on rev

umbreak (changes requested):

umbreak (approved):

PR #1630, opened 2020-09-18
Implemented IOSemaphore and IORef wrappers

bogdanromanx:


umbreak (commented):
Looks good. Maybe we wanna keep it in `testkit` module instead? Since maybe `sourcing` or another module wants to make use of this now that we use IO *everywhere*

umbreak (approved):

PR #1631, opened 2020-09-18
Plugin loading

wwajerowicz:


bogdanromanx (approved):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/plugin/PluginLoader.scala
```diff
@@ -0,0 +1,103 @@
...
+      .enableAllInfo()
+      .scan()
+      .getClassesImplementing(classOf[PluginDef].getName)
+      .getNames
+      .asScala
+      .toList
+
+    pluginDefClasses match {
+      case pluginDef :: Nil =>
+        IO.pure(pluginClassLoader.create[PluginDef](pluginDef, println)())
```
imsdu, line 39: We need to use a logger and not println

delta/service/src/test/scala/ch/epfl/bluebrain/nexus/delta/service/plugin/PluginLoaderSpec.scala
```diff
@@ -0,0 +1,36 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.Permissions
+import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions.Permission
+import ch.epfl.bluebrain.nexus.delta.sdk.testkit.PermissionsDummy
+import ch.epfl.bluebrain.nexus.delta.service.plugin.PluginLoader.PluginLoaderConfig
+import izumi.distage.model.definition.ModuleDef
+import monix.execution.Scheduler.Implicits.global
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+class PluginLoaderSpec extends AnyWordSpecLike with ScalatestRouteTest with Matchers {
```
imsdu, line 13: It would be nice to have tests about error cases too, try to load several plugins too, one working and one failing.
It could be done later but what if we try to load different versions of the same plugin ?

imsdu (changes requested):

PR #1635, opened 2020-09-21
Added Support for Any (*) ACL Target

umbreak:
Also changed the naming `Target` for `AclAddress` and every reference to target in the docs


wwajerowicz (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Target.scala
```diff
@@ -29,32 +22,46 @@ sealed trait Target extends Product with Serializable {
 
 object Target {
 
+  /**
+    * An ACL target location where a certain ACL is anchored.
+    */
+  sealed trait TargetLocation extends Target {
```
bogdanromanx, line 28: I think that possibly a disjoint ADT would be better, something like:
```
sealed trait Target
case object Root
case class Org(label)
case class Proj(label, label)
```
and
```
sealed trait TargetSelection
case object AnyOrg  /*
case object AnyOrgAnyProj  /*/*
case class FixedOrgAnyProj(label) /label/*
```
... because it allows one to discriminate between fixed addressable resources (targets) and queries (target selections). I don't think there's a need to unify these as it generates confusion. The routes will be able to identify for GET requests if the provided segments are wildcards or fixed values and choose the appropriate intent.

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/AclTargetsSpec.scala
```diff
@@ -4,7 +4,14 @@ import cats.syntax.functor._
 import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.{nxv, schemas}
 import ch.epfl.bluebrain.nexus.delta.sdk.AclResource
 import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceRef.Latest
-import ch.epfl.bluebrain.nexus.delta.sdk.model.acls.Target.{Organization, Project, Root}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.acls.Target.{
```
bogdanromanx, line 7: is there anyway to configure scalafmt to replace these with wildcard imports?

bogdanromanx (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Target.scala
umbreak: It is easier for me as I did it. In listings you can pass either one or another, so there is a reason to have a super trait for both.

umbreak: Target is a superclass for `TargetLocation` and `TargetFilterLocation`, so you can select whatever you want :)

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/AclAddressFilter.scala
```diff
@@ -0,0 +1,89 @@
...
+
+object AclAddressFilter {
+  type AnyOrganization           = AnyOrganization.type
+  type AnyOrganizationAnyProject = AnyOrganizationAnyProject.type
+
+  /**
+    * An ACL filtering that matches any [[AclAddress.Organization]] address
+    */
+  final case object AnyOrganization extends AclAddressFilter {
+    val string = s"/*"
```
bogdanromanx, line 39: no need for the s here.

bogdanromanx (approved):

PR #1643, opened 2020-09-25
Added ResourceF JsonLdEncoder

umbreak:


delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/JsonLd.scala
```diff
@@ -119,7 +119,7 @@ object JsonLd {
     * @throws IllegalArgumentException when the provided ''expanded'' json does not match the expected value
     */
   final def expandedUnsafe(expanded: Json, rootId: IriOrBNode): ExpandedJsonLd =
-    expanded.asArray.flatMap(headOnlyOptionOr(_)(Json.obj())).flatMap(_.asObject) match {
+    expanded.asArray.flatMap(_.singleEntryOr(Json.obj())).flatMap(_.asObject) match {
```
imsdu, line 122: Better name

umbreak: Yeah, I don't know which better name. `headOnlyOptionOr` seemed too long. I thought `sequence.singleEntryOr(defaultValue)` was clear enough. But if you have another suggestion I'll change it.

imsdu (approved):

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/JsonLd.scala
imsdu: Sorry, I meant this is a good name :)

PR #1644, opened 2020-09-28
Added support for ordering Json keys

umbreak:


imsdu (dismissed):

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/utils/OrderingJsonKeys.scala
```diff
@@ -0,0 +1,37 @@
+package ch.epfl.bluebrain.nexus.delta.rdf.utils
+
+/**
+  * Ordering for Json keys.
+  */
+trait OrderingJsonKeys extends Ordering[String]
```
bogdanromanx, line 6: If it's not too much trouble, I would recommend renaming this to JsonKeyOrdering.

bogdanromanx (approved):

bogdanromanx (approved):

PR #1647, opened 2020-09-29
Implemented service permisssions

bogdanromanx:
Fixes: #1516

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -39,6 +39,7 @@ val flywayVersion                   = "6.5.2"
 val fs2Version                      = "2.4.4"
 val guavaVersion                    = "29.0-jre"
 val http4sVersion                   = "0.21.7"
+val h2Version                       = "1.4.200"
 val iamVersion                      = "1.3.0"
 val jenaVersion                     = "3.15.0"
 val jsonldjavaVersion               = "0.13.2"
@@ -94,6 +95,7 @@ lazy val akkaPersistenceQuery    = "com.typesafe.akka"  %% "akka-persistence-que
 lazy val akkaSlf4j               = "com.typesafe.akka"  %% "akka-slf4j"                          % akkaVersion
 lazy val akkaStream              = "com.typesafe.akka"  %% "akka-stream"                         % akkaVersion
 lazy val akkaTestKit             = "com.typesafe.akka"  %% "akka-testkit"                        % akkaVersion
+lazy val akkaTestKitTyped        = "com.typesafe.akka"  %% "akka-actor-testkit-typed"            % akkaVersion
 lazy val alleycatsCore           = "org.typelevel"      %% "alleycats-core"                      % catsVersion
 lazy val alpakkaFiles            = "com.lightbend.akka" %% "akka-stream-alpakka-file"            % alpakkaVersion
 lazy val alpakkaSse              = "com.lightbend.akka" %% "akka-stream-alpakka-sse"             % alpakkaVersion
@@ -123,6 +125,7 @@ lazy val dockerTestKit = Seq(
 
 lazy val fs2           = "co.fs2"                     %% "fs2-core"                % fs2Version
 lazy val flyway        = "org.flywaydb"                % "flyway-core"             % flywayVersion
+lazy val h2            = "com.h2database"              % "h2"                      % h2Version
 lazy val http4sCirce   = "org.http4s"                 %% "http4s-circe"            % http4sVersion
 lazy val http4sClient  = "org.http4s"                 %% "http4s-blaze-client"     % http4sVersion
 lazy val http4sDsl     = "org.http4s"                 %% "http4s-dsl"              % http4sVersion
@@ -390,8 +393,14 @@ lazy val service = project
   .dependsOn(sourcing, rdf, sdk, sdkTestkit, testkit % "test->compile", sdkTestkit % "test->compile")
   .settings(
     libraryDependencies ++= Seq(
-      scalaTest % Test
-    )
+      akkaSlf4j        % Test,
+      akkaTestKitTyped % Test,
+      h2               % Test,
+      kryo             % Test,
+      logback          % Test,
+      scalaTest        % Test
+    ),
+    Test / fork          := true
   )
 
 lazy val app = project
--- a/build.sbt
+++ b/build.sbt
@@ -315,13 +315,13 @@ lazy val sourcing = project
       doobiePostgres,
       fs2,
       flyway,
+      kryo,
       monixBio,
       scalaLogging,
       streamz,
       akkaActorTyped         % Test,
       akkaPersistenceTestKit % Test,
       akkaSlf4j              % Test,
-      kryo                   % Test,
       logback                % Test
     ) ++ akkaPersistenceJdbc,
     Test / fork          := true
@@ -396,7 +396,6 @@ lazy val service = project
       akkaSlf4j        % Test,
       akkaTestKitTyped % Test,
       h2               % Test,
-      kryo             % Test,
       logback          % Test,
       scalaTest        % Test
     ),
--- a/delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/PermissionsBehaviours.scala
+++ b/delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/PermissionsBehaviours.scala
@@ -1,0 +1,209 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.testkit
+
+import java.time.Instant
+
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.{nxv, schemas}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceRef.Latest
+import ch.epfl.bluebrain.nexus.delta.sdk.model.identities.Identity
+import ch.epfl.bluebrain.nexus.delta.sdk.model.identities.Identity.{Anonymous, Subject}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions.PermissionsRejection._
+import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions.{Permission, PermissionSet}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{Label, ResourceF}
+import ch.epfl.bluebrain.nexus.delta.sdk.testkit.PermissionsBehaviours._
+import ch.epfl.bluebrain.nexus.delta.sdk.{Permissions, PermissionsResource}
+import ch.epfl.bluebrain.nexus.testkit.TestHelpers.genString
+import ch.epfl.bluebrain.nexus.testkit.{IOFixedClock, IOValues}
+import monix.bio.{IO, Task}
+import monix.execution.Scheduler
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+import scala.concurrent.TimeoutException
+import scala.concurrent.duration._
+
+/**
+  * The collection of behaviours for permissions.
+  */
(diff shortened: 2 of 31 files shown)


delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/processor/ShardedAggregate.scala
```diff
@@ -17,12 +17,14 @@ import scala.reflect.ClassTag
...
   * (State, Command, Event, Rejection)
   *
+  * @param eventLog      access to the persistent event log
   * @param entityTypeKey the key of an entity type, unique
   * @param clusterSharding the sharding for this aggregate
   * @param retryStrategy the retry strategy to adopt
   * @param askTimeout the ask timeout
   */
 private[processor] class ShardedAggregate[State, Command, Event, Rejection](
+    eventLog: EventLog[Event],
```
imsdu, line 27: I was seeing Aggregate and eventLog as two separate things but why not, if it is easier to use that way

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/permissions/PermissionsImpl.scala
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  /**
+    * Constructs a new [[Permissions]] instance backed by a sharded aggregate.
+    *
+    * @param minimum the minimum collection of permissions
+    * @param agg     the permissions aggregate
+    * @param base    the base uri of the system API
+    */
+  final def apply(minimum: Set[Permission], agg: PermissionsAggregate, base: Uri): Permissions =
```
umbreak, line 134: Use `BaseUri` instead of `Uri`

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/permissions/PermissionsImpl.scala
```diff
@@ -0,0 +1,155 @@
...
+    *  a cluster. The implementation is effectful as it allocates a new cluster sharding entity across the cluster.
+    *
+    * @param minimum         the minimum collection of permissions
+    * @param base            the base uri of the system API
+    * @param aggregateConfig the aggregate configuration
+    * @param eventLog        an event log instance for events of type [[PermissionsEvent]]
+    */
+  final def apply(
+      minimum: Set[Permission],
+      base: Uri,
```
umbreak, line 148: Use `BaseUri` instead of `Uri`

umbreak (changes requested):

delta/service/src/test/resources/permissions-test.conf
```diff
@@ -0,0 +1,14 @@
+include "akka-test.conf"
+
+akka {
+  actor {
+    serialization-bindings {
```
umbreak, line 5: This is duplicated, already present in `akka-test.conf`

delta/service/src/test/resources/akka-test.conf
```diff
@@ -0,0 +1,67 @@
...
+  actor {
+    provider = "akka.cluster.ClusterActorRefProvider"
+    warn-about-java-serializer-usage = off
+
+    serializers {
+      kryo = "io.altoo.akka.serialization.kryo.KryoSerializer"
+    }
+
+    serialization-bindings {
+      "ch.epfl.bluebrain.nexus.delta.sdk.model.permissions.PermissionsCommand"   = "kryo"
```
umbreak, line 11: Is there some better way for kryo serialization instead of this config file? Annotations, marker trait?

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/permissions/PermissionsImpl.scala
umbreak, line 148: `BaseUri` instead of `Uri`

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/permissions/PermissionsImpl.scala
```diff
@@ -0,0 +1,155 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.{Permissions, PermissionsResource}
+import ch.epfl.bluebrain.nexus.delta.service.permissions.PermissionsImpl.{entityId, entityType, PermissionsAggregate}
+import ch.epfl.bluebrain.nexus.sourcing._
+import ch.epfl.bluebrain.nexus.sourcing.processor.{AggregateConfig, ShardedAggregate, StopStrategy}
+import monix.bio.{IO, UIO}
+
+final class PermissionsImpl private (
+    override val minimum: Set[Permission],
+    agg: PermissionsAggregate,
+    base: Uri
```
umbreak, line 20: `BaseUri` instead of `Uri`

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/permissions/PermissionsImpl.scala
umbreak, line 134: `BaseUri` instead of `Uri`

umbreak (changes requested):

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/processor/ShardedAggregate.scala
imsdu: Ok to add eventLog in there but maybe we can find a more relevant name ?

build.sbt
```diff
@@ -39,6 +39,7 @@ val flywayVersion                   = "6.5.2"
 val fs2Version                      = "2.4.4"
 val guavaVersion                    = "29.0-jre"
 val http4sVersion                   = "0.21.7"
+val h2Version                       = "1.4.200"
```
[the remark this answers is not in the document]
bogdanromanx: It doesn't provide the ability to replay the log (no persistence query api)

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/processor/ShardedAggregate.scala
bogdanromanx: the type is EventLog.... so it seemed appropriate. Do you have a better suggestion?

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/PermissionsBehaviours.scala
```diff
@@ -0,0 +1,209 @@
...
+    "return none for unknown rev" in {
+      permissions.accepted.fetchAt(9999L).rejected shouldEqual RevisionNotFound(9999L, 5L)
+    }
+  }
+
+}
+
+object PermissionsBehaviours {
+
+  import ch.epfl.bluebrain.nexus.delta.sdk.Permissions.{schemas, _}
```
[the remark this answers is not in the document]
bogdanromanx: yes, it conflicts with the global import:
`import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.{nxv, schemas}`

delta/sdk-testkit/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/PermissionsDummySpec.scala
```diff
@@ -1,164 +1,22 @@
...
-  val perm3: Permission = Permission.unsafe(genString())
-  val perm4: Permission = Permission.unsafe(genString())
-
-  val epoch: Instant            = Instant.EPOCH
-  implicit val subject: Subject = Identity.User("user", Label.unsafe("realm"))
-
-  implicit val scheduler: Scheduler = Scheduler.global
+class PermissionsDummySpec
+    extends AnyWordSpecLike
+    with PermissionsBehaviours
```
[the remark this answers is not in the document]
bogdanromanx: the PermissionsBehaviours doesn't extend anything, it just specifies its dependencies:
```
trait PermissionsBehaviours { this: AnyWordSpecLike with Matchers with IOValues with IOFixedClock =>
```
I think this is fine, or the alternative would be to extend from PermissionsBehaviours (although not ideal)

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/permissions/PermissionsImpl.scala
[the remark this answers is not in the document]
bogdanromanx: good point, I'll make use of it

delta/service/src/test/resources/permissions-test.conf
bogdanromanx: it needs to be in this file, I'll remove it from the other

delta/service/src/test/resources/akka-test.conf
bogdanromanx: I think it's a good idea to keep these explicit. One needs to know how things are serialised, especially events.

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/processor/ShardedAggregate.scala
imsdu: I was thinking more about Aggregate/ShardedAggregate but I have nothing coming in mind so it may be ok :)

delta/service/src/test/resources/akka-test.conf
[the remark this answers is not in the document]
bogdanromanx: I'm not fond of the idea for two reasons:
- the sdk should not have any information about methods of serialisation
- one needs to control how serialisation happens (e.g. we want to serialise PermissionsEvent as json but the rest with kryo); adding a marker trait would make the SDK module responsible of the implementation, which would be incorrect.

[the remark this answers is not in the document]
imsdu: We can't have an Event trait and use it as a marker ?

And there is a resolution strategy when they are several candidates so we can still use a different serialization for Permissions: https://doc.akka.io/docs/akka/current/serialization.html?language=scala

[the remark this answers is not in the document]
bogdanromanx: We can have a global Event trait, but it won't help much... because you'd have to have a serialiser for it that knows how to serialise instances of that type. So it would be part of the serialisation logic in one place, and the rest in another. Additionally, you'd still have to create a different one for the Command, State and Rejection... but this is bad for the reasons mentioned above.

The serialisation config is a map... so there's no resolution strategy for it...:
```
  serialization-bindings {
    "docs.serialization.JsonSerializable" = jackson-json
    "docs.serialization.CborSerializable" = jackson-cbor
    "com.google.protobuf.Message" = proto
    "docs.serialization.MyOwnSerializable" = myown
  }
```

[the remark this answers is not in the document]
imsdu: I was talking about this:
```You only need to specify the name of a trait or abstract base class of the messages. In case of ambiguity, i.e. the message implements several of the configured classes, the most specific configured class will be used, i.e. the one of which all other candidates are superclasses. If this condition cannot be met, because e.g. two marker traits that have been configured for serialization both apply and neither is a subtype of the other, a warning will be issued.```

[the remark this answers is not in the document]
Comment (bogdanromanx) [reply delta/service/src/test/resources/akka-test.conf]: Are you suggesting to create two non-sealed traits, like so:
```
trait Message extends Product with Serializable
trait Event extends Message
```
... and mix them in all SDK types? (commands, states, rejections and events)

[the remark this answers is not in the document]
bogdanromanx: If we modify sourcing to wrap the state in a case class, we can avoid having to mix in anything for commands, states and rejections.
And we can create a marker trait for Events.

umbreak (approved):

imsdu (approved):

PR #1659, opened 2020-09-30
Expose the permissions events through the Permissions contract

bogdanromanx:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala
```diff
@@ -3,18 +3,20 @@ package ch.epfl.bluebrain.nexus.delta.sdk
...
 import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions._
-import monix.bio.{IO, UIO}
+import fs2.Stream
+import monix.bio.{IO, Task, UIO}
 
 /**
   * Operations pertaining to managing permissions.
   */
-trait Permissions {
+trait Permissions[Offset] {
```
umbreak, line 19: It has to be a generic type? Cannot be an abstract type? We are creating a lot of unnecessary noise having it as a type parameter (like `Permissions[_]` or `Permissions[Nothing]`)

umbreak (changes requested):
Not convinced about the offset as generic type

umbreak (approved):

wwajerowicz (approved):

umbreak (changes requested):
I'm still not sure if this generic type is worth it. Did you try wrapping the events response in an IO. would that work?

PR #1662, opened 2020-10-01
Move integration tests to main repo

### LINKED ISSUE #1648: Port the integration tests to the BlueBrain/nexus repo
- create a new module `/tests` to host the integration tests
- create a github workflow to run static analysis, scalafmt, test:compile during review
- create a github workflow to execute the tests during publish; find a means to notify the entire team when the tests fail (could be an email config, slack message)
- remove all dependencies on previous nexus modules (nexus-commons, nexus-rdf etc.)

imsdu:
Fixes #1648


.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,24 @@
+name: Continuous Integration RDF
```
umbreak, line 1: wrong name

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/CirceEq.scala
```diff
@@ -0,0 +1,38 @@
...
+  implicit private val printer: Printer = Printer.noSpaces.copy(dropNullValues = true)
+
+  def equalIgnoreArrayOrder(json: Json): IgnoredArrayOrder = IgnoredArrayOrder(json)
+
+  case class IgnoredArrayOrder(json: Json) extends Matcher[Json] {
+    private def sortKeys(value: Json): Json = {
+      def canonicalJson(json: Json): Json =
+        json.arrayOrObject[Json](
+          json,
+          arr => Json.fromValues(arr.sortBy(_.hashCode()).map(canonicalJson)),
```
umbreak, line 17:
```scala
arr => Json.fromValues(arr.map(canonicalJson).sortBy(_.hashCode))
```

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/TestHelpers.scala
```diff
@@ -84,8 +89,8 @@ trait TestHelpers {
     * @return the content of the referenced resource as a json value
     */
   @SuppressWarnings(Array("TryGet"))
-  final def jsonContentOf(resourcePath: String, replacements: Map[String, String]): Json =
-    parse(contentOf(resourcePath, replacements)).toTry.get
+  final def jsonContentOf(resourcePath: String, attributes: Map[String, Any]): Json =
```
umbreak, line 92: We can add another jsonContentOf

```scala
final def jsonContentOf(resourcePath: String, attributes: (String, Any)*): Json =
    jsonContentOf(resourcePath, attributes.toMap)
```

And refactor in the places we use `jsonContentOf` to use this instead. Honestly wrapping it into a Map doesn't add any extra value :)

tests/src/main/resources/application.conf
```diff
@@ -0,0 +1,22 @@
...
+      max-open-requests = 64
+    }
+  }
+
+  loggers = ["akka.event.slf4j.Slf4jLogger"]
+  logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
+  log-dead-letters = off
+  stdout-loglevel = "OFF"
+  log-dead-letters-during-shutdown = off
+  loglevel = INFO
```
umbreak, line 19: Do we really want the integration tests at this loglevel by default? WARN or ERROR would be better I guess

umbreak (changes requested):

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/TestHelpers.scala
imsdu: In most cases yes but sometimes when we reuse common attributes, we can concat maps for example

imsdu: I have added the second method though so we can use it too when needed

bogdanromanx (approved):

PR #1713, opened 2020-10-20
Extract Acls module

### LINKED ISSUE #1517: Extract the Acls module, refactor it to implement the new contract and use the new RDF api

wwajerowicz:
Fixes #1517


umbreak (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
```diff
@@ -0,0 +1,149 @@
...
+  ): IO[AclRejection, AclResource] = eval(DeleteAcl(address, rev, caller))
+
+  private def eval(cmd: AclCommand): IO[AclRejection, AclResource] =
+    for {
+      evaluationResult <- agg.evaluate(cmd.address.string, cmd).mapError(_.value)
+      resource         <- IO.fromOption(
+                            evaluationResult.state.toResource,
+                            UnexpectedInitialState(cmd.address)
+                          )
+      _                <- index.put(cmd.address, resource)
```
umbreak, line 78: So are we populating the cache synchronously here instead of how we did it before reading events from the primary store asynchronously?

It could be a good option but worth it to discussion. Maybe tomorrow morning meeting?

umbreak (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
wwajerowicz: That's what we do currently as well. We update the cache on change to propagate the changes faster, but we also update it from primary store events to rebuild the cache in case of restarts.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
```diff
@@ -0,0 +1,149 @@
...
+    extends Acls {
+
+  override def fetch(address: AclAddress): UIO[Option[AclResource]] = agg.state(address.string).map(_.toResource)
+
+  override def fetchAt(address: AclAddress, rev: Long): IO[AclRejection.RevisionNotFound, Option[AclResource]] =
+    if (rev == 0L) UIO.pure(None)
+    else
+      eventLog
+        .currentEventsByPersistenceId(
+          s"$entityType-${URLEncoder.encode(address.string, "UTF-8")}",
```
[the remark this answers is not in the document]
imsdu: Yep, we should have a trait object but I would vote for putting it in sourcing which should be responsible for computing persistence ids.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
umbreak: Ah ok. So that's the exception.
So where is the code that populates the cache from the primary store events? I guess it should be present on the method:
```scala
 private def cache(aclsConfig: AclsConfig)(implicit as: ActorSystem[Nothing]): AclsCache = ???
```

That, or making the `AclsCache` a class/object and do the populate of acl resources there.

umbreak (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
imsdu: Populating the cache incrementally from calling fetch could also be an option no ?
So that only the "hot" acls find themselves in cache.

wwajerowicz: I think we need to have all the ACLs in the cache, since we use the cache for listings.

build.sbt
```diff
@@ -388,7 +388,7 @@ lazy val service = project
     moduleName := "delta-service"
   )
   .settings(shared, compilation, assertJavaVersion, coverage, release)
-  .dependsOn(sourcing, rdf, sdk, sdkTestkit % "test->compile;test->test", testkit % "test->compile")
+  .dependsOn(sourcing, rdf, sdk, sourcing, sdkTestkit % "test->compile;test->test", testkit % "test->compile")
```
bogdanromanx: double sourcing dep

delta/service/src/test/resources/akka-test.conf
```diff
@@ -25,7 +25,8 @@ akka {
       "ch.epfl.bluebrain.nexus.delta.sdk.model.Event" = "kryo"
       "ch.epfl.bluebrain.nexus.delta.service.identity.IdentitiesImpl$FetchGroups" = "kryo"
       "ch.epfl.bluebrain.nexus.delta.service.cache.SubscriberCommand$Unsubscribe$" = "kryo"
-      "ch.epfl.bluebrain.nexus.delta.service.cache.KeyValueStoreSpec$RevisionedValue" = "kryo"
+      "ch.epfl.bluebrain.nexus.delta.service.cache.KeyValueStoreSpec$RevisionedValue" = "kryo",
```
bogdanromanx, line 28: why the comma here?

build.sbt
imsdu, line 391: Sourcing appears twice

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
```diff
@@ -0,0 +1,187 @@
...
+    )
+  }
+
+  /**
+    * Constructs an [[AclsImpl]] instance.
+    *
+    * @param agg      the sharded aggregate
+    * @param eventLog the event log
+    * @param cache    the cache
+    * @return
```
imsdu, line 160: to remove

imsdu (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
```diff
@@ -0,0 +1,187 @@
...
+  final def apply(
+      config: AclsConfig,
+      permissions: UIO[Permissions],
+      eventLog: EventLog[Envelope[AclEvent]]
+  )(implicit
+      as: ActorSystem[Nothing],
+      clock: Clock[UIO]
+  ): UIO[AclsImpl] = {
+    val index = cache(config)
+    aggregate(permissions, config.aggregate).map { agg =>
```
bogdanromanx, line 180: I would flatMap and call startIndexing in a UIO.delay

bogdanromanx (changes requested):

umbreak (dismissed):

bogdanromanx (approved):

imsdu (approved):

PR #1726, opened 2020-10-26
Fixed issue with evaluation happening before initial state retrieved

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1727


delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/processor/ProcessorCommand.scala
```diff
@@ -44,6 +44,9 @@ object ProcessorCommand {
   sealed trait ReadonlyCommand                                                           extends EventSourceCommand
   final case class RequestState[State](id: String, replyTo: ActorRef[StateReply[State]]) extends ReadonlyCommand
   final case class RequestLastSeqNr(id: String, replyTo: ActorRef[LastSeqNr])            extends ReadonlyCommand
+  final case class RequestStateInternal[State](id: String, replyTo: ActorRef[ResponseStateInternal[State]])
```
imsdu, line 47: We should make them private

bogdanromanx (approved):

PR #1729, opened 2020-10-27
Added StopStrategyConfig and SnapshotStrategyConfig

umbreak:


delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/config/AggregateConfig.scala
```diff
@@ -0,0 +1,34 @@
...
+    askTimeout: Timeout,
+    evaluationMaxDuration: FiniteDuration,
+    evaluationExecutionContext: ExecutionContext,
+    stashSize: Int
+) {
+
+  /**
+    * The underlying processor ''AggregateConfig''
+    */
+  def processor: AggregateProcessorConfig =
```
imsdu, line 32: We could have in as subconfiguration as the others

imsdu (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/config/AggregateConfig.scala
umbreak: That complicates things for pureconfig reader, making it more verbose. I don't think it is worth it.

In the app.conf this is - and should be - on the top level

```
    aggregate {
      snapshot-strategy = ${app.defaults.snapshot-every-500}
      stop-strategy = ${app.defaults.passivate-never}
      ask-timeout = 15 seconds
      evaluation-max-duration = 3 seconds
      stash-size = 10
    }
```

That resembles to the case class

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/SnapshotStrategyConfig.scala
```diff
@@ -0,0 +1,52 @@
...
+/**
+  * A Snapshot configuration that will be made every numberOfEvents and keepNSnapshots will be kept
+  * deleteEventsOnSnapshot allows to delete events prior which are older than
+  * the older snapshot
+  *
+  * @param numberOfEvents         the optional frequency we need to trigger snapshots
+  * @param keepNSnapshots         the optional number of snapshots we need to keep
+  * @param deleteEventsOnSnapshot the optional flag to decide if we need to delete old events
+  */
+final case class SnapshotStrategyConfig private (
```
imsdu, line 14: I would split the config into an ADT.

And are there cases where we don't want to snapshot every now and then ? It would be even simpler

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/processor/AggregateProcessorConfig.scala
```diff
@@ -6,14 +6,14 @@ import scala.concurrent.ExecutionContext
...
-  * Aggregate configuration.
+  * Aggregate event source processor configuration.
   *
   * @param askTimeout                 timeout for the message exchange with the aggregate actor
   * @param evaluationMaxDuration      timeout for evaluating commands
   * @param evaluationExecutionContext the execution context where commands are to be evaluated
   * @param stashSize                  the maximum size allowed for stashing when evaluating
   */
-final case class AggregateConfig(
+final case class AggregateProcessorConfig(
```
imsdu, line 16: EventSourceProcessorConfig

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/SnapshotStrategyConfig.scala
umbreak: Configurations should be simple and they should not mirror the actual thing. The ADT just adds complexity in the pureconfig reader side.

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/processor/StopStrategyConfig.scala
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+  * The configuration for a [[StopStrategy]]
+  *
+  * @param lapsedSinceLastInteraction   Some(duration) if the actor should stop after no new messages are received in the ''duration'' interval;
+  *                                     None to keep the actor alive
+  * @param lapsedSinceRecoveryCompleted Some(duration) if the actor should stop (and passivate);
+  *                                     None to keep the actor alive (and no passivation)
+  */
+final case class StopStrategyConfig(
```
imsdu, line 15: ADT ?

umbreak: No ADTs on configuration. It is just gonna be more difficult to generate a pureconfig reader.

delta/app/src/main/resources/app.conf
```diff
@@ -189,6 +191,46 @@ app {
...
+
+    # never snapshot
+    snapshot-never {
+      number-of-events = null
+      keep-snapshots = null
+      delete-events-on-snapshot = null
+    }
+
+    # Snapshot after 500 events consumed
+    snapshot-every-500 {
```
imsdu, line 207: I would call it snapshot-default as we can still override its properties

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/config/AggregateConfig.scala
imsdu: You can do it without modifying the conf by slightly modilfying your ConfigReader

umbreak: what's the benefit?

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/SnapshotStrategyConfig.scala
imsdu: You get at least part of the complexity here.

And I think we should fail if one of the property is not defined, not falling back to NoSnapshot

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/config/AggregateConfig.scala
imsdu: It is a little more readable as you have the four fields embedded in the processorConfig so you understand it is not used for something else

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/processor/StopStrategyConfig.scala
imsdu: Yes but it could be nice to see directly in the conf that the never is expressed in some way in the conf other than a reference

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/SnapshotStrategyConfig.scala
umbreak: @imsdu this is already the case, see the companion object (notice that class is private, so it will be constructed through the companion object)

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/processor/StopStrategyConfig.scala
umbreak: StopStrategyConfig has no restrictions. The configuration does not decide if it is a transient or a persistent stopStrategy, that is decide by us in the code. So it is always correct to have both parameters as optional.

Notice that in the case of `SnapshotStrategyConfig` I dealt with corner cases by having the case class private and the companion object constructor deal with the exceptions. This approach provides the safety that no wrong arguments can be passed, while keeping it simple so that pureconfig reader can be built easily

bogdanromanx (approved):

PR #1740, opened 2020-10-30
Add projects routes

imsdu:


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala
```diff
@@ -0,0 +1,143 @@
...
+                  delete {
+                    parameter("rev".as[Long]) { rev => completeIO(projects.deprecate(ref, rev).map(_.void)) }
+                  }
+                )
+              }
+            },
+            (uuid & uuid & pathEndOrSingleSlash) { (orgUuid, projectUuid) =>
+              operationName(s"$prefixSegment/project/{orgUuid}/{projectUuid}") {
+                get {
+                  def checkOrg(p: Option[ProjectResource]): IO[ProjectRejection, Option[ProjectResource]] =
```
umbreak, line 109: Move this logic to `Projects.scala`. We should have a method there as follows:
```scala
def fetch(orgUuid: UUID, projectUuid: UUID): IO[ProjectNotFound, Option[ProjectResource]] =
fetch(projectUuid).flatMap {
case Some(res) if res.value.organizationUuid != orgUuid=> IO.raiseError(ProjectNotFound(orgUuid, projectUuid))
case other => IO.pure(other)
```

umbreak (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala
```diff
@@ -0,0 +1,143 @@
...
+      }
+    }
+
+  def routes: Route =
+    baseUriPrefix(baseUri.prefix) {
+      extractSubject { implicit subject =>
+        pathPrefix("projects") {
+          concat(
+            // List projects
+            (get & extractUri & paginated & projectsSearchParams & pathEndOrSingleSlash) { (uri, pagination, params) =>
```
bogdanromanx, line 60: I would recommend reordering the directives such that the first two are `get` and `pathEndOrSingleSlash` to avoid computing the others if there's no match

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala
```diff
@@ -0,0 +1,143 @@
...
+import kamon.instrumentation.akka.http.TracingDirectives.operationName
+import monix.bio.IO
+import monix.execution.Scheduler
+
+/**
+  * The project routes
+  * @param identities the identity module
+  * @param projects   the projects module
+  */
+final class ProjectsRoutes private (identities: Identities, projects: Projects)(implicit
```
bogdanromanx, line 29: if the constructor is public and paginationConfig is replaced with the ProjectsConfig the wiring can be done only using the constructor

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectEvent.scala
```diff
@@ -118,4 +131,46 @@ object ProjectEvent {
...
+      case "uuid"              => nxv.uuid.prefix
+      case "organizationLabel" => nxv.organizationLabel.prefix
+      case "organizationUuid"  => nxv.organizationUuid.prefix
+      case "rev"               => nxv.rev.prefix
+      case "instant"           => nxv.instant.prefix
+      case "subject"           => nxv.eventSubject.prefix
+      case other               => other
+    })
+
+  implicit val apiMappingEncoder: Encoder[Map[String, Iri]] =
```
bogdanromanx, line 150: make this private

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectEvent.scala
```diff
@@ -118,4 +131,46 @@ object ProjectEvent {
       instant: Instant,
       subject: Subject
   ) extends ProjectEvent
+
+  private val context = ContextValue(contexts.resource, contexts.projects)
+
+  implicit private[projects] val config: Configuration = Configuration.default
```
bogdanromanx, line 137: do we need this to be scoped to the package?

bogdanromanx (changes requested):
Missing the wiring

umbreak (dismissed):

bogdanromanx (dismissed):

bogdanromanx (approved):

PR #1743, opened 2020-11-02
Updated event serialization to use circe json

### LINKED ISSUE #1725: Configure app event serialization
The current default is the use of kryo of persisting events. Json is more appropriate as it's readable and can facilitate migrations.

bogdanromanx:
Fixes #1725


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/serialization/EventSerializer.scala
```diff
@@ -0,0 +1,112 @@
+package ch.epfl.bluebrain.nexus.delta.serialization
```
imsdu, line 1: It should be more in the service module, where we can also test it at the same time as the module impls no ?

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/serialization/EventSerializer.scala
```diff
@@ -0,0 +1,112 @@
...
+
+  implicit final private val aclAddressEncoder: Encoder[AclAddress] = Encoder.encodeString.contramap(_.string)
+  implicit final private val aclAddressDecoder: Decoder[AclAddress] = Decoder.decodeString.emap { str =>
+    AclAddress.fromString(str).leftMap(_.getMessage)
+  }
+
+  final private case class AclEntry(identity: Identity, permissions: Set[Permission])
+  final private val aclEntryCodec                     = deriveConfiguredCodec[AclEntry]
+  implicit final private val aclEncoder: Encoder[Acl] =
+    Encoder.encodeList(aclEntryCodec).contramap(acl => acl.value.toList.map { case (id, perms) => AclEntry(id, perms) })
```
imsdu, line 91: Encoding it as a map is a problem ?

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/serialization/EventSerializer.scala
```diff
@@ -0,0 +1,112 @@
...
+  implicit final private val aclDecoder: Decoder[Acl] = Decoder.decodeList[AclEntry](aclEntryCodec).map { entries =>
+    Acl(entries.map(e => e.identity -> e.permissions).toMap)
+  }
+
+  implicit final val uriEncoder: Encoder[Uri] = Encoder.encodeString.contramap(_.toString)
```
imsdu, line 96: We dont' want to use the ones already defined ?

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/serialization/EventSerializer.scala
umbreak: Yeah, just import `ch.epfl.bluebrain.nexus.delta.sdk.implicits._` and you'll have those

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/serialization/EventSerializer.scala
```diff
@@ -0,0 +1,112 @@
...
+    val string = new String(bytes, StandardCharsets.UTF_8)
+    decode[E](string)
+      .getOrElse(throw new IllegalArgumentException(s"Unable to decode for manifest '$manifest' event '$string'"))
+  }
+}
+
+@nowarn("cat=unused")
+object EventSerializer {
+
+  final val permissionsEventManifest: String  = "permissions"
```
umbreak, line 71: Why don't you use the names from the modules instead of defining those hardcoded strings here?
`Permissions.moduleType`
`Realms.moduleType`
...

umbreak (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/serialization/EventSerializer.scala
bogdanromanx: you'd need a KeyEncoder / KeyDecoder for that, which yields the same amount of code and makes the json format different than how we serialize the events in the API. It's not worth it.

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/serialization/EventSerializer.scala
bogdanromanx: yep

imsdu (approved):

imsdu (approved):

PR #1749, opened 2020-11-03
Added SHACL support on RDF module

umbreak:


delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/shacl/ValidationReport.scala
```diff
@@ -0,0 +1,60 @@
...
+  /**
+    * @param ignoreTargetedNodes flag to decide whether or not ''targetedNodes''
+    *                         should be ignored from the validation logic
+    * @return true if the validation report has been successful or false otherwise
+    */
+  def isValid(ignoreTargetedNodes: Boolean = false): Boolean =
+    (ignoreTargetedNodes && conforms) || (!ignoreTargetedNodes && targetedNodes > 0 && conforms)
+}
+
+object ValidationReport extends TestHelpers {
```
imsdu, line 33: TestHelpers in the sources ?

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/shacl/ShaclEngine.scala
```diff
@@ -0,0 +1,108 @@
...
+  * Extend the [[ValidationEngine]] form TopQuadrant in order to add triples to the report
+  * with the number of targetedNodes
+  *
+  * @param dataset        the Dataset to operate on
+  * @param shapesGraphURI the URI of the shapes graph (must be in the dataset)
+  * @param shapesGraph    the ShapesGraph with the shapes to validate against
+  */
+// $COVERAGE-OFF$
+@SuppressWarnings(Array("NullParameter"))
+final class ShaclEngine private (dataset: Dataset, shapesGraphURI: URI, shapesGraph: ShapesGraph)
```
imsdu, line 29: It would be nice to have tests on it also to see how it is being used

build.sbt
```diff
@@ -332,7 +334,7 @@ lazy val testPlugin = project
 
 lazy val rdf = project
   .in(file("delta/rdf"))
-  .dependsOn(testkit % "test->compile")
+  .dependsOn(testkit)
```
imsdu, line 337: We should keep the scope here

imsdu, line 337: We should keep the scope here

imsdu (changes requested):

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/shacl/ShaclEngine.scala
umbreak: This code does not need to be tested because it is a third party code which is tested by topQuadrant itself.
You'll have examples of this on `Resources` or `Schemas` afterwards

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/ClasspathResourceUtils.scala
```diff
@@ -0,0 +1,80 @@
+package ch.epfl.bluebrain.nexus.delta.kernel.utils
+
+import scala.io.{Codec, Source}
+import io.circe.Json
+import io.circe.parser.parse
+import org.fusesource.scalate.TemplateEngine
+
+// $COVERAGE-OFF$
+trait ClasspathResourceUtils {
```
bogdanromanx, line 9: We should probably test these, to immediately identify when we have problems.
Additionally, since the code is used in "compile" scope and it's side effecting, I strongly recommend changing the signatures to return `Task[A]`; this would also allow for memoization.

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/Vocabulary.scala
```diff
@@ -123,6 +145,11 @@ object Vocabulary {
     val updatedBy             = Metadata("updatedBy")
     val userInfoEndpoint      = Metadata("userInfoEndpoint")
     val uuid                  = Metadata("uuid")
+    val resourceId            = Metadata("resourceId")
```
bogdanromanx, line 148: we should probably sort these alphabetically... they will be easier to find :)

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/shacl/ShaclEngine.scala
```diff
@@ -0,0 +1,108 @@
...
+    extends ValidationEngine(dataset, shapesGraphURI, shapesGraph, null) {
+  private var targetedNodes = 0
+
+  override def validateNodesAgainstConstraint(focusNodes: util.Collection[RDFNode], constraint: Constraint): Unit = {
+    super.validateNodesAgainstConstraint(focusNodes, constraint)
+    targetedNodes += 1
```
bogdanromanx, line 35: this seems to count the constraints not the nodes

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/shacl/ShaclEngine.scala
bogdanromanx: I think it's worth adding a test to make sure that the local code will continue to function when doing lib updates

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/utils/JsonUtils.scala
```diff
@@ -64,6 +64,24 @@ trait JsonUtils {
   def remoteAllKeys(json: Json, keys: String*): Json =
     removeNested(json, keys.map(k => (kk => kk == k, _ => true)))
 
+  /**
+    * Replace in the passed ''json'' the found key value pairs in ''from'' with the value in ''toValue''
+    */
+  def replace(json: Json, from: (String, Json), toValue: Json): Json = {
+    def inner(obj: JsonObject): JsonObject =
```
bogdanromanx, line 71: `val (fromKey, fromValue) = from` would make the pattern match below more readable

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/shacl/ShaclEngine.scala
```diff
@@ -0,0 +1,108 @@
...
+    SHACLFunctions.registerFunctions(finalShapesModel)
+    applySkipShapesCheck(dataModel, finalShapesModel, validateShapes, reportDetails)
+  }
+
+  private def applySkipShapesCheck(
+      dataModel: Model,
+      finalShapesModel: Model,
+      validateShapes: Boolean,
+      reportDetails: Boolean
+  ): IO[String, ValidationReport] = {
```
bogdanromanx, line 88: I think it will be worth it to apply a timeout here to avoid cases where submitting a large model will yield a lot of errors. The timeout could also be handled in the application code (service).

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/shacl/ShaclEngine.scala
bogdanromanx, line 29: this current API will not allow us to cache the shacl engine after the rules have been processed; does the lib allow for that?

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/TestHelpers.scala
```diff
@@ -33,21 +25,6 @@ trait TestHelpers {
     inner("", length)
   }
 
-  /**
-    * Loads the content of the argument classpath resource as a string.
-    *
-    * @param resourcePath the path of a resource available on the classpath
-    * @return the content of the referenced resource as a string
-    */
-  final def contentOf(resourcePath: String): String = {
```
bogdanromanx, line 42: If the `ClasspathResourceUtils` would return a `Task[A]`, I would still retain the current signatures and call `ClasspathResourceUtils.contentOf(...).hideErrors.unsafeRunSync()`

bogdanromanx (changes requested):

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/shacl/ShaclEngine.scala
umbreak: I don't understand your question. In any case, I will postpone anything related to the caching of schemas for when we are going to actually do it.

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/shacl/ShaclEngine.scala
umbreak: I postpone this for later. And I think it is not the responsibility of the `ShaclEngine` to handle the timeout, but of the routes or the `Schemas`

bogdanromanx: agreed

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/shacl/ShaclEngine.scala
bogdanromanx: The ShaclEngine takes both the shapes and the dataset to be validated during construction. This makes the engine instance dependent on the data itself. The problem with this is that the loading or "parsing" of the shacl rules would happen for each new resource, instead of happening only for changes in the shapes.
It's fine to defer this for when we'll work on the schema state.

umbreak: The validation of the shacl shapes does NOT happen every time, that's why we have the flag `validateShapes: Boolean` when validating. In the future we can make public `applySkipShapesCheck` and that's it.

imsdu (approved):

bogdanromanx (approved):

PR #1750, opened 2020-11-03
Acls routes.

wwajerowicz:


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -0,0 +1,275 @@
...
+
+  implicit val iriLens: Lens[AclAddress, Iri] = {
+    case AclAddress.Root    => aclsIri
+    case Organization(org)  => aclsIri / org.value
+    case Project(org, proj) => aclsIri / org.value / proj.value
+  }
+
+  implicit val aclContext: ContextValue = Acl.context
+
+  implicit val aclResponseEncoder: Encoder.AsObject[AclResource] = Encoder.AsObject
```
umbreak, line 62: Some of this implicits might not be necessary. But we'll be able to refactor this once I simplify the `ResourceF` (which also yields some changes on the Acl structure)

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -0,0 +1,275 @@
...
+            extractSubject { implicit subject =>
+              parameter("rev" ? 0L) { rev =>
+                concat(
+                  (put & entity(as[AclInput])) { aclEntity =>
+                    authorizeFor(address, aclsWrite).apply {
+                      val status = if (rev == 0L) Created else OK
+                      completeIO(status, acls.replace(address, aclEntity.toAcl, rev).map(_.void))
+                    }
+                  },
+                  (patch & entity(as[PatchAcl])) { aclPatch =>
```
umbreak, line 129: it would be cleaner if you combine the current directives with the `authorizeFor` directive:
```scala
(patch & entity(as[PatchAcl]) & authorizeFor(address, aclsWrite)) {
   case AppendAcl(acl) => ???
   case SubtractAcl(acl) => ???
}
```

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -0,0 +1,275 @@
...
+
+  }
+
+  sealed trait PatchAcl
+  object PatchAcl {
+    final case class SubtractAcl(acl: Acl) extends PatchAcl
+    final case class AppendAcl(acl: Acl)   extends PatchAcl
+
+    implicit val patchAclDecoder: Decoder[PatchAcl] =
+      Decoder.instance { hc =>
```
umbreak, line 254: Why don't you use circe configured derivation? you already have encoders and decoders for `AclInput`, and you can have 

```scala
    final case class SubtractAcl(acl: AclInput) extends PatchAcl
    final case class AppendAcl(acl: AclInput)   extends PatchAcl
```

umbreak (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -0,0 +1,277 @@
...
+                      authorizeFor(address, aclsRead).apply {
+                        (parameter("rev".as[Long].?) & parameter("ancestors" ? false)) {
+                          case (Some(_), true)    => reject(simultaneousRevAndAncestorsRejection)
+                          case (Some(rev), false) =>
+                            completeIO(
+                              acls
+                                .fetchAt(address, rev)
+                                .map { acl =>
+                                  val searchResults = Seq(acl).flatten
+                                  SearchResults(searchResults.size.toLong, searchResults)
```
umbreak, line 182: why asInstanceOf is needed? if the type is not recognized, fixing it in the `.map[SearchResults[AclResource]]` should be enough. Otherwise something has to be changed

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/DeltaDirectives.scala
```diff
@@ -290,6 +290,22 @@ trait DeltaDirectives extends RdfMarshalling with QueryParamsUnmarshalling {
...
+  def completeUnscoredSearch[A](
+      io: UIO[Seq[A]]
+  )(implicit
+      s: Scheduler,
+      cr: RemoteContextResolution,
+      ordering: JsonKeyOrdering,
+      S: SearchEncoder[A],
+      additionalContext: ContextValue
+  ): Route =
+    completeUIO(io.map(res => SearchResults(res.length.toLong, res).asInstanceOf[SearchResults[A]]))
```
umbreak, line 307: same, is not enough fixing the type in the map?

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/AuthDirectivesSpec.scala
```diff
@@ -1,39 +1,90 @@
...
+          IO.pure(
+            Caller(user, Set(user))
+          )
+        case _                  => IO.raiseError(InvalidAccessToken)
+
+      }
+    }
+  }
+
+  val acls = mock[Acls]
```
umbreak, line 55: Why not using an acl dummy?

umbreak (changes requested):
the acl context is missing [here](https://github.com/BlueBrain/nexus/tree/master/delta/sdk/src/main/resources/contexts). So if the [acl context uri](https://github.com/BlueBrain/nexus/blob/master/delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/Vocabulary.scala#L156) is ever been referenced (it should), the context should be created and added to the `RemoteContextResolution`, otherwise expansion/compaction/n-triples/dot won't work

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -0,0 +1,277 @@
...
+                      }
+                    }
+                  },
+                  delete {
+                    authorizeFor(address, aclsWrite).apply {
+                      completeIO(OK, acls.delete(address, rev).map(_.void))
+                    }
+                  },
+                  (get & parameter("self" ? true)) {
+                    case true  =>
```
umbreak, line 145: Why would you need `acls/read` permissions only to read acls when the self=false and not in this case? If that was previously the logic, it seems flawed.

umbreak (changes requested):

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/AuthDirectivesSpec.scala
wwajerowicz: Acl dummy is in sdk-testkit, which depends on sdk, we can't include sdk-testkit in sdk, because there will be circular dependencies in sbt

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
wwajerowicz: it seems that currently, you can always read your own (`self`) ACLs.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
```diff
@@ -174,4 +174,27 @@ object AclsImpl {
...
     } yield acls
+
+  /**
+    * Constructs an [[AclsImpl]] instance.
+    *
+    * @param config       ACLs configurate
+    * @param permissions  [[Permissions]] instance
+    * @param eventLog     the event log
+    */
+  final def apply(
```
imsdu, line 185: We should only need one apply method no ?

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
wwajerowicz: I could do something like 
```scala
    final case class Subtract(acl: Seq[AclEntry]) extends PatchAcl
    final case class Append(acl: Seq[AclEntry])   extends PatchAcl
```
I can't use AclInput because the JSON to decode would need too look like:
```json
{
  "acl": {
    "acl": [...]
  }
}
```

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -0,0 +1,277 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.search.SearchResults.{encodeResults, searchResultsJsonLdEncoder, SearchEncoder}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{BaseUri, Label, ResourceF}
+import ch.epfl.bluebrain.nexus.delta.sdk.syntax._
+import ch.epfl.bluebrain.nexus.delta.sdk.{AclResource, Acls, Identities, Lens}
+import io.circe.generic.semiauto.deriveDecoder
+import io.circe.syntax.EncoderOps
+import io.circe._
+import monix.execution.Scheduler
+
+class AclsRoutes(identities: Identities, acls: Acls)(implicit
```
imsdu, line 35: There are also routes for acls events ? Or am I mistaken

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -0,0 +1,277 @@
...
+                    }
+                  },
+                  (get & parameter("self" ? true)) {
+                    case true  =>
+                      (parameter("rev".as[Long].?) & parameter("ancestors" ? false)) {
+                        case (Some(_), true)    => reject(simultaneousRevAndAncestorsRejection)
+                        case (Some(rev), false) =>
+                          completeIO(
+                            acls
+                              .fetchSelfAt(address, rev)
```
imsdu, line 151: Shouldn't the operations return a SearchResult so as not having to map it here again ?

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
umbreak: it is the decision of the routes to convert a `AclResource` to a `SerchResult`, so it seems accurate to me that it is done by the routes.

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
wwajerowicz: I think we still need them, because ACLs are encoded a bit differently than other resources:
-  they have `_path` as part of the metadata
- they don't have `_deprecated`

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
imsdu: But then it is accurate to have the search context in this operation ?

bogdanromanx (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -116,43 +117,54 @@ class AclsRoutes(identities: Identities, acls: Acls)(implicit
...
+          },
           extractAclAddress { address =>
             extractSubject { implicit subject =>
               parameter("rev" ? 0L) { rev =>
                 concat(
                   (put & entity(as[AclInput])) { aclEntity =>
-                    authorizeFor(address, aclsWrite).apply {
-                      val status = if (rev == 0L) Created else OK
-                      completeIO(status, acls.replace(address, aclEntity.toAcl, rev).map(_.void))
+                    operationName(s"$prefixSegment/acls/replace") {
```
bogdanromanx, line 133: I would recommend that we keep the operationName to match the uri with escaped addresses, like so:
```
operationName(s"$prefixSegment/acls/{}/{}")
or
operationName(s"$prefixSegment/acls/{}")
or
operationName(s"$prefixSegment/acls")
```
replace or append looks like an RPC api, so the use of kamon tags would be more appropriate, like `"operation" -> "replace"`

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -116,43 +117,54 @@ class AclsRoutes(identities: Identities, acls: Acls)(implicit
     pathPrefix("acls") {
       extractCaller { implicit caller =>
         concat(
+          // SSE realms
+          (pathPrefix("events") & pathEndOrSingleSlash) {
+            operationName(s"$prefixSegment/acls/events") {
+              lastEventId { offset =>
+                completeStream(acls.events(offset))
```
bogdanromanx, line 124: there's no authorization here

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -204,19 +215,21 @@ class AclsRoutes(identities: Identities, acls: Acls)(implicit
...
+              parameter("self" ? true) {
+                case true  =>
+                  completeSearch(
+                    acls
+                      .listSelf(addressFilter)
+                      .map(aclCol => SearchResults(aclCol.value.size.toLong, aclCol.value.values.toSeq))
+                  )
+                case false =>
+                  completeSearch(
+                    acls
```
bogdanromanx, line 228: we would need authorization here as well, wouldn't we?

bogdanromanx (changes requested):

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/AuthDirectivesSpec.scala
imsdu: In AuthDirectives, we only want to fetch the acls so maybe we could just pass a function there ?
Or extract a simpler contract from Acls and pass it there so that we can create a dummy implementation just like Identities here ?

mockito is a Pandora box we should avoid to open imho

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/AuthDirectivesSpec.scala
```diff
@@ -46,34 +97,64 @@ class AuthDirectivesSpec
...
 
+    "correctly authorize user" in {
+      Get("/user") ~> addCredentials(OAuth2BearerToken("alice")) ~> authorizationRoute ~> check {
+        response.status shouldEqual StatusCodes.OK
+        asString(response) shouldEqual "alice"
+      }
+    }
+
+    "correctly reject Anonymous " in {
+      Get("/user") ~> authorizationRoute ~> check {
```
imsdu, line 153: We also should try with another user without the permission

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
```diff
@@ -161,15 +161,15 @@ object AclsImpl {
...
+      permissions: Permissions,
       eventLog: EventLog[Envelope[AclEvent]]
   )(implicit
       as: ActorSystem[Nothing],
       sc: Scheduler,
       clock: Clock[UIO]
   ): UIO[AclsImpl] =
     for {
-      agg  <- aggregate(permissions, config.aggregate)
+      agg  <- aggregate(UIO.delay(permissions), config.aggregate)
```
imsdu, line 172: I would rather change the UIO[Permissions] for Permissions in the state machine

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
wwajerowicz: `AclResource` is defined as 
```scala
type AclResource = ResourceF[AclAddress, Acl]
```
and `Acl` doesn't have the `AclAddress` to make `_path` from the resource value

umbreak: Don't worry about this point for now. With the changes I've introduced in `ResourceF[A]` `Acl` will actually have the `AclAddress`. But you can keep as is now.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Acl.scala
```diff
@@ -80,6 +87,28 @@ final case class Acl(value: Map[Identity, Set[Permission]]) {
 
 object Acl {
 
+  implicit private[Acl] val config: Configuration = Configuration.default
+
+  implicit def aclEncoder(implicit base: BaseUri): Encoder.AsObject[Acl] = Encoder.AsObject.instance { acl =>
```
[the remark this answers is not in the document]
wwajerowicz: this is needed in the routes

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Acl.scala
```diff
@@ -80,6 +87,28 @@ final case class Acl(value: Map[Identity, Set[Permission]]) {
...
+      "acl" -> Json.fromValues(
+        acl.value.map { case (identity, permissions) =>
+          Json.obj("identity" -> identity.asJson, "permissions" -> permissions.asJson)
+        }
+      )
+    )
+
+  }
+
+  implicit val aclDecoder: Decoder[Acl] = Decoder.instance { hc =>
```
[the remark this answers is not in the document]
wwajerowicz: this is not needed, deleted

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Acl.scala
```diff
@@ -80,6 +87,28 @@ final case class Acl(value: Map[Identity, Set[Permission]]) {
...
+    hc.get[Seq[(Identity, Set[Permission])]]("acl").map(Acl(_: _*))
+  }
+
+  val context: ContextValue = ContextValue(contexts.acls)
```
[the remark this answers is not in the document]
wwajerowicz: this is needed in the routes.

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
wwajerowicz: Actually, we need filtering here. Current behaviour is that when user list e.g. `/*/*` we show him all the ACLs that match the pattern to which user has access.

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -236,8 +238,9 @@ object AclsRoutes {
 
   type AclResponseResource = ResourceF[Iri, Acl]
 
-  private val aclsRead  = Permission.unsafe("acls/read")
-  private val aclsWrite = Permission.unsafe("acls/write")
+  private val aclsRead   = Permission.unsafe("acls/read")
```
umbreak, line 244: Please do not create more permissions, but use the ones already defined in [Permissions](https://github.com/BlueBrain/nexus/blob/master/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Permissions.scala#L130)

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/AclsRoutesSpec.scala
```diff
@@ -71,7 +71,12 @@ class AclsRoutesSpec
     )
   private val acls                                       = AclsDummy(
     PermissionsDummy(
-      Set(Permission.unsafe("acls/read"), Permission.unsafe("acls/write"), Permission.unsafe("acls/manage"))
+      Set(
+        Permission.unsafe("acls/read"),
```
umbreak, line 75: same here

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/AclsRoutesSpec.scala
```diff
@@ -157,7 +162,8 @@ class AclsRoutesSpec
     }
 
     "get the events stream" in {
-      Get("/v1/acls/events") ~> Accept(`*/*`) ~> routes ~> check {
+      Get("/v1/acls/events") ~> addCredentials(token) ~> Accept(`*/*`) ~> routes ~> check {
+        print(response.asString)
```
umbreak, line 166: remove

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/AclCollection.scala
```diff
@@ -71,6 +71,18 @@ final case class AclCollection private (value: SortedMap[AclAddress, AclResource
...
 
+  /**
+    * Generates a new [[AclCollection]] only containing [[Acl]]s
+    * where user has `permission` on the [[Acl]] or ancestor
+    *
+    * @param identities the identities of the user
+    * @param permission the permission to filter by
+    */
+  def filterByPermission(identities: Set[Identity], permission: Permission): AclCollection =
+    AclCollection.apply(value.filter { case (address, _) =>
```
umbreak, line 82: Do we need the .apply?

umbreak (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -0,0 +1,289 @@
...
+        "_path" -> Json.fromString(r.id.string)
+      )
+    }
+    .mapJsonObject(_.remove("_deprecated"))
+
+  implicit val aclWriteResponseEncoder: Encoder.AsObject[ResourceF[AclAddress, Unit]] = Encoder.AsObject
+    .instance { r: ResourceF[AclAddress, Unit] =>
+      r.copy(id = iriLens.get(r.id)).void.asJsonObject deepMerge r.value.asJsonObject
+    }
+    .mapJsonObject(_.remove("_deprecated"))
```
umbreak, line 75: We don't need to remove deprecated, we are adding this metadata field to all resources. So basically, you can use the default derived one

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -0,0 +1,289 @@
...
+
+  implicit val aclContext: ContextValue = Acl.context
+
+  implicit val aclResponseEncoder: Encoder.AsObject[AclResource] = Encoder.AsObject
+    .instance { r: AclResource =>
+      r.copy(id = iriLens.get(r.id)).void.asJsonObject deepMerge r.value.asJsonObject deepMerge JsonObject(
+        "_path" -> Json.fromString(r.id.string)
+      )
+    }
+    .mapJsonObject(_.remove("_deprecated"))
```
umbreak, line 69: same, don't remove deprecated

delta/app/src/test/resources/acls/write-response-routes.json
```diff
@@ -0,0 +1,11 @@
+{
+  "@context": "https://bluebrain.github.io/nexus/contexts/resource.json",
```
umbreak, line 2: it should contain `"_deprecated": "false"`, like any other resource type

delta/sdk/src/main/resources/contexts/acls.json
```diff
@@ -0,0 +1,17 @@
+{
+  "@context": {
+    "nxv": "https://bluebrain.github.io/nexus/vocabulary/",
+    "_path": "nxv:path",
+    "AccessControlList": "nxv:AccessControlList",
```
umbreak, line 5: It can just be
```json
{
  "@context": {
      "nxv": "https://bluebrain.github.io/nexus/vocabulary/",
      "@vocab": "https://bluebrain.github.io/nexus/vocabulary/",
      "_path": "nxv:path"
   },
  "@id": "https://bluebrain.github.io/nexus/contexts/acls.json"
}
```

umbreak (changes requested):

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/AuthDirectivesSpec.scala
wwajerowicz: I believe creating a separate contract for a single test is a bit of an overhead.
We should avoid mockito if we can, but I believe in this case it's actually simpler to use it since we can't use dummy implementation in here.

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
wwajerowicz: Added the routes.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
wwajerowicz: Removed the other apply method.

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/AuthDirectivesSpec.scala
wwajerowicz: Added the test.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/AclCollection.scala
wwajerowicz: We don't :)

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/AuthDirectivesSpec.scala
imsdu: We can reuse it in the different route tests.

umbreak (dismissed):
Looks fine

umbreak (approved):

bogdanromanx (approved):

imsdu (approved):

PR #1756, opened 2020-11-05
Added AccessUrl

umbreak:
- Added `_self` (AccessUrl) metadata field to all resource types.
- Refactored `ResourceF` to only contain one type parameter.
- Fixed project fetch json response. Now it returns the mappings as expected (with `prefix` and `namespace` fields)


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
```diff
@@ -29,37 +29,39 @@ import org.slf4j.{Logger, LoggerFactory}
   */
 // $COVERAGE-OFF$
 class DeltaModule(appCfg: AppConfig, config: Config) extends ModuleDef with ClasspathResourceUtils {
-  private val resourceCtx      = ioJsonContentOf("/contexts/resource.json").memoizeOnSuccess
-  private val permissionsCtx   = ioJsonContentOf("/contexts/permissions.json").memoizeOnSuccess
+  private val aclsCtx          = ioJsonContentOf("/contexts/acls.json").memoizeOnSuccess
```
imsdu, line 32: I think it would better if we define the Context (Iri + IO) in the same place.

delta/sdk/src/test/resources/resource-compacted.jsonld
```diff
@@ -9,8 +9,9 @@
   "_rev": 1,
   "_deprecated": false,
   "_createdAt": "1970-01-01T00:00:00Z",
-  "_createdBy": "http://nexus.com/anonymous",
+  "_createdBy": "http://example.com/anonymous",
```
imsdu, line 12: Better to replace with localhost

imsdu (changes requested):

bogdanromanx (approved):

imsdu (approved):

PR #1762, opened 2020-11-09
Defined Resources events, states, rejections and contract

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1491

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
@@ -123,7 +123,7 @@ trait Resources {
     * @param id        the resource identifier
     * @param project   the project where the resource belongs
     * @param schemaOpt the optional schema of the resource. A None value ignores the schema from this operation
-    * @param rev       the permissions revision
+    * @param rev       the resources revision
     * @return the resource as a resource at the specified revision
     */
   def fetchAt(
@@ -132,6 +132,30 @@ trait Resources {
       schemaOpt: Option[ResourceRef],
       rev: Long
   ): IO[RevisionNotFound, Option[DataResource]]
+
+  /**
+    * Fetches a resource by tag.
+    *
+    * @param id        the resource identifier
+    * @param project   the project where the resource belongs
+    * @param schemaOpt the optional schema of the resource. A None value ignores the schema from this operation
+    * @param tag       the tag revision
+    * @return the resource as a resource at the specified revision
+    */
+  def fetchBy(
+      id: Iri,
+      project: ProjectRef,
+      schemaOpt: Option[ResourceRef],
+      tag: Label
+  ): IO[TagNotFound, Option[DataResource]] =
+    fetch(id, project, schemaOpt).flatMap {
+      case Some(resource) =>
+        resource.value.tags.get(tag) match {
+          case Some(rev) => fetchAt(id, project, schemaOpt, rev).leftMap(_ => TagNotFound(tag))
+          case None      => IO.raiseError(TagNotFound(tag))
+        }
+      case None           => IO.pure(None)
+    }
 
   /**
     * A non terminating stream of events for resources. After emitting all known events it sleeps until new events
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
@@ -1,0 +1,305 @@
+package ch.epfl.bluebrain.nexus.delta.sdk
+
+import akka.persistence.query.{NoOffset, Offset}
+import cats.effect.Clock
+import cats.syntax.all._
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.schemas
+import ch.epfl.bluebrain.nexus.delta.rdf.graph.Graph
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.ExpandedJsonLd
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution
+import ch.epfl.bluebrain.nexus.delta.rdf.shacl.ShaclEngine
+import ch.epfl.bluebrain.nexus.delta.sdk.model.identities.Identity.Subject
+import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectRef
+import ch.epfl.bluebrain.nexus.delta.sdk.model.resources.ResourceCommand.{CreateResource, DeprecateResource, TagResource, UpdateResource}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.resources.ResourceEvent.{ResourceCreated, ResourceDeprecated, ResourceTagAdded, ResourceUpdated}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.resources.ResourceRejection._
+import ch.epfl.bluebrain.nexus.delta.sdk.model.resources.ResourceState._
+import ch.epfl.bluebrain.nexus.delta.sdk.model.resources.{ResourceCommand, ResourceEvent, ResourceRejection, ResourceState}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.schemas.Schema
+import ch.epfl.bluebrain.nexus.delta.sdk.model.search.Pagination.FromPagination
+import ch.epfl.bluebrain.nexus.delta.sdk.model.search.SearchParams.ResourceSearchParams
+import ch.epfl.bluebrain.nexus.delta.sdk.model.search.SearchResults.UnscoredSearchResults
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{Envelope, ResourceRef}
+import ch.epfl.bluebrain.nexus.delta.sdk.utils.IOUtils
+import fs2.Stream
+import io.circe.Json
+import monix.bio.{IO, Task, UIO}
+
+/**
+  * Operations pertaining to managing resources.
+  */
+trait Resources {
+
+  /**
+    * Creates a new resource with the passed id.
+    *
+    * @param id      the resource identifier
+    * @param project the project where the resource belongs
+    * @param schema  the schema to validate the resource against
+    * @param source  the resource payload
+    */
+  def create(
+      id: Iri,
+      project: ProjectRef,
+      schema: ResourceRef,
+      source: Json
+  )(implicit caller: Subject): IO[ResourceRejection, DataResource]
+
+  /**
+    * Creates a new resource where the id is either present on the payload or self generated.
+    *
+    * @param project the project where the resource belongs
+    * @param source  the resource payload
+    * @param schema  the schema to validate the resource against
+    */
+  def create(
+      project: ProjectRef,
+      schema: ResourceRef,
+      source: Json
+  )(implicit caller: Subject): IO[ResourceRejection, DataResource]
+
+  /**
+    * Updates an existing resource.
+    *
+    * @param id        the resource identifier
+    * @param project   the project where the resource belongs
+    * @param schemaOpt the optional schema of the resource. A None value ignores the schema from this operation
+    * @param rev       the current revision of the resource
+    * @param source    the resource payload
+    */
+  def update(
+      id: Iri,
+      project: ProjectRef,
+      schemaOpt: Option[ResourceRef],
+      rev: Long,
+      source: Json
+  )(implicit caller: Subject): IO[ResourceRejection, DataResource]
+
+  /**
+    * Adds a tag to an existing resource.
+    *
+    * @param id      the resource identifier
+    * @param project the project where the resource belongs
+    * @param schemaOpt the optional schema of the resource. A None value ignores the schema from this operation
+    * @param tag the tag name
+    * @param tagRev the tag revision
+    * @param rev     the current revision of the resource
+    */
+  def tag(
+      id: Iri,
+      project: ProjectRef,
+      schemaOpt: Option[ResourceRef],
+      tag: String,
+      tagRev: Long,
+      rev: Long
+  )(implicit caller: Subject): IO[ResourceRejection, DataResource]
+
+  /**
+    * Deprecates an existing resource.
+    *
+    * @param id        the resource identifier
+    * @param project   the project where the resource belongs
+    * @param schemaOpt the optional schema of the resource. A None value ignores the schema from this operation
+    * @param rev       the revision of the resource
+    */
+  def deprecate(
+      id: Iri,
+      project: ProjectRef,
+      schemaOpt: Option[ResourceRef],
+      schema: ResourceRef,
+      rev: Long
+  )(implicit caller: Subject): IO[ResourceRejection, DataResource]
+
+  /**
+    * Fetches a resource.
+    *
+    * @param id        the resource identifier
+    * @param project   the project where the resource belongs
+    * @param schemaOpt the optional schema of the resource. A None value ignores the schema from this operation
+    * @return the resource in a Resource representation, None otherwise
+    */
+  def fetch(id: Iri, project: ProjectRef, schemaOpt: Option[ResourceRef]): UIO[Option[DataResource]]
+
+  /**
+    * Fetches a resource at a specific revision.
+    *
+    * @param id        the resource identifier
+    * @param project   the project where the resource belongs
+    * @param schemaOpt the optional schema of the resource. A None value ignores the schema from this operation
+    * @param rev       the permissions revision
+    * @return the resource as a resource at the specified revision
+    */
+  def fetchAt(
+      id: Iri,
+      project: ProjectRef,
+      schemaOpt: Option[ResourceRef],
+      rev: Long
+  ): IO[RevisionNotFound, Option[DataResource]]
+
+  /**
+    * Lists resources with optional filters.
+    *
+    * @param pagination the pagination settings
+    * @param params     filter parameters of the resources
+    * @return a paginated results list of resources sorted by their creation date.
+    */
+  def list(
+      pagination: FromPagination,
+      params: ResourceSearchParams = ResourceSearchParams.none
+  ): UIO[UnscoredSearchResults[DataResource]]
+
+  /**
+    * A non terminating stream of events for resources. After emitting all known events it sleeps until new events
+    * are recorded.
+    *
+    * @param offset the last seen event offset; it will not be emitted by the stream
+    */
+  def events(offset: Offset = NoOffset): Stream[Task, Envelope[ResourceEvent]]
+
+  /**
+    * The current resource events. The stream stops after emitting all known events.
+    *
+    * @param offset the last seen event offset; it will not be emitted by the stream
+    */
+  def currentEvents(offset: Offset = NoOffset): Stream[Task, Envelope[ResourceEvent]]
+
+}
+
+object Resources {
+
+  /**
+    * The resources module type.
+    */
+  final val moduleType: String = "resource"
+
+  private[delta] def next(state: ResourceState, event: ResourceEvent): ResourceState = {
+    // format: off
+    def created(e: ResourceCreated): ResourceState = state match {
+      case Initial     => Current(e.id, e.project, e.source, e.compacted, e.expanded, e.rev, deprecated = false, e.schema, e.types, Map.empty, e.instant, e.subject, e.instant, e.subject)
+      case s: Current  => s
+    }
+
+    def updated(e: ResourceUpdated): ResourceState = state match {
+      case Initial    => Initial
+      case s: Current => s.copy(rev = e.rev, types = e.types, source = e.source, compacted = e.compacted, expanded = e.expanded, updatedAt = e.instant, updatedBy = e.subject)
+    }
+
+    def tagAdded(e: ResourceTagAdded): ResourceState = state match {
+      case Initial    => Initial
+      case s: Current => s.copy(rev = e.rev, tags = s.tags + (e.tag -> e.targetRev), updatedAt = e.instant, updatedBy = e.subject)
+    }
+    // format: on
+
+    def deprecated(e: ResourceDeprecated): ResourceState = state match {
+      case Initial    => Initial
+      case s: Current => s.copy(rev = e.rev, deprecated = true, updatedAt = e.instant, updatedBy = e.subject)
+    }
+    event match {
+      case e: ResourceCreated    => created(e)
+      case e: ResourceUpdated    => updated(e)
+      case e: ResourceTagAdded   => tagAdded(e)
+      case e: ResourceDeprecated => deprecated(e)
+    }
+  }
+
+  @SuppressWarnings(Array("OptionGet"))
+  private[delta] def evaluate(
+      fetchSchema: ResourceRef => UIO[Option[SchemaResource]]
+  )(state: ResourceState, cmd: ResourceCommand)(implicit
+      clock: Clock[UIO] = IO.clock,
+      rcr: RemoteContextResolution
+  ): IO[ResourceRejection, ResourceEvent] = {
+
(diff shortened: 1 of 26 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
```diff
@@ -0,0 +1,115 @@
...
+    * @param compacted   the compacted JSON-LD representation of the resource
+    * @param expanded    the expanded JSON-LD representation of the resource
+    * @param subject     the subject which created this event
+    */
+  final case class CreateResource(
+      id: Iri,
+      project: ProjectRef,
+      schema: ResourceRef,
+      source: Json,
+      compacted: CompactedJsonLd,
```
imsdu, line 50: Shouldn't the compacted/expanded computation append in the evaluate part ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
```diff
@@ -0,0 +1,115 @@
...
+      expanded: ExpandedJsonLd,
+      subject: Subject
+  ) extends ResourceCommand
+
+  /**
+    * Command that signals the intent to update an existing resource.
+    *
+    * @param id        the resource identifier
+    * @param project   the project where the resource belongs
```
imsdu, line 59: We don't need the project here as a resource can't switch from a project to another right ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
```diff
@@ -0,0 +1,115 @@
...
+    * @param schemaOpt the optional schema of the resource. A None value ignores the schema from this operation
+    * @param targetRev the revision that is being aliased with the provided ''tag''
+    * @param tag       the tag of the alias for the provided ''tagRev''
+    * @param rev       the last known revision of the resource
+    * @param subject   the subject which created this event
+    */
+  final case class TagResource(
+      id: Iri,
+      project: ProjectRef,
+      schemaOpt: Option[ResourceRef],
```
imsdu, line 92: The schema can't be changed via tagging

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
```diff
@@ -0,0 +1,115 @@
...
+      rev: Long,
+      subject: Subject
+  ) extends ResourceCommand
+
+  /**
+    * Command that signals the intent to deprecate a resource.
+    *
+    * @param id        the resource identifier
+    * @param project   the project where the resource belongs
+    * @param schemaOpt the optional schema of the resource. A None value ignores the schema from this operation
```
imsdu, line 104: Same

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
```diff
@@ -0,0 +1,115 @@
...
+    * @param compacted the compacted JSON-LD representation of the resource
+    * @param expanded  the expanded JSON-LD representation of the resource
+    * @param rev       the last known revision of the resource
+    * @param subject   the subject which created this event
+    */
+  final case class UpdateResource(
+      id: Iri,
+      project: ProjectRef,
+      schemaOpt: Option[ResourceRef],
+      source: Json,
```
imsdu, line 71: If I only want to change the schema, I still have to pass the source ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceEvent.scala
```diff
@@ -0,0 +1,170 @@
...
+    * @param expanded    the expanded JSON-LD representation of the resource
+    * @param rev         the resource revision
+    * @param instant     the instant when this event was created
+    * @param subject     the subject which created this event
+    */
+  final case class ResourceUpdated(
+      id: Iri,
+      project: ProjectRef,
+      types: Set[Iri],
+      source: Json,
```
imsdu, line 88: Can it happen that an update doesn't change the source/compacted/expanded ?
If yes, we can make it optional ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
```diff
@@ -0,0 +1,115 @@
...
+    * @param tag       the tag of the alias for the provided ''tagRev''
+    * @param rev       the last known revision of the resource
+    * @param subject   the subject which created this event
+    */
+  final case class TagResource(
+      id: Iri,
+      project: ProjectRef,
+      schemaOpt: Option[ResourceRef],
+      targetRev: Long,
+      tag: String,
```
imsdu, line 94: We could have a label here no ?

imsdu: String offers too much possibilities

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceEvent.scala
```diff
@@ -0,0 +1,170 @@
...
+    * @param source      the representation of the resource as posted by the subject
+    * @param compacted   the compacted JSON-LD representation of the resource
+    * @param expanded    the expanded JSON-LD representation of the resource
+    * @param rev         the resource revision
+    * @param instant     the instant when this event was created
+    * @param subject     the subject which created this event
+    */
+  final case class ResourceUpdated(
+      id: Iri,
+      project: ProjectRef,
```
imsdu, line 86: We can't update the schema ?
If we can, we could have 2 types of events no ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceEvent.scala
```diff
@@ -0,0 +1,170 @@
...
+    * @param targetRev the revision that is being aliased with the provided ''tag''
+    * @param tag       the tag of the alias for the provided ''targetRev''
+    * @param rev       the resource revision
+    * @param instant   the instant when this event was created
+    * @param subject   the subject which created this event
+    */
+  final case class ResourceTagAdded(
+      id: Iri,
+      project: ProjectRef,
+      types: Set[Iri],
```
imsdu, line 111: We don't need the types as they don't change and are not a part of the resource identifier no ?

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
umbreak: The project field is always passed along, since it is the way to generate the persistenceId (which is going to be projectRef + id)

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
umbreak: True, but it _can_ be passed to the API. If you see the evaluation logic, when it is passed and it does not match the expected one, it will reject.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
umbreak: The schema cannot be changed at all, but it is passed for the same reason as I explained before

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceEvent.scala
umbreak: No

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
imsdu: Then you have it in the state, no need to pass it again though the command

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceEvent.scala
umbreak: The types are some information we want to carry on the event. This is for indexing purposes, since some indexing decisions will depend on the types available

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
umbreak: It is needed, you will see it more clear when you have the dummy or Impl

imsdu: What I meant is you can take the projectRef of the state to create the event.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceEvent.scala
imsdu: We could issue a rejection if the source doesn't change ? Or we don't care ?

umbreak: That's currently allowed. The source might be the same but the expanded/compacted form might be different, since remember that it can contain a reference to a context which could have been updated meanwhile.

We could have something that checks if source && compacted && expanded are the same in the updated command and in the current state, to reject. Even though again, that's not the current behaviour

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceEvent.scala
umbreak: Once the resource has been created against one schema it cannot be updated against another schema.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
umbreak: Ideally yes. But since there is an edge case when the expanded and compacted has to be computed before the evaluate function, I prefered to keep them all outside.

The exception is:
You can create a resource without specifying the id. In this case, we attempt to fetch the id from the payload, meaning that you have to expand the resource **before** calling evaluate, in order to extract its id

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
```diff
@@ -0,0 +1,305 @@
...
+          for {
+            graph <- asGraph(c.id, c.expanded)
+            _     <- validate(s.schema, c.id, graph)
+            types  = c.expanded.types.toSet
+            time  <- IOUtils.instant
+          } yield ResourceUpdated(c.id, c.project, types, c.source, c.compacted, c.expanded, s.rev + 1, time, c.subject)
+
+      }
+
+    def tag(c: TagResource) =
```
bogdanromanx, line 267: we don't verify deprecation status here; is this the same behaviour from 1.4.x?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/AccessUrl.scala
```diff
@@ -108,4 +110,8 @@ object AccessUrl {
   private def context(mappings: ApiMappings): JsonLdContext =
     JsonLdContext(ContextValue.empty, prefixMappings = mappings.prefixMappings, aliases = mappings.aliases)
 
+  private val defaultMappings: ApiMappings = ApiMappings(
+    Map("_" -> schemas.resources, "schema" -> schemas.shacl, "resolver" -> schemas.resolvers)
```
bogdanromanx, line 114: I think these used to be defined as plural (`schemas`, `resolvers`)...

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceCommand.scala
bogdanromanx: This is a shame because of the message sizes exchanged between the nodes... but you're right, we need to extract the id before sending the message.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceEvent.scala
bogdanromanx: It would be a nice optimization, but not in the scope of this PR.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceEvent.scala
bogdanromanx: Also for backwards compatibility of the SSE.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceEvent.scala
```diff
@@ -0,0 +1,170 @@
...
+  final case class ResourceDeprecated(
+      id: Iri,
+      project: ProjectRef,
+      types: Set[Iri],
+      rev: Long,
+      instant: Instant,
+      subject: Subject
+  ) extends ResourceEvent
+
+  //TODO: Maybe add a resources context
```
bogdanromanx, line 138: what do you mean by that?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceRejection.scala
```diff
@@ -0,0 +1,159 @@
...
+    */
+  final case class ResourceNotFound(id: Iri) extends ResourceRejection(s"Resource '$id' not found.")
+
+  /**
+    * Rejection returned when attempting to create a resource where the passed id does not match the id on the payload.
+    *
+    * @param id        the resource identifier
+    * @param payloadId the resource identifier on the payload
+    */
+  final case class ResourceIdUnexpected(id: Iri, payloadId: Iri)
```
bogdanromanx, line 54: `UnexpectedResourceId`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceRejection.scala
```diff
@@ -0,0 +1,159 @@
...
+      extends ResourceRejection(s"Resource '$id' failed to validate against the constrains defined in schema '$schema'")
+
+  /**
+    * Rejection returned when attempting to update/deprecate a resource with a different schema than the resource schema.
+    *
+    * @param id       the resource identifier
+    * @param provided the resource provided schema
+    * @param expected the resource schema
+    */
+  final case class ResourceSchemaUnexpected(id: Iri, provided: ResourceRef, expected: ResourceRef)
```
bogdanromanx, line 74: `UnexpectedResourceSchema`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceRejection.scala
```diff
@@ -0,0 +1,159 @@
...
+      )
+
+  /**
+    * Rejection returned when attempting to create a SHACL engine.
+    *
+    * @param id      the resource identifier
+    * @param schema  the resource provided schema
+    * @param details the SHACL engine errors
+    */
+  final case class ResourceShaclEngineRejection(id: Iri, schema: ResourceRef, details: String)
```
bogdanromanx, line 86: What are the situations when a simple model would not be accepted by the engine? It seems that the failure is more of a schema problem, not the provided resource.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceRejection.scala
```diff
@@ -0,0 +1,159 @@
...
+    * @param ref     the schema reference
+    */
+  final case class SchemaIsDeprecated(ref: ResourceRef) extends ResourceRejection(s"Schema '$ref' is deprecated.")
+
+  /**
+    * Rejection returned when attempting to validate a resource against a schema that is deprecated.
+    *
+    * @param ref     the schema reference
+    */
+  final case class SchemaNotFound(ref: ResourceRef) extends ResourceRejection(s"Schema '$ref' not found.")
```
bogdanromanx, line 132: Not for this PR, but would be nice to consider providing the full resolution tree in this situation when resolvers would be implemented.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceRejection.scala
```diff
@@ -0,0 +1,159 @@
...
+
+  final case class ResourceJsonLdPayloadRejection(idOpt: Option[Iri], rdfError: RdfError)
```
bogdanromanx, line 134: `InvalidJsonLdFormat`?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/search/SearchParams.scala
```diff
@@ -119,6 +120,33 @@ object SearchParams {
...
+    * Search parameters to filter data resources.
+    *
+    * @param deprecated the optional deprecation status of the project resources
+    * @param rev        the optional revision of the project resources
+    * @param createdBy  the optional subject who created the project resource
+    * @param updatedBy  the optional subject who updated the resource
+    * @param types      the resource types
+    * @param schema     the optional schema
+    */
+  final case class ResourceSearchParams(
```
bogdanromanx, line 133: I think these will have to be ported to the view plugin; I would remove this for now.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
```diff
@@ -0,0 +1,305 @@
...
+      fetchSchema: ResourceRef => UIO[Option[SchemaResource]]
+  )(state: ResourceState, cmd: ResourceCommand)(implicit
+      clock: Clock[UIO] = IO.clock,
+      rcr: RemoteContextResolution
+  ): IO[ResourceRejection, ResourceEvent] = {
+
+    def asGraph(id: Iri, expanded: ExpandedJsonLd): IO[ResourceRejection, Graph] =
+      expanded.toGraph.leftMap(err => ResourceJsonLdPayloadRejection(Some(id), err))
+
+    def validate(schemaRef: ResourceRef, id: Iri, graph: Graph): IO[ResourceRejection, Unit] = {
```
bogdanromanx, line 216: I think the entire validation logic should be the "schema's" responsibility. It would simplify this logic if the STM would only take in a validation function `Graph => UIO[InvalidResource, Unit]`.
This would also enable for later schema caching...

bogdanromanx (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
```diff
@@ -0,0 +1,305 @@
...
+    *
+    * @param id        the resource identifier
+    * @param project   the project where the resource belongs
+    * @param schemaOpt the optional schema of the resource. A None value ignores the schema from this operation
+    * @param rev       the revision of the resource
+    */
+  def deprecate(
+      id: Iri,
+      project: ProjectRef,
+      schemaOpt: Option[ResourceRef],
```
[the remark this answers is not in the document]
umbreak: No, just `schemaOpt `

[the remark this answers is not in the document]
umbreak: No, just schemaOpt

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/AccessUrl.scala
umbreak: No, it is as I defined: https://github.com/BlueBrain/nexus/blob/v1.4.x/delta/src/main/scala/ch/epfl/bluebrain/nexus/kg/directives/ProjectDirectives.scala#L24

bogdanromanx: right, please ignore

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceRejection.scala
umbreak: Whatever it is, at this point has to be signaled in the resource STM, since it can happen there. We can change it if needed when schemas are created

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
umbreak: We will see when schemas is done, then the logic can be ported if that's what we want. Now it fits well here, since resources are the only thing that needs to be validated by schemas. So it is quite self contained.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceRejection.scala
bogdanromanx: Please see my comment on porting the validation to the schema (below)

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
bogdanromanx: Happy to consider changing this in a subsequent PR when schemas are implemented.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
umbreak: Yes you're right

imsdu (approved):

bogdanromanx (approved):

PR #1783, opened 2020-11-10
Grant Anonymous full permissions on / when no acls set

### LINKED ISSUE #1766: Default ACLs are not returned when there are no ACLs defined on Root
Because there are no default ACLs no further access can be granted to a user and no functions can be consumed.

bogdanromanx:
Fixes #1766

### DIFF
--- a/delta/sdk-testkit/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/AclsBehaviors.scala
+++ b/delta/sdk-testkit/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/AclsBehaviors.scala
@@ -4,17 +4,18 @@ import java.time.Instant
 
 import akka.persistence.query.Sequence
 import ch.epfl.bluebrain.nexus.delta.kernel.utils.ClassUtils
-import ch.epfl.bluebrain.nexus.delta.sdk.Acls
 import ch.epfl.bluebrain.nexus.delta.sdk.generators.AclGen.resourceFor
 import ch.epfl.bluebrain.nexus.delta.sdk.model.acls.AclAddress.Organization
 import ch.epfl.bluebrain.nexus.delta.sdk.model.acls.AclAddressFilter.{AnyOrganization, AnyOrganizationAnyProject, AnyProject}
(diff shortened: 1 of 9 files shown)


delta/sdk-testkit/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/AclsBehaviors.scala
```diff
@@ -70,15 +71,36 @@ trait AclsBehaviors {
...
 
   "An ACLs implementation" should {
-    val acls =
+    val (acls, permissions) =
       create
         .timeoutWith(10.seconds, new TimeoutException("Unable to create a permissions instance"))
         .memoizeOnSuccess
         .accepted
 
+    "return the full permissions for Anonymous if no permissions are defined" in {
```
imsdu, line 83: We should test again that the permissions for Anonymous are gone when we add some acls

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Acls.scala
```diff
@@ -301,4 +295,40 @@ object Acls {
...
+      acls.fetch(AclAddress.Root).flatMap {
+        case Some(value) => UIO.pure(AclCollection(value))
+        case None        =>
+          perms.fetchPermissionSet.map(ps =>
+            AclCollection(
+              Current(
+                acl = Acl(AclAddress.Root, Identity.Anonymous -> ps),
+                rev = 0L,
+                createdAt = Instant.EPOCH,
+                createdBy = Identity.Anonymous,
```
umbreak, line 316: Why anonymous? Wasn't this chosen by configuration?

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Acls.scala
bogdanromanx: it was always anonymous; I also don't understand the value of having anything different from Anonymous driven by config, since the API can be used to grant any custom access.

umbreak (dismissed):

imsdu (approved):

PR #1784, opened 2020-11-10
Created Resources implementation

umbreak:
Fixes: https://github.com/BlueBrain/nexus/issues/1521

- Added DB serialization for `ResourceEvent`.
- Added Encoder/Decoder for `ResourceRef` (that required some additional methods in `Iri`).
- Small fixes on decoding (forcing strict decoding for input payload from the client only).

### DIFF
--- a/delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/IriOrBNode.scala
+++ b/delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/IriOrBNode.scala
@@ -2,7 +2,7 @@ package ch.epfl.bluebrain.nexus.delta.rdf
 
 import java.util.UUID
 
-import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri.unsafe
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri.{unsafe, Query}
 import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.{BNode, Iri}
 import io.circe.{Decoder, Encoder}
 import org.apache.jena.iri.{IRI, IRIFactory}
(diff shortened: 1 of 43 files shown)


imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceRef.scala
```diff
@@ -23,24 +31,49 @@ object ResourceRef {
...
     * A reference annotated with a tag.
     *
-    * @param iri the reference identifier as an iri
-    * @param tag the reference tag
+    * @param original the original iri
+    * @param iri      the reference identifier as an iri
+    * @param tag      the reference tag
     */
-  final case class Tag(iri: Iri, tag: String) extends ResourceRef
+  final case class Tag(original: Iri, iri: Iri, tag: String) extends ResourceRef
```
imsdu, line 54: tag is a label too here no ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceRef.scala
```diff
@@ -13,7 +16,12 @@ sealed trait ResourceRef extends Product with Serializable {
     */
   def iri: Iri
```
imsdu, line 17: It is not clear when reading the comments the difference between an the iri and the original

imsdu (changes requested):

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/IriOrBNode.scala
```diff
@@ -47,6 +48,51 @@ object IriOrBNode {
...
+      */
+    def removeQueryParams(keys: String*): Iri =
+      if (rawQuery().isEmpty) this
+      else queryParams(Query(query().toMap -- keys))
+
+    /**
+      * Override the current query parameters with the passed ones
+      */
+    def queryParams(query: Query): Iri = {
+      val sb = new StringBuilder
```
imsdu, line 75: The iriFactory has methods that could help here

imsdu (changes requested):

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/IriOrBNode.scala
umbreak: The construct method does not work, it is [not implemented](https://github.com/apache/jena/blob/master/jena-iri/src/main/java/org/apache/jena/iri/IRIFactory.java#L473)

imsdu (approved):

bogdanromanx (approved):

PR #1787, opened 2020-11-11
Refactored ResourcesDummy and ResourcesImpl considering Projects

umbreak:


delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ResourcesDummy.scala
```diff
@@ -23,88 +24,112 @@ import monix.bio.{IO, Task, UIO}
...
   override def deprecate(
       id: Iri,
-      project: ProjectRef,
+      projectRef: ProjectRef,
       schemaOpt: Option[ResourceRef],
       rev: Long
   )(implicit caller: Subject): IO[ResourceRejection, DataResource] =
-    eval(DeprecateResource(id, project, schemaOpt, rev, caller))
+    for {
+      project <- projects.activeProject(projectRef).leftMap(WrappedProjectRejection)
```
imsdu, line 99: When deprecating, the project could be deprecated, it is not a problem no ?

Also like schemas and other stuff, we may want to fetch a specific revision of the project to avoid mutability if the api mappings changes

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ResourcesDummy.scala
```diff
@@ -23,88 +24,112 @@ import monix.bio.{IO, Task, UIO}
...
-      .currentState((project, id), Initial, Resources.next)
-      .map(_.flatMap(_.toResource))
-      .map(validateSameSchema(_, schemaOpt))
+  override def fetch(
+      id: Iri,
+      projectRef: ProjectRef,
+      schemaOpt: Option[ResourceRef]
+  ): IO[ResourceRejection, Option[DataResource]] =
+    for {
+      project  <- projects.activeProject(projectRef).leftMap(WrappedProjectRejection)
```
imsdu, line 109: We should be able to fetch the resource of a deprecated project.
Here also, we may want to have a specific revision

imsdu (changes requested):

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ResourcesDummy.scala
imsdu: Also we start with fetching the resource, it means the fetch of the project should not fail

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ResourcesDummy.scala
umbreak: No, deprecating a resource is not allowed when project is deprecated (at least not in v.1.4.x)

No, there is no immutability in terms of apimappings. This is metadata what we are talking about

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ResourcesDummy.scala
umbreak: No specific revision of project, but yes you're right, you can still fetch resources on deprecated projects

bogdanromanx (approved):

imsdu (approved):

Issue #1661, opened 2020-10-01
Have the possibility to use caller identities on CrossProject resolution


umbreak:
In some cases we want the CrossProjectResolver resolution depend on the current caller identities and not on the identities defined in the CrossProjectResolver. 

One use case for it comes from @annakristinkaufmann , where project `A` has a crossProjectResolver to resources on other projects and project `A` is used as an entry point to access all resources using the endpoints `/v1/resolvers/{org}/projectA/_/{resource_id}`.

One proposal to implement this feature, keeping backwards compatibility, would be to allow for empty identities on `CrossProjectResolver` creation. If no identities are defined, we will use the caller identities.

When identities are present, the resolution would use those identities to fetch resources and ignore the caller identities

imsdu:
Fixed in #1813

PR #1790, opened 2020-11-12
Defining resolver commands, events, states and rejections

imsdu:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,267 @@
...
+  def fetch(id: Iri, project: ProjectRef): UIO[Option[ResolverResource]]
+
+  /**
+    * Fetches the resolver at a given revision
+    * @param id      the resolver identifier
+    * @param project the project where the resolver will belong
+    * @param rev     the current revision of the resolver
+    */
+  def fetchAt(id: Iri, project: ProjectRef, rev: Long): IO[RevisionNotFound, Option[ResolverResource]]
+
```
umbreak, line 76: fetchBy is also necessary. You can fetch a resolver by a tag

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,267 @@
...
+    * Update an existing resolver
+    * @param id             the resolver identifier
+    * @param project        the project where the resolver will belong
+    * @param rev            the current revision of the resolver
+    * @param resolverFields the payload to update the resolver
+    */
+  def update(id: Iri, project: ProjectRef, rev: Long, resolverFields: ResolverFields)(implicit
+      caller: Caller
+  ): IO[ResolverRejection, ResolverResource]
+
```
umbreak, line 51: tag is also necessary. You can tag resolvers

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,267 @@
...
+  import ch.epfl.bluebrain.nexus.delta.sdk.utils.IOUtils.instant
+  import io.scalaland.chimney.dsl._
+
+  private[delta] def next(state: ResolverState, event: ResolverEvent): ResolverState = {
+    (state, event) match {
+      // Creating an in-project resolver
+      case (Initial, e: InProjectResolverCreated)                           =>
+        e.into[InProjectCurrent]
+          .withFieldConst(_.deprecated, false)
```
umbreak, line 112: It doesn't seem to bring much better readability than using current directly (`Current(id = ..., deprecated = ...)`)

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,267 @@
...
+
+      // Deprecating a in-project resolver
+      case (c: InProjectCurrent, e: ResolverDeprecated) if !c.deprecated    =>
+        c.copy(rev = e.rev, deprecated = true, updatedAt = e.instant, updatedBy = e.subject)
+      // Deprecating a cross-project resolver
+      case (c: CrossProjectCurrent, e: ResolverDeprecated) if !c.deprecated =>
+        c.copy(rev = e.rev, deprecated = true, updatedAt = e.instant, updatedBy = e.subject)
+
+      // Cases that should happen and that should have been filtered out by the evaluation
+      case (s, _)                                                           => s
```
umbreak, line 151: case _ => state

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverCommand.scala
```diff
@@ -0,0 +1,110 @@
...
+    */
+  sealed trait UpdateResolver extends ResolverCommand {
+
+    /**
+      * @return the last known revision of the resolver
+      */
+    def rev: Long
+  }
+
+  /**
```
umbreak, line 42: missing TagResolver

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverEvent.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+    * @return the project where the resolver belongs to
+    */
+  def project: ProjectRef
+
+}
+
+object ResolverEvent {
+
```
umbreak, line 30: missing `ResolverTagAdded`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverState.scala
```diff
@@ -0,0 +1,179 @@
...
+        id = id,
+        project = project,
+        priority = priority
+      )
+
+    override def toResource: Option[ResolverResource] =
+      Some(
+        ResourceF(
+          id = AccessUrl.resolver(project, id)(_).iri,
+          accessUrl = AccessUrl.resolver(project, id)(_).value,
```
umbreak, line 104: `AccessUrl.resolver(project, id)(_).shortForm(mappings)`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverState.scala
```diff
@@ -0,0 +1,179 @@
...
+  ) extends Current {
+
+    def resolver: InProjectResolver =
+      InProjectResolver(
+        id = id,
+        project = project,
+        priority = priority
+      )
+
+    override def toResource: Option[ResolverResource] =
```
umbreak, line 100: you need to pass the mappings

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,267 @@
...
+  def fetchAt(id: Iri, project: ProjectRef, rev: Long): IO[RevisionNotFound, Option[ResolverResource]]
+
+  /**
+    * A non terminating stream of events for resolvers. After emitting all known events it sleeps until new events
+    * are recorded.
+    *
+    * @param offset the last seen event offset; it will not be emitted by the stream
+    */
+  def events(offset: Offset = NoOffset): Stream[Task, Envelope[ResolverEvent]]
+
```
umbreak, line 84: add another for events for the current project:
```scala
def events(
      projectRef: ProjectRef,
      offset: Offset
  ): IO[WrappedProjectRejection, Stream[Task, Envelope[ResolverEvent]]]
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,267 @@
...
+  /**
+    * The current resolver events. The stream stops after emitting all known events.
+    *
+    * @param offset the last seen event offset; it will not be emitted by the stream
+    */
+  def currentEvents(offset: Offset = NoOffset): Stream[Task, Envelope[ResolverEvent]]
```
umbreak, line 90: we can drop this one, I don't know for what would be used

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,267 @@
...
+      /**
+        * Updating a resolver
+        */
+      // Update a non existing resolver
+      case (Initial, c: UpdateResolver)                            =>
+        IO.raiseError(ResolverNotFound(c.id, c.project))
+      // Invalid revision has been provided
+      case (s: Current, c: UpdateResolver) if c.rev != s.rev       =>
+        IO.raiseError(IncorrectRev(c.rev, s.rev))
+
```
umbreak, line 213: what about the case where current is already deprecated?

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,267 @@
...
+          c.into[ResolverDeprecated]
+            .withFieldConst(_.rev, s.rev + 1)
+            .withFieldConst(_.instant, now)
+            .withFieldConst(_.subject, caller.subject)
+            .transform
+        }
+
+    }
+
+    isOpen(command.project) >> stateMachine
```
umbreak, line 264: This check should be done in the `Resolvers` and not in the state machine, since otherwise in some cases you are gonna ask 2 times for the project (The method Resolvers.create(project, resolverFields) needs to make the same check to fetch the project in order to then compute the id)

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,267 @@
...
+      isOpen: ProjectRef => IO[ReadOnlyProject, Unit]
+  )(state: ResolverState, command: ResolverCommand)(implicit
+      caller: Caller,
+      clock: Clock[UIO]
+  ): IO[ResolverRejection, ResolverEvent] = {
+
+    // All provided identities must be part of the caller identities
+    def validateIdentities(identities: Set[Identity]): IO[ResolverRejection, Unit] = {
+      if (identities.isEmpty)
+        IO.raiseError(NoIdentities)
```
umbreak, line 165: I think we are going to allow empty identities in order to support [this](https://github.com/BlueBrain/nexus/issues/1661)

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverFields.scala
```diff
@@ -0,0 +1,52 @@
...
+
+import cats.data.{NonEmptyList, NonEmptySet}
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.sdk.model.identities.Identity
+import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectRef
+
+/**
+  * Payload for a resolver creation/update request
+  */
+trait ResolverFields extends Product with Serializable {
```
umbreak, line 11: Shouldn't it be `sealed`?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverRejection.scala
```diff
@@ -0,0 +1,92 @@
...
+    * Rejection returned when attempting to update/deprecate a resolver that is already deprecated.
+    *
+    * @param id the resolver identifier
+    */
+  final case class ResolverIsDeprecated(id: Iri) extends ResolverRejection(s"Resolver '$id' is deprecated.")
+
+  /**
+    * Rejection returned when the project or the organization the resolver belongs to has been deprecated
+    */
+  final case class ReadOnlyProject(project: ProjectRef)
```
umbreak, line 87: This is not the only error that can occur. It can also be that the project is not found.
However, why don't you just do:

```scala
  final case class WrappedProjectRejection(rejection: ProjectRejection) extends ResourceRejection(rejection.reason)
```

So that we don't end up having different client signaling messages for these things.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverState.scala
```diff
@@ -0,0 +1,179 @@
...
+
+  /**
+    * @return the schema reference that resolvers conforms to
+    */
+  final def schema: ResourceRef = Latest(schemas.resolvers)
+
+  /**
+    * @return the collection of known types of resolvers resources
+    */
+  final def types: Set[Iri] = Set(nxv.Schema)
```
umbreak, line 29: why there is a `nxv.Schema` here?

The types should be: nxv.Resolver plus `nxv.CrossProject` or `nxv.InProject`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverState.scala
```diff
@@ -0,0 +1,179 @@
...
+        priority = priority
+      )
+
+    override def toResource: Option[ResolverResource] =
+      Some(
+        ResourceF(
+          id = AccessUrl.resolver(project, id)(_).iri,
+          accessUrl = AccessUrl.resolver(project, id)(_).value,
+          rev = rev,
+          types = types,
```
umbreak, line 106:
```scala
types = Set(nxv.Resolver, nxv.InProject)
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverState.scala
```diff
@@ -0,0 +1,179 @@
...
```
umbreak, line 167:
```scala
types = Set(nxv.Resolver, nxv.CrossProject)
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverState.scala
```diff
@@ -0,0 +1,179 @@
...
+        projects = projects,
+        identities = identities,
+        priority = priority
+      )
+
+    override def toResource: Option[ResolverResource] =
+      Some(
+        ResourceF(
+          id = AccessUrl.resolver(project, id)(_).iri,
+          accessUrl = AccessUrl.resolver(project, id)(_).value,
```
umbreak, line 165:
```scala
AccessUrl.resolver(project, id)(_).shortForm(mappings)
```

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,267 @@
...
+        val missing = identities.diff(caller.identities)
+        if (missing.isEmpty) {
+          IO.unit
+        } else {
+          IO.raiseError(InvalidIdentities(missing))
+        }
+      }
+    }
+
+    val stateMachine = (state, command) match {
```
bogdanromanx, line 176: I think that having separate states, commands and events for variations of resolvers is making things much more complicated...

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
bogdanromanx: I don't think we're using that in 1.4.x series, do you think it's useful?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
bogdanromanx: Agree

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/TestHelpers.scala
```diff
@@ -28,6 +28,15 @@ trait TestHelpers extends ClasspathResourceUtils {
     inner("", length)
   }
 
+  /**
+    * Generate the cross product of the two given sequences
+    */
+  def crossProduct[A, B](first: Seq[A], second: Seq[B]): Seq[(A, B)] =
```
bogdanromanx, line 34: isn't the following equivalent?
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Apply.scala#L112

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
imsdu: If it is rewritten with a single state/command/event with a discriminator field, the complexity will appear anyway via this field no ?
And unused fields will appear for in-project resolvers.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
imsdu: Yep, it is not great, it seems a little more helpful in the evaluate method and not that much either

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
umbreak: I think it is: v1/resolvers/{org}/{proj}/events

But yeah, if we don't need it we can skip it

bogdanromanx: we don't have that according to the docs

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/Priority.scala
```diff
@@ -0,0 +1,29 @@
...
+object Priority {
+
+  private val min = 0
+  private val max = 1000
+
+  /**
+    * Attempts to get a priority from an integer
+    */
+  def apply(value: Int): Either[ResolverPriorityIntervalError, Priority] =
+    if (value >= min && value <= max)
```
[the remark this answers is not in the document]
imsdu: I remembered of this little guy:
```
Either.cond(
      value >= min && value <= max,
      new Priority(value),
      ResolverPriorityIntervalError(value, min, max)
    )
```

umbreak (changes requested):
It looks much better now. But we could improve it further by:

```scala
sealed trait ResolverValue extends Product with Serializable {
   def tpe: ResolverType
}

object ResolverValue {
    final case class InProject(priority: Priority) extends ResolverValue {
      override val: ResolverType = ResolverType.InProject
   }
    final case class CrossProject(
      resourceTypes: Set[Iri],
      projects: NonEmptyList[ProjectRef],
      identities: Set[Identity]
    ) extends ResolverValue {
      override val: ResolverType = ResolverType.CrossProject
   }
}
```

And then use this on the `ResolverEvent`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,301 @@
...
+    def validateCrossProjectSetup(
+        resolverType: ResolverType,
+        crossProjectSetup: Option[CrossProjectSetup]
+    ): IO[ResolverRejection, Unit] = {
+      resolverType match {
+        case InProject    => IO.unit
+        case CrossProject =>
+          crossProjectSetup
+            .map { setup =>
+              if (setup.identities.isEmpty)
```
umbreak, line 179: It is fine if we don't want to allow this yet to fix the issue I described, but can you then please add a `TODO`:

```scala
//TODO: In order to fix the issue https://github.com/BlueBrain/nexus/issues/1661 we should allow empty identities
```

umbreak (changes requested):

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/TestHelpers.scala
imsdu: It somehow clashed with a product method in scala.collection.
But (first, second).tupled do it fine

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
imsdu: I prefer to do everything in a separate PR

umbreak: Ok. Then add a TODO please, as I said in another comment

imsdu: I finally added it :)

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,292 @@
...
+      caller: Caller
+  ): IO[ResolverRejection, ResolverResource]
+
+  /**
+    * Create a new resolver with the provided id
+    * @param id             the resolver identifier
+    * @param project        the project where the resolver will belong
+    * @param resolverFields the payload to create the resolver
+    */
+  def create(id: Iri, project: ProjectRef, resolverFields: ResolverFields)(implicit
```
umbreak, line 40: As you can see in `Resources` this is not going to be `id: Iri` but `id: IdSegment`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,292 @@
...
+  ): IO[ResolverRejection, ResolverResource]
+
+  /**
+    * Update an existing resolver
+    * @param id             the resolver identifier
+    * @param project        the project where the resolver will belong
+    * @param rev            the current revision of the resolver
+    * @param resolverFields the payload to update the resolver
+    */
+  def update(id: Iri, project: ProjectRef, rev: Long, resolverFields: ResolverFields)(implicit
```
umbreak, line 51: same

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,292 @@
...
+  /**
+    * Add a tag to an existing resolver
+    *
+    * @param id        the resolver identifier
+    * @param project   the project where the resolver belongs
+    * @param tag       the tag name
+    * @param tagRev    the tag revision
+    * @param rev       the current revision of the resolver
+    */
+  def tag(id: Iri, project: ProjectRef, tag: Label, tagRev: Long, rev: Long)(implicit
```
umbreak, line 64: same

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,292 @@
...
+      caller: Subject
+  ): IO[ResolverRejection, ResolverResource]
+
+  /**
+    * Deprecate an existing resolver
+    * @param id      the resolver identifier
+    * @param project the project where the resolver belongs
+    * @param rev     the current revision of the resolver
+    */
+  def deprecate(id: Iri, project: ProjectRef, rev: Long)(implicit
```
umbreak, line 74: same

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,292 @@
...
+  def fetchAt(id: Iri, project: ProjectRef, rev: Long): IO[RevisionNotFound, Option[ResolverResource]]
+
+  /**
+    * Fetches a resolver by tag.
+    *
+    * @param id        the resolver identifier
+    * @param project   the project where the resolver belongs
+    * @param tag       the tag revision
+    */
+  def fetchBy(
```
umbreak, line 100: same

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,292 @@
...
+      id: Iri,
+      project: ProjectRef,
+      tag: Label
+  ): IO[ResolverRejection, Option[ResolverResource]]
```
umbreak, line 104: fetchBy can be implemented directly here since you have all necessary bits:
```scala
fetch(id, projectRef).flatMap {
      case Some(resource) =>
        resource.value.tags.get(tag) match {
          case Some(rev) => fetchAt(id, projectRef, rev).leftMap(_ => TagNotFound(tag))
          case None      => IO.raiseError(TagNotFound(tag))
        }
      case None           => IO.pure(None)
    }
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,292 @@
...
+      caller: Caller,
+      clock: Clock[UIO]
+  ): IO[ResolverRejection, ResolverEvent] = {
+
+    def validateResolverValue(value: ResolverValue): IO[ResolverRejection, Unit] =
+      value match {
+        case CrossProjectValue(_, _, _, identityResolution) =>
+          identityResolution match {
+            case UseCurrentCaller          => IO.unit
+            case ProvidedIdentities(value) =>
```
umbreak, line 173:
```scala
case ProvidedIdentities(value) if value.isEmpty => IO.raiseError(NoIdentities)
case ProvidedIdentities(value)  => ...
```

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,292 @@
...
+      // Deprecating a resolver
+      case (c: Current, e: ResolverDeprecated)                            =>
+        c.copy(rev = e.rev, deprecated = true, updatedAt = e.instant, updatedBy = e.subject)
+
+      // Cases that should happen and that should have been filtered out by the evaluation
+      case _                                                              => state
+    }
+  }
+
+  private[delta] def evaluate(state: ResolverState, command: ResolverCommand)(implicit
```
umbreak, line 163: I got used to the way we have done it for all the other state machines with inner methods...it helps with the readability (and then you don't need many of those comments). Many times when you need a comment, you can better do with a method :)

```scala
def evaluate(state: ResolverState, command: ResolverCommand) = {
   def create(c: CreateResolver): IO[ResolverRejection, ResolverCreated] = ???
   def update(c: UpdateResolver): IO[ResolverRejection, ResolverUpdated] = ???
   def tag(c: TagResolver): IO[ResolverRejection, ResolverTagAdded] = ???
   def(c: DeprecateResolver): IO[ResolverRejection, ResolverDeprecated] = ???

   command match {
      case c: CreateResolver    => create(c)
      case c: UpdateResolver    => update(c)
      case c: TagResolver       => tag(c)
      case c: DeprecateResolver => deprecate(c)
    }

}
```

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -0,0 +1,301 @@
...
+        * Tagging a resolver
+        */
+      // Resolver can't be found
+      case (Initial, c: TagResolver)                                              =>
+        IO.raiseError(ResolverNotFound(c.id, c.project))
+      // Invalid revision
+      case (s: Current, c: TagResolver) if c.rev != s.rev                         =>
+        IO.raiseError(IncorrectRev(c.rev, s.rev))
+      // Revision to tag is invalid
+      case (s: Current, c: TagResolver) if c.targetRev < 0 || c.targetRev > s.rev =>
```
umbreak, line 249: why do you need this `c.targetRev < 0` ? 
In this case, the state is `Current`, which in turn can never have rev =0. The only necessary check is `if c.targetRev >= s.rev`

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverFields.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+
+/**
+  * Fields for a resolver
+  * @param id                id of the resolver
+  * @param priority          resolution priority when attempting to find a resource
+  * @param value             additional fields to configure the resolver
+  */
+final case class ResolverFields(id: Option[Iri], priority: Priority, value: ResolverValue)
```
umbreak, line 11: So now the "input" ADT and the Events ADT share the same object `ResolverValue`. It is fine for me, but we have to be especially careful then with serialization in this case. 
It is better when we have different ones, so we can have separate serializations, but I understand that here it is a big overhead to pay

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
imsdu: Well, I thought it wasn't in the scope of this PR

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverFields.scala
imsdu: For ResolverFields, we have to implement a custom deserializer anyway.
For serializing the Resolver, a custom serializer will also be needed so as not to break the api.
It should help.

umbreak (approved):

umbreak (approved):

PR #1791, opened 2020-11-12
Added id expansion logic to Resources bundle

umbreak:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/IdSegment.scala
```diff
@@ -0,0 +1,88 @@
...
+
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.{ContextValue, JsonLdContext}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.IdSegment.{IriSegment, ResourceRefSegment, StringSegment}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.{ApiMappings, Project}
+
+/**
+  * A segment from the positional API that should be an Id
+  */
+sealed trait IdSegment extends Product with Serializable { self =>
```
imsdu, line 11: As segment comes from akka-http, something like RawId would maybe be better ?

```case class RawId[A](value: A)```

And then we would be two methods in the companion object:

```
def parseIri[A](id: Raw[A])(implicit iriDecoder: IriDecoder[A])
def parseRef[A](id: Raw[A])(implicit iriDecoder: IriDecoder[A]) = parseIri(id).map(ResourceRef(_))
```

A could be either a String or an Iri as getting a ResourceRef for an Iri never fails, we can map the Iris to a ResourceRef not in the routes but in Resources if needed

**Review by imsdu — CHANGES_REQUESTED** [review]

File: delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resources/ResourcesImpl.scala [hunk 2]
```diff
@@ -30,123 +31,137 @@ final class ResourcesImpl private (
     projects: Projects,
     eventLog: EventLog[Envelope[ResourceEvent]]
 )(implicit rcr: RemoteContextResolution, uuidF: UUIDF)
-    extends Resources {
+    extends ResourcesCommons(projects)
```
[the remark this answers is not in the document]
umbreak: It is either having this class or just copy pasting all the methods in both `ResourcesDummy` or `ResourcesImpl`. Both options are ok for me

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/IdSegment.scala
umbreak: We need the 3 cases, since `ResourceRef` and `Iri` might not be the same. Imagine someone from some plugin is using this `Resources` and wants to pass a ResourceRef.Tag, not an `Iri`

umbreak: And yes I wanted to call it `IdSegment` precisely because akka http has `Segment`.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resources/ResourcesImpl.scala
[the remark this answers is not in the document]
bogdanromanx: I'm not fond of an abstract class either; I would also prefer an object where the common functionality can be defined; it can even be the companion object of Resources with private[delta].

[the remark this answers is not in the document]
umbreak: It is not an object, since it needs the `projects: Projects` and all this implicit parameters. So just having random methods in `Resources` companion object which do nothing there is not really useful, nor clear either.

I will rather duplicate all those methods then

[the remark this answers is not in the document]
imsdu: projects could stays in ResourcesImpl and the other stuff go to the object no ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/IdSegment.scala
imsdu: My thinking about the naming was that so far, it is called by the routes but it could be called by other modules too and Segment might not be so relevant but it is ok (and naming is a hard thing)

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/IdSegment.scala
```diff
@@ -0,0 +1,88 @@
...
+}
+
+sealed trait IdSegmentExtractor[A <: IdSegment] {
+  private[model] def toIri(segment: A): Option[Iri]
+  private[model] def toResourceRef(segment: A): Option[ResourceRef]
+}
+
+object IdSegmentExtractor {
+
+  implicit def idSegmentExtractor(implicit project: Project): IdSegmentExtractor[IdSegment] =
```
imsdu, line 67: It may be better to break it in 3 as project is not needed for 2 of them

imsdu (changes requested):

PR #1792, opened 2020-11-13
Apply authorization to organizations and projects listings.

### LINKED ISSUE #1781: Apply filtering based on authorization on organization listings

### LINKED ISSUE #1782: Apply filtering based on authorization on project listings

wwajerowicz:
Fixes #1781 
Fixes #1782

### DIFF
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/OrganizationsRoutes.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/OrganizationsRoutes.scala
@@ -16,6 +16,7 @@ import ch.epfl.bluebrain.nexus.delta.sdk.directives.AuthDirectives
 import ch.epfl.bluebrain.nexus.delta.sdk.error.ServiceError.AuthorizationFailed
 import ch.epfl.bluebrain.nexus.delta.sdk.model.BaseUri
 import ch.epfl.bluebrain.nexus.delta.sdk.model.acls.AclAddress
+import ch.epfl.bluebrain.nexus.delta.sdk.model.acls.AclAddressFilter.AnyOrganization
 import ch.epfl.bluebrain.nexus.delta.sdk.model.identities.Caller
 import ch.epfl.bluebrain.nexus.delta.sdk.model.organizations.OrganizationRejection._
 import ch.epfl.bluebrain.nexus.delta.sdk.model.organizations.{Organization, OrganizationRejection}
@@ -52,9 +53,17 @@ final class OrganizationsRoutes(identities: Identities, organizations: Organizat
   import baseUri.prefixSegment
   implicit val orgContext: ContextValue = Organization.context
 
-  private def orgsSearchParams: Directive1[OrganizationSearchParams] =
-    searchParams.tmap { case (deprecated, rev, createdBy, updatedBy) =>
-      OrganizationSearchParams(deprecated, rev, createdBy, updatedBy)
+  private def orgsSearchParams(implicit caller: Caller): Directive1[OrganizationSearchParams] =
+    searchParams.tflatMap { case (deprecated, rev, createdBy, updatedBy) =>
+      onSuccess(acls.listSelf(AnyOrganization(true)).runToFuture).map { aclsCol =>
+        OrganizationSearchParams(
+          deprecated,
+          rev,
+          createdBy,
+          updatedBy,
+          org => aclsCol.exists(caller.identities, orgs.read, AclAddress.Organization(org.label))
+        )
+      }
     }
 
   private def fetchByUUID(uuid: UUID, permission: Permission)(implicit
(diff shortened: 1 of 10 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/search/SearchParams.scala
```diff
@@ -45,13 +47,15 @@ object SearchParams {
...
+    * @param filter     the additional filter to select realms
     */
   final case class RealmSearchParams(
       issuer: Option[String] = None,
       deprecated: Option[Boolean] = None,
       rev: Option[Long] = None,
       createdBy: Option[Subject] = None,
-      updatedBy: Option[Subject] = None
+      updatedBy: Option[Subject] = None,
+      filter: Realm => Boolean = _ => true
```
imsdu, line 58: I think it is better not put a default value on this param as it is the most critical one

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/search/SearchParams.scala
```diff
@@ -76,12 +80,14 @@ object SearchParams {
...
     * @param updatedBy  the optional subject who updated the resource
+    * @param filter     the additional filter to select organizations
     */
   final case class OrganizationSearchParams(
       deprecated: Option[Boolean] = None,
       rev: Option[Long] = None,
       createdBy: Option[Subject] = None,
-      updatedBy: Option[Subject] = None
+      updatedBy: Option[Subject] = None,
+      filter: Organization => Boolean = _ => true
```
imsdu, line 90: Same

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/search/SearchParams.scala
```diff
@@ -103,13 +109,15 @@ object SearchParams {
...
+    * @param filter       the additional filter to select projects
     */
   final case class ProjectSearchParams(
       organization: Option[Label] = None,
       deprecated: Option[Boolean] = None,
       rev: Option[Long] = None,
       createdBy: Option[Subject] = None,
-      updatedBy: Option[Subject] = None
+      updatedBy: Option[Subject] = None,
+      filter: Project => Boolean = _ => true
```
imsdu, line 120: Same

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/OrganizationsRoutes.scala
```diff
@@ -52,9 +53,17 @@ final class OrganizationsRoutes(identities: Identities, organizations: Organizat
   import baseUri.prefixSegment
   implicit val orgContext: ContextValue = Organization.context
 
-  private def orgsSearchParams: Directive1[OrganizationSearchParams] =
-    searchParams.tmap { case (deprecated, rev, createdBy, updatedBy) =>
-      OrganizationSearchParams(deprecated, rev, createdBy, updatedBy)
+  private def orgsSearchParams(implicit caller: Caller): Directive1[OrganizationSearchParams] =
+    searchParams.tflatMap { case (deprecated, rev, createdBy, updatedBy) =>
+      onSuccess(acls.listSelf(AnyOrganization(true)).runToFuture).map { aclsCol =>
```
imsdu, line 58: How was working the search before ? If a user has no orgs/read, we returned a 403, a 404 or a empty list with a 200 ?

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/search/SearchParams.scala
wwajerowicz: It's more consistent with other params which are all optional and you can override it only when you need it like the other params.

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/OrganizationsRoutes.scala
wwajerowicz: If user has no `org/read` anywhere we return empty list.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/search/SearchParams.scala
imsdu: The default value here can give you access to resources you shouldn't have access to and it can be easy to misuse it silently.
That is why I wouldn't put a default value like the others because if we screw up for the others, the result is false but doesn't expose data you should not have access to.

wwajerowicz: I guess it would make sense to not have the default for `OrganizationSearchParams` and `ProjectSearchParams`, but for the `RealmSearchParams` we'll probably never going to use the filter, so it make sense to have a default there.

wwajerowicz: In this case we probably also need to remove the default value for organizationParams in this function
```scala
 def list(
      pagination: FromPagination,
      params: OrganizationSearchParams = OrganizationSearchParams.none
  ): UIO[UnscoredSearchResults[OrganizationResource]]
```
 in `Organizations` since there is no logical default anymore.
And similarly in `Projects`.

imsdu (approved):

PR #1793, opened 2020-11-13
Refactored DeltaDirectives

umbreak:
Refactored DeltaDirectives, splitting the responsibility into 2 different areas:
- **UriDirectives**: Directives related to Uri extractions (path segments and query parameters). I've increased the tests for these directives.
- **DeltaDirectives**: Directives related to routes. In this regard, I've refactored the code using the [magnet pattern](http://spray.io/blog/2012-12-13-the-magnet-pattern/) in order to get rid of all the different method names we had in order to complete a route (12 different name variations and more to come...). Now there are just 2 methods and they support a richer pool of cases:
   - **emit**
      - UIO[A]
      - UIO[Option[A]]
      - UIO[SearchResults[A]]
      - IO[E, A]
      - Stream[Task, Envelope[E]]
      - IO[E, Stream[Task, Envelope[E]]]
      - A
   - **discardEntityAndEmit** supporting:
      - UIO[A]
      - UIO[Option[A]]
      - A

Both **emit** and **discardEntityAndEmit** support changing the status code.


delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/direcives/DeltaDirectivesSpec.scala
```diff
@@ -1,4 +1,4 @@
-package ch.epfl.bluebrain.nexus.delta.routes
+package ch.epfl.bluebrain.nexus.delta.routes.direcives
```
imsdu, line 1: directives

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/directives/DeltaDirectives.scala
```diff
@@ -0,0 +1,419 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{Envelope, Event}
+import ch.epfl.bluebrain.nexus.delta.syntax._
+import fs2.Stream
+import monix.bio.{IO, Task, UIO}
+import monix.execution.Scheduler
+import streamz.converter._
+
+import scala.util.Try
+
+object DeltaDirectives extends UriDirectives with RdfMarshalling {
```
imsdu, line 35: Some documentation in the second part could help better understand and separate more the different blocks

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/directives/DeltaDirectives.scala
umbreak: For now I've changed the method names so it should be quite self explanatory looking at it and at the parameters it requires.

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/directives/DeltaDirectives.scala
```diff
@@ -0,0 +1,419 @@
...
+    implicit def IOStreamSupport[E: JsonLdEncoder: HttpResponseFields, A <: Event: JsonLdEncoder](
+        io: IO[E, Stream[Task, Envelope[A]]]
+    )(implicit s: Scheduler, jo: JsonKeyOrdering, cr: RemoteContextResolution): ToResponseJsonLd =
+      new ToResponseJsonLd {
+        override def apply(statusOverride: Option[StatusCode]): Route =
+          toResponseJsonLd(statusOverride.getOrElse(OK), io)
+      }
+  }
+
+  trait LowPrioToResponseJsonLd {
```
imsdu, line 320: It should be sealed no ?

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/directives/DeltaDirectives.scala
imsdu: Ok, it was also because at least when reviewing due to the github layout, it is not so easy to see where starts and ends the main parts ToResponse/DiscardResponse/...

bogdanromanx (approved):

imsdu (approved):

PR #1794, opened 2020-11-16
Added resource routes

umbreak:
Related to https://github.com/BlueBrain/nexus/issues/1771


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -0,0 +1,196 @@
...
+                }
+              }
+            )
+          }
+        )
+      }
+    )
+
+  private def fetchProject(orgUuid: UUID, projectUuid: UUID): Directive1[ProjectResource] =
+    onSuccess(projects.fetch(projectUuid).runToFuture).flatMap {
```
imsdu, line 167: Only the ref is interesting here, we should create a method findRefByUuid that return the ref from the uuids.
This way, we only aim for the cache once.
It is something we will need for other resources so we could put it in an object

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -0,0 +1,196 @@
...
+      },
+      // Create a resource without schema nor id segment
+      (post & noParameter("rev") & pathEndOrSingleSlash & entity(as[Json])) { source =>
+        operationName(s"$prefixSegment/resources/{org}/{project}") {
+          authorizeFor(AclAddress.Project(ref), resourcePermissions.write).apply {
+            emit(StatusCodes.Created, resources.create(ref, resourceSchema, source).map(_.void))
+          }
+        }
+      },
+      idSegment { schemaSegment =>
```
imsdu, line 102: If we could parse to something like a NoSegment, couldn't we parse the two segments at the same time and make the routes a little simpler ?

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -0,0 +1,196 @@
...
+                      emit(resources.events(offset))
+                    }
+                  }
+                }
+              }
+            },
+            projectRef { ref =>
+              routesForProject(ref)
+            },
+            (uuid & uuid) { (orgUuid, projectUuid) =>
```
imsdu, line 71: We could have a directive projectRefFromUuid

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -0,0 +1,196 @@
...
+            authorizeFor(AclAddress.Project(ref), events.read).apply {
+              lastEventId { offset =>
+                emit(resources.events(ref, offset).leftWiden[ResourceRejection])
+              }
+            }
+          }
+        }
+      },
+      // Create a resource without schema nor id segment
+      (post & noParameter("rev") & pathEndOrSingleSlash & entity(as[Json])) { source =>
```
bogdanromanx, line 95: I think currently we also allow `POST /resources/{org}/{proj}/_`

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -0,0 +1,196 @@
...
+
+  private val resourceSchema = IriSegment(schemas.resources)
+
+  def routes: Route =
+    baseUriPrefix(baseUri.prefix) {
+      extractCaller { implicit caller =>
+        pathPrefix("resources") {
+          concat(
+            // SSE resources
+            (pathPrefix("events") & pathEndOrSingleSlash) {
```
bogdanromanx, line 57: we seem to be missing `GET /resources/{org}/events`

delta/app/src/test/resources/organizations/org-resource.json
```diff
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://bluebrain.github.io/nexus/contexts/resource.json",
-    "https://bluebrain.github.io/nexus/contexts/organizations.json"
+    "https://bluebrain.github.io/nexus/contexts/organizations.json",
+    "https://bluebrain.github.io/nexus/contexts/resource.json"
```
bogdanromanx, line 4: we should standardize on plural

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
umbreak: I think it is perfectly fine as is. More abstraction is just going to make it more complicated. Besides now it is in order:

```scala
idSegment { schemaSegment =>
   // something to do after only the schema segment has been consumed
   idSegment { idSegment =>
      // something to do after both schema and id segment have been consumed
   }
}
```

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
umbreak: Yes, but would it make things clearer? I find this line already very explicit :)

bogdanromanx (changes requested):
Looks good, some small things...

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
imsdu: I just find the route a little complicated to understand and am trying to find a way to make it a little simpler. But yes, my solution may not be a good one

delta/app/src/test/resources/organizations/org-resource.json
umbreak: Notice that resource is just the nexus metadata.  It is not a domain resource as organizations, projects, schemas,  etc. So maybe we can call it differently. Ideas?

bogdanromanx: hmmm, good point; maybe renaming it to `metadata.json`?

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
umbreak: Yes, this is already supported: https://github.com/BlueBrain/nexus/blob/74b8a7ec73bd1866e8c65b6b0a8cbe0529b6b3a6/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala#L106

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -65,109 +62,115 @@ final class ResourcesRoutes(
...
-          (post & noParameter("rev") & pathEndOrSingleSlash) {
-            operationName(s"$prefixSegment/resources/{org}/{project}/{schema}") {
-              authorizeFor(AclAddress.Project(ref), resourcePermissions.write).apply {
-                entity(as[Json]) { source =>
-                  emit(StatusCodes.Created, resources.create(ref, schemaSegment, source).map(_.void))
+            // SSE resources for all events belonging to an organization
+            ((label | orgLabelFromUuidLookup) & pathPrefix("events") & pathEndOrSingleSlash) { org =>
+              get {
+                operationName(s"$prefixSegment/resources/{org}/events") {
+                  authorizeFor(AclAddress.Root, events.read).apply {
```
bogdanromanx, line 69: I think the scope of the permission should be `AclAddress.Organization(org)`

bogdanromanx (changes requested):

bogdanromanx (approved):

imsdu (approved):

PR #1796, opened 2020-11-17
Wiring resources

umbreak:
Fixes to https://github.com/BlueBrain/nexus/issues/1771

- Added custom Kryo serialization for `Path` and `IRI`. Probably a Jena Model can be added in the future too.
- Simplified the passing of contexts in RDF using `ContextValue`.
- Added `/source` missing endpoint to resource routes.
- Added default context to compute compacted and expanded when source does not have a context.
- Added wiring for resource routes.

### DIFF
--- a/delta/app/src/main/resources/kamon.conf
+++ b/delta/app/src/main/resources/kamon.conf
@@ -62,6 +62,17 @@ kamon {
   }
 
   modules {
+
+    host-metrics {
+      # Set to 'no' might help when running it locally with Mac Os 11
+      enabled = no
+    }
+
+    process-metrics {
+      # Set to 'no' might help when running it locally with Mac Os 11
+      enabled = no
+    }
+
     jaeger {
       enabled = ${app.monitoring.jaeger.enabled}
     }
--- a/delta/app/src/main/resources/kamon.conf
+++ b/delta/app/src/main/resources/kamon.conf
@@ -65,12 +65,12 @@ kamon {
 
     host-metrics {
       # Set to 'no' might help when running it locally with Mac Os 11
-      enabled = no
+      enabled = yes
     }
 
     process-metrics {
       # Set to 'no' might help when running it locally with Mac Os 11
-      enabled = no
+      enabled = yes
     }
 
     jaeger {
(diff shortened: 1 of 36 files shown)


delta/app/src/main/resources/kamon.conf
```diff
@@ -62,6 +62,17 @@ kamon {
   }
 
   modules {
+
+    host-metrics {
+      # Set to 'no' might help when running it locally with Mac Os 11
+      enabled = no
```
bogdanromanx, line 68: I don't these should be the defaults...

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ResourceSourceParser.scala
```diff
@@ -28,39 +31,58 @@ trait ResourceSourceParser {
...
+  }
 
   private def checkSameId(iri: Iri, expanded: ExpandedJsonLd): IO[UnexpectedResourceId, Unit] =
     expanded.rootId.asIri match {
       case Some(sourceId) if sourceId != iri => IO.raiseError(UnexpectedResourceId(iri, sourceId))
       case _                                 => IO.unit
     }
+
+  private def ctxOrProjectDefaults(source: Json, project: Project): (ContextValue, Json) =
+    source.topContextValueOrEmpty match {
```
bogdanromanx, line 75: I'm wondering if we should check first to see if we have any triples... what if the json is fully expanded?

bogdanromanx (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ResourceSourceParser.scala
umbreak: What is the problem if the json is fully expanded? it will stay expanded as is

bogdanromanx: We would unnecessarily add a context.

imsdu (dismissed):

bogdanromanx (approved):

imsdu (dismissed):

Issue #1795, opened 2020-11-17
Initial ACLs only applied in certain cases, inconsistent behaviour 


umbreak:
When the underlying API calls the `fetchWithAncestors` it works as expected:
```
$ curl -s "http://localhost:8080/v1/acls?ancestors=true" | jq
{
  "@context": [
    "https://bluebrain.github.io/nexus/contexts/metadata.json",
    "https://bluebrain.github.io/nexus/contexts/search.json",
    "https://bluebrain.github.io/nexus/contexts/acls.json"
  ],
  "_results": [
    {
      "@id": "http://localhost:8080/v1/acls",
      "@type": "AccessControlList",
      "_path": "/",
      "acl": [
        {
          "identity": {
            "@id": "http://localhost:8080/v1/anonymous",
            "@type": "Anonymous"
          },
          "permissions": [
            "realms/write",
            ...
          ]
        }
      ],
      "_rev": 0,
      "_deprecated": false,
      "_createdAt": "1970-01-01T00:00:00Z",
      "_createdBy": "http://localhost:8080/v1/anonymous",
      "_updatedAt": "1970-01-01T00:00:00Z",
      "_updatedBy": "http://localhost:8080/v1/anonymous",
      "_constrainedBy": "https://bluebrain.github.io/nexus/schemas/acls.json",
      "_self": "http://localhost:8080/v1/acls"
    }
  ],
  "_total": 1
}
```

The other cases do not work:

```
$ curl -s "http://localhost:8080/v1/acls/*/*?ancestors=true" | jq
{
  "@context": [
    "https://bluebrain.github.io/nexus/contexts/metadata.json",
    "https://bluebrain.github.io/nexus/contexts/search.json",
    "https://bluebrain.github.io/nexus/contexts/acls.json"
  ],
  "_results": [],
  "_total": 0
}


$ curl -s "http://localhost:8080/v1/acls?ancestors=false" | jq
{
  "@context": [
    "https://bluebrain.github.io/nexus/contexts/metadata.json",
    "https://bluebrain.github.io/nexus/contexts/search.json",
    "https://bluebrain.github.io/nexus/contexts/acls.json"
  ],
  "_results": [],
  "_total": 0
}
```

In order to support initial ACLs, this logic has to be present in any fetch/listing method

PR #1800, opened 2020-11-18
Fixed initial ACLs on /

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1795

### DIFF
--- a/delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/AclsDummy.scala
+++ b/delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/AclsDummy.scala
@@ -36,10 +36,7 @@ final class AclsDummy private (
   private val minimum: Set[Permission] = permissions.minimum
 
   override def fetch(address: AclAddress): UIO[Option[AclResource]] =
-    cache.get.map(_.value.get(address)).map {
-      case Some(value) => Some(value)
-      case None        => Initial.toResource(address, minimum)
-    }
+    cache.get.map(_.value.get(address).orElse(Initial.toResource(address, minimum)))
 
   override def fetchWithAncestors(address: AclAddress): UIO[AclCollection] =
     fetch(address).flatMap { resourceOpt =>
@@ -53,10 +50,7 @@ final class AclsDummy private (
   override def fetchAt(address: AclAddress, rev: Long): IO[RevisionNotFound, Option[AclResource]] =
     journal
       .stateAt(address, rev, Initial, Acls.next, RevisionNotFound.apply)
-      .map {
-        case None     => Initial.toResource(address, minimum)
-        case stateOpt => stateOpt.flatMap(_.toResource(address, minimum))
-      }
+      .map(stateOpt => stateOpt.getOrElse(Initial).toResource(address, minimum))
 
   override def list(filter: AclAddressFilter): UIO[AclCollection] =
     cache.get.map(_.fetch(filter)).map { col =>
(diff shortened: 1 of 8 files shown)


delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
```diff
@@ -31,11 +32,19 @@ final class AclsImpl private (
     index: AclsCache
 ) extends Acls {
 
+  private val minimum: Set[Permission] = permissions.minimum
```
wwajerowicz, line 35: This can change between being assigned and being used.

wwajerowicz (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
umbreak: I provided the same logic as in [1.4.x](https://github.com/BlueBrain/nexus/blob/v1.4.x/delta/src/main/scala/ch/epfl/bluebrain/nexus/iam/acls/package.scala#L50) which I think it is the correct anyway, since we want immutability. We don't want to fetch the ACLs and get different results from the same revision depending on what we do with the permissions collection

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/AclsDummy.scala
```diff
@@ -32,19 +33,39 @@ final class AclsDummy private (
 )(implicit clock: Clock[UIO])
     extends Acls {
 
+  private val minimum: Set[Permission] = permissions.minimum
```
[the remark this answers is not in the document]
umbreak: Same reasoning

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/AclsDummy.scala
```diff
@@ -32,19 +33,39 @@ final class AclsDummy private (
...
+
   override def fetch(address: AclAddress): UIO[Option[AclResource]] =
-    cache.get.map(_.value.get(address))
+    cache.get.map(_.value.get(address)).map {
```
imsdu, line 39: value should have a getOrElse method

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
```diff
@@ -31,11 +32,19 @@ final class AclsImpl private (
...
+
   override def fetch(address: AclAddress): UIO[Option[AclResource]] =
-    agg.state(address.string).map(_.toResource).named("fetchAcl", moduleType)
+    agg.state(address.string).map(_.toResource(address, minimum)).named("fetchAcl", moduleType)
 
   override def fetchWithAncestors(address: AclAddress): UIO[AclCollection] =
```
imsdu, line 40: It looks identical to the method in the dummy, maybe we could put it directly in the Acls ?

imsdu (changes requested):

wwajerowicz (dismissed):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/acls/AclsImpl.scala
umbreak: There is few duplication, but in this case is minimal. I don't think it really helps for readability to put it in Acls.

imsdu (approved):

bogdanromanx (changes requested):
I think we need to add some additional tests.

PR #1802, opened 2020-11-18
Minor integration tests fixes.

wwajerowicz:


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/admin/OrgsSpec.scala
```diff
@@ -282,13 +282,6 @@ class OrgsSpec extends BaseSpec with EitherValuable {
       )
     }
 
-    "fail when wrong revision is provided" taggedAs OrgsTag in {
```
bogdanromanx, line 285: why are you removing this test?

tests/src/test/resources/iam/realms/fetch-updated-response.json
```diff
@@ -1,7 +1,7 @@
 {
-  "@context": [
```
bogdanromanx, line 2: most json files require formatting

bogdanromanx (changes requested):

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/admin/OrgsSpec.scala
wwajerowicz: I thought it was duplicate of the test from 193, but it's actually different. I'll add it back.

umbreak (approved):

bogdanromanx (approved):

PR #1803, opened 2020-11-19
Created ElasticSearch plugin STM

### LINKED ISSUE #1494: ES Plugin: define state machine and api

bogdanromanx:
Fixes #1494


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchPlugin.scala
```diff
@@ -0,0 +1,22 @@
...
+class ElasticSearchPlugin extends Plugin {
+  /**
+    * Plugin information
+    */
+  override def info: PluginInfo = ???
+
+  /**
+    * Optional routes provided by the plugin.
+    */
+  override def route: Option[Route] = ???
```
umbreak, line 16: We would also need some sort of priority? When we assemble the routes in the core, we should probably pick up all the routes from plugins and core in something like:
```scala
val routes: List[(Priority, Route)] = ???
va sortedRoutes = routes.sortBy{ case (prio, _) => prio }.map { case (_, route) => route}
```

umbreak: Alternatively, the priority  information can also be in `PluginInfo`

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
```diff
@@ -0,0 +1,130 @@
...
+
+    event match {
+      case e: ElasticSearchViewCreated    => created(e)
+      case e: ElasticSearchViewUpdated    => updated(e)
+      case e: ElasticSearchViewTagAdded   => tagAdded(e)
+      case e: ElasticSearchViewDeprecated => deprecated(e)
+    }
+  }
+
+  private[elasticsearch] def evaluate(permissions: UIO[Permissions])(
```
umbreak, line 57: I don't think we need the UIO here. I got rid of it in the previous PR on Acls state machine

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
```diff
@@ -0,0 +1,130 @@
...
+      state: ElasticSearchViewState,
+      cmd: ElasticSearchViewCommand
+  )(implicit clock: Clock[UIO] = IO.clock): IO[ElasticSearchViewRejection, ElasticSearchViewEvent] = {
+
+    def validate(value: ElasticSearchViewValue): IO[ElasticSearchViewRejection, Unit] =
```
umbreak, line 62: `validatePermissions` or `checkPermissions`? validate could just be anything

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewType.scala
```diff
@@ -0,0 +1,34 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.model
+
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.nxv
+
+/**
+  * Enumeration of ElasticSearch view types.
+  */
+sealed trait ElasticSearchViewType extends Product with Serializable {
```
umbreak, line 9: It seems an overkill. This is usually done just with plain Vocabulary

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewCommand.scala
```diff
@@ -0,0 +1,101 @@
...
+    * @param id      the view id
+    * @param project a reference to the parent project
+    * @param value   the view configuration
+    * @param source  the original json value provided by the caller
+    * @param subject the identity associated with this command
+    */
+  final case class CreateElasticSearchView(
+      id: Iri,
+      project: ProjectRef,
+      value: ElasticSearchViewValue,
```
umbreak, line 44: shouldn't the command contain just the Json and inside the state machine you should try to convert this json to a `ElasticSearchViewValue`?

This will simplify the routes, having them just get a `entity(as[Json])` and the rest of the logic can be done in the state machine

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewCommand.scala
```diff
@@ -0,0 +1,101 @@
...
+    * @param project a reference to the parent project
+    * @param value   the view configuration
+    * @param rev     the last known revision of the view
+    * @param source  the original json value provided by the caller
+    * @param subject the identity associated with this command
+    */
+  final case class UpdateElasticSearchView(
+      id: Iri,
+      project: ProjectRef,
+      value: ElasticSearchViewValue,
```
umbreak, line 62: same

umbreak (changes requested):

imsdu (changes requested):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
```diff
@@ -0,0 +1,130 @@
...
+            perms <- permissions
+            set   <- perms.fetchPermissionSet
+            _     <- if (set.contains(v.permission)) IO.unit else IO.raiseError(PermissionIsNotDefined(v.permission))
+          } yield ()
+      }
+
+    def create(c: CreateElasticSearchView) = state match {
+      case Initial =>
+        for {
+          _ <- validate(c.value)
```
umbreak, line 76: On creation we might want to validate if the ES index already exists. But since the indices are UUID values, maybe that's not necessary

umbreak (changes requested):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewCommand.scala
imsdu: We could put the json in the value no ?

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchPlugin.scala
bogdanromanx: that could possibly be useful, but it's not in the scope of this PR... I've just implemented the plugin contract with `???`

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewType.scala
bogdanromanx: I used the same structure like for resolvers... and it's useful for pattern matching.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewCommand.scala
bogdanromanx: it's simple to return back to the client the deserialization errors directly from the routes; we could put the json directly in the value, but it will generate slightly more noise in its use; I would keep it as is.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchPluginDef.scala
```diff
@@ -0,0 +1,26 @@
...
+    */
+  override def info: PluginInfo = ???
+
+  /**
+    * Initialize the plugin.
+    *
+    * @param locator distage dependencies [[Locator]]
+    * @return [[Plugin]] instance.
+    */
+  override def initialise(locator: Locator): Task[Plugin] = ???
```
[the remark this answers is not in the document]
bogdanromanx: this is the contract of the plugindef... I won't modify it just yet

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  private[elasticsearch] def evaluate(permissions: UIO[Permissions])(
+      state: ElasticSearchViewState,
+      cmd: ElasticSearchViewCommand
+  )(implicit clock: Clock[UIO] = IO.clock): IO[ElasticSearchViewRejection, ElasticSearchViewEvent] = {
+
+    def validate(value: ElasticSearchViewValue): IO[ElasticSearchViewRejection, Unit] =
+      value match {
+        case _: AggregateElasticSearchViewValue => IO.unit
+        case v: IndexingElasticSearchViewValue  =>
```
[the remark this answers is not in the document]
bogdanromanx: any suggestions on how to do that?

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  private[elasticsearch] def evaluate(permissions: UIO[Permissions])(
+      state: ElasticSearchViewState,
+      cmd: ElasticSearchViewCommand
+  )(implicit clock: Clock[UIO] = IO.clock): IO[ElasticSearchViewRejection, ElasticSearchViewEvent] = {
+
+    def validate(value: ElasticSearchViewValue): IO[ElasticSearchViewRejection, Unit] =
+      value match {
+        case _: AggregateElasticSearchViewValue => IO.unit
```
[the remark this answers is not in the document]
bogdanromanx: what do you mean by this?

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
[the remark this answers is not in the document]
umbreak: How it is done in 1.4.x. First you need to have an ElasticSearch client. Then you do the following
1. Create an index
2. [Put the mappings](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-mapping.html)
3. The response will tell you if the mapping is correct or not.
4. Delete the index

You can find it in the 1.4.x brach.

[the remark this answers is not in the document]
imsdu: I don't know if any lib do that, if there is a json schema or something else, in the worst case, we can create a temporary index and delete it afterwards.
For now, I would only pass a function to evaluate like we did with realm to resolve the wellKnown.
Implement the real function in another PR.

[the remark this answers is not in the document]
bogdanromanx: the problem with this is the availability of the ES cluster; I would prefer a static validation, but there's no json schema available; maybe an AST?

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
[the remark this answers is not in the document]
imsdu: Shouldn't all the ViewRef point to an existing view ?

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
bogdanromanx: I think the management of indices is part of the indexing logic; I would not put it here.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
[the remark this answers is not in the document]
umbreak: If there is static validation go for it, otherwise that's the way we have been doing it and it works.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
```diff
@@ -0,0 +1,144 @@
...
+      clock: Clock[UIO] = IO.clock,
+      uuidF: UUIDF = UUIDF.random
+  ): IO[ElasticSearchViewRejection, ElasticSearchViewEvent] = {
+
+    def validate(id: Iri, project: ProjectRef, value: ElasticSearchViewValue): IO[ElasticSearchViewRejection, Unit] =
+      value match {
+        case v: AggregateElasticSearchViewValue =>
+          val ref = ViewRef(project, id)
+          for {
+            _ <- if (v.views.contains(ref)) IO.raiseError(CannotSelfReference(ref)) else IO.unit
```
umbreak, line 74: So far an `AggregatedElasticSearchView` can only reference to `ElasticSearchView`(s), not to other `AggregatedElasticSearchView`(s). Being that the case, this extra check is irrelevant, since `validateRef` function should already be checking that.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
```diff
@@ -0,0 +1,144 @@
...
+      case e: ElasticSearchViewCreated    => created(e)
+      case e: ElasticSearchViewUpdated    => updated(e)
+      case e: ElasticSearchViewTagAdded   => tagAdded(e)
+      case e: ElasticSearchViewDeprecated => deprecated(e)
+    }
+  }
+
+  private[elasticsearch] def evaluate(
+      validatePermission: Permission => IO[PermissionIsNotDefined, Unit],
+      validateMapping: Json => IO[InvalidElasticSearchMapping, Unit],
```
umbreak, line 62: just a function from Json to a result? How are you going to specify the actual elasticsearch index where this mapping is going to be validated against? It should be:
```scala
type ValidateMapping: (ElasticSearchIndex, Json) => IO[InvalidElasticSearchMapping, Unit]
```

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchView.scala
```diff
@@ -0,0 +1,91 @@
...
+      id: Iri,
+      project: ProjectRef,
+      uuid: UUID,
+      resourceSchemas: Set[Iri],
+      resourceTypes: Set[Iri],
+      resourceTag: Option[Label],
+      sourceAsText: Boolean,
+      includeMetadata: Boolean,
+      includeDeprecated: Boolean,
+      mapping: Json,
```
imsdu, line 68: It would be nice to have like we do with Label/String to have an ElasticsearchMapping type to show that this has been validated and is not a simple json

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
imsdu, line 62: As explained later, it would be nice to have a ElasticsearchMapping type, that would change this method to: 
```validateMapping: Json => IO[InvalidElasticSearchMapping, ElasticSearchMapping]```

**Review by imsdu — CHANGES_REQUESTED** [review]

File: delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ViewRef.scala [hunk 14]
```diff
@@ -0,0 +1,25 @@
...
+  * @param project the view parent project
+  * @param viewId  the view id
+  */
+final case class ViewRef(project: ProjectRef, viewId: Iri)
+
+object ViewRef {
+
+  // required for NonEmptySet
+  // sort by project first and then by view id
+  implicit final val viewRefOrder: Order[ViewRef] = Order.from { (fst, snd) =>
```
umbreak, line 19: More readable and simple:
```scala
implicit val viewRefOrder: Ordering[ViewRef] = 
   Ordering.by{case ViewRef(project, viewId) => (project,viewId)}
```

We will need an `Ordering` for `ProjectRef` and `Iri`, but that's fine, since it will be used in multiple places (ordering on `ProjectRef` might be used in resolvers too)

umbreak (changes requested):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
umbreak: Where `ElasticSearchIndexName` is just a placeholder for the index string:
```scala
final case class ElasticSearchIndexName private(value: String) extends AnyVal
// And here it will go the apply which will make sure it is safely created from a regex, as usually
```

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
bogdanromanx: I was thinking that we could allow AggregateElasticSearchViews in references and flatten their references. Happy to remove it if you don't find it useful.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
bogdanromanx: to validate a mapping we don't need a specific index name; we can always create an arbitrary index with this mapping to test it out

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
imsdu: If we introduce this, we also introduce the problem of circular references so we could let that for later no ?

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
umbreak: That's not true for updates. Certain updates to the mapping are incompatible with previous mappings.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
bogdanromanx: we only need to check direct references because they're most likely a mistake; for the rest it can be filtered out.

umbreak: It could potentially be useful (but I'm not aware of a current use case for it). It would also be a bit more complicated to implement. I don't have any strong opinion, so as you prefer.

In case you go for it, please add this to the documentation additions.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
bogdanromanx: I tried this locally, it doesn't work very well with the current code. We just need to make sure the value is validated; after that it doesn't make much use since we're not going to do anything else with the value.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
bogdanromanx: yeah, I'll remove it for now.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
bogdanromanx: Updates to the mappings will generate a completely new index.

umbreak: Yes that's right. Then forget about it.

umbreak (approved):

imsdu (approved):

PR #1835, opened 2020-11-25
Return metadata on write operations.

### LINKED ISSUE #1809: Provide consistent metadata responses for all write operations
Affected resources:
- realms
- acls (path)
- organizations
- projects
- resolvers (incoming and outgoing)
- schemas (incoming and outgoing)
- resources (incoming and outgoing)

wwajerowicz:
Fixes #1809


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala
```diff
@@ -54,7 +58,7 @@ final case class ResourceF[A](
   // format: off
   override def equals(obj: Any): Boolean =
     obj match {
-      case b @ ResourceF(_, _, `rev`, `types`, `deprecated`, `createdAt`, `createdBy`, `updatedAt`, `updatedBy`, `schema`, `value`) =>
+      case b @ ResourceF(_, _, `rev`, `types`, `deprecated`, `createdAt`, `createdBy`, `updatedAt`, `updatedBy`, `schema`, `value`, _, _) =>
```
umbreak, line 61: You can, and probably should, add the incoming and outgoing to the computation of the equals and hashcode.
The way to do it is pretty similar to the `id` and `accessUrl`

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -110,20 +110,20 @@ class AclsRoutes(identities: Identities, acls: Acls)(implicit
                   (put & entity(as[ReplaceAcl])) { case ReplaceAcl(AclValues(values)) =>
                     authorizeFor(address, aclsPermissions.write).apply {
                       val status = if (rev == 0L) Created else OK
-                      emit(status, acls.replace(Acl(address, values: _*), rev).map(_.void))
+                      emit(status, acls.replace(Acl(address, values: _*), rev).map(_.map(_.metadata)))
```
umbreak, line 113: This is quite annoying. We could have the following in some syntax:

```scala
trait IOResourceFSyntax {
  implicit final def resourceFSyntax[E, A](io:IO[E, ResourceF[A]]): IOResourceFOps = new IOResourceFOps(io)
}
final class IOResourceFOps[E, A](private val io:IO[E, ResourceF[A]]) extends AnyVal {
   def mapValue[B](f: A => B): IO[E, B] = io.map(_.map(f))
}
```

And then use `mapValue`

Something like that

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/realms/Realm.scala
```diff
@@ -65,4 +79,7 @@ object Realm {
   val context: ContextValue                             = ContextValue(contexts.realms)
   implicit val realmJsonLdEncoder: JsonLdEncoder[Realm] =
     JsonLdEncoder.fromCirce(context)
+
+  implicit val metadataEncoder: Encoder.AsObject[Metadata]    = deriveConfiguredEncoder[Metadata]
```
umbreak, line 83: I think this one should be private and only the `JsonLdEncoder[Metadata]` public

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Acl.scala
```diff
@@ -113,4 +128,13 @@ object Acl {
 
   implicit def aclJsonLdEncoder(implicit base: BaseUri): JsonLdEncoder[Acl] =
     JsonLdEncoder.fromCirce(context)
+
+  implicit private[Acl] val config: Configuration = Configuration.default.copy(transformMemberNames = {
+    case "path" => nxv.path.prefix
+    case other  => other
+  })
+
+  implicit val metadataEncoder: Encoder.AsObject[Metadata] = deriveConfiguredEncoder[Metadata]
```
umbreak, line 137: I think this one should be private and only the `JsonLdEncoder[Metadata]` public

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/acls/Acl.scala
```diff
@@ -113,4 +128,13 @@ object Acl {
 
   implicit def aclJsonLdEncoder(implicit base: BaseUri): JsonLdEncoder[Acl] =
     JsonLdEncoder.fromCirce(context)
+
+  implicit private[Acl] val config: Configuration = Configuration.default.copy(transformMemberNames = {
```
umbreak, line 132: shouldn't that be private?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/organizations/Organization.scala
```diff
@@ -35,4 +49,7 @@ object Organization {
   val context: ContextValue                                           = ContextValue(contexts.organizations)
   implicit val organizationJsonLdEncoder: JsonLdEncoder[Organization] =
     JsonLdEncoder.fromCirce(context)
+
+  implicit val metadataEncoder: Encoder.AsObject[Metadata]    = deriveConfiguredEncoder[Metadata]
```
umbreak, line 53: I think this one should be private and only the `JsonLdEncoder[Metadata]` public

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/Project.scala
```diff
@@ -57,4 +73,7 @@ object Project {
   implicit val projectJsonLdEncoder: JsonLdEncoder[Project] =
     JsonLdEncoder.fromCirce(context)
 
+  implicit val metadataEncoder: Encoder.AsObject[Metadata]    = deriveConfiguredEncoder[Metadata]
```
umbreak, line 76: I think this one should be private and only the `JsonLdEncoder[Metadata]` public

umbreak (changes requested):

umbreak (changes requested):
I see you call all the implicit values `metadataJsonLdEncoder`. I would rather have more specific names like `realmMetadataJsonLdEncoder` in case for some reasons you import both and then only one gets picked up because of collision names

bogdanromanx (changes requested):
The new keys need to be added to the context json files.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala
imsdu: I talked about it before but why not having something like this in ResourceF instead of the function:
```
sealed trait ResourceLink
final case class FixedLink(uri: Uri)
final case class CompactableLink(uri: Uri, mappings: ApiMappings, base: ProjectBase)
//describing other type of links
```
We would get rid of the functions where some have default values and therefore we wouldn't need to override the equals and hashCode.
And we would compute the different links in the encoder of ResourceLink and merge the resulting json with the rest of the resource fields

umbreak: I'm not sure if exactly that would work, since you need the `baseUri` to build those `uri` that you mention. The `ResourceLink` has to be built in order to build a `ResourceF`...so every time you build a resourceF you'll end up with the need of an `implicit baseUri: BaseUri`.

I don't see any simple way to do it without having the already mentioned drawback

imsdu: It would be paths more than uris.
And we would build the different links when encoding a little like what we do with subjects for the createdBy/updatedBy.

umbreak: With paths works. But now you have the logic scattered between `ResourceLink` and `AccessUrl`

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
umbreak: Or maybe a more generic one, with Functor:
```scala
trait IOFunctorSyntax {
  implicit final def ioFunctorOps[E, A: Functor](io:IO[E, A]): IOFunctorOps = new IOFunctorOps(io)
}
final class IOFunctorOps[E, A: Functor](private val io:IO[E, A]) {
   def mapValue[B](f: A => B): IO[E, B] = io.map(_.map(f))
}
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala
umbreak: Maybe something directly in `AccessUrl` or next to it called `PartialAccessUrl`, which interact with one another could work

imsdu: The scattered logic should be something we can solve

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/syntax/IOResourceFSyntax.scala
```diff
@@ -0,0 +1,18 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.syntax
+
+import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceF
+import monix.bio.IO
+
+trait IOResourceFSyntax {
```
umbreak, line 6: Since we have the opportunity let's do something more generic:
```scala
trait IOFunctorSyntax {
  implicit final def ioFunctorOps[E, A: Functor](io:IO[E, A]): IOFunctorOps = new IOFunctorOps(io)
}
final class IOFunctorOps[E, A: Functor](private val io:IO[E, A]) {
   def mapValue[B](f: A => B): IO[E, B] = io.map(_.map(f))
}
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/syntax/package.scala
```diff
@@ -14,3 +14,4 @@ package object syntax
     with IriUriSyntax
     with IterableSyntax
     with KamonSyntax
+    with IOResourceFSyntax
```
umbreak, line 17: This `IOResourceFSyntax ` has to be added as well to `ch.epfl.bluebrain.nexus.delta.syntax`

umbreak (changes requested):

umbreak (approved):

imsdu (approved):

PR #1837, opened 2020-11-26
Added endpoint to fetch tags

umbreak:


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/models/TagFields.scala
```diff
@@ -8,7 +8,7 @@ import io.circe.generic.extras.semiauto.deriveConfiguredDecoder
 import scala.annotation.nowarn
 
 /**
-  * The tag fields used as input
+  * The tag fields used as input/output on the routes
```
imsdu, line 11: As it is used both for input and output, we could just rename it Tag

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -152,22 +151,32 @@ final class ResourcesRoutes(
...
-                          authorizeFor(AclAddress.Project(ref), resourcePermissions.write).apply {
-                            entity(as[TagFields]) { case TagFields(tagRev, tag) =>
-                              operationName(s"$prefixSegment/resources/{org}/{project}/{schema}/{id}/tags") {
-                                emit(resources.tag(id, ref, schemaOpt, tag, tagRev, rev).map(_.void))
+                        (pathPrefix("tags") & pathEndOrSingleSlash) {
+                          operationName(s"$prefixSegment/resources/{org}/{project}/{schema}/{id}/tags") {
+                            concat(
+                              // Fetch a resource tags
+                              get {
+                                authorizeFor(AclAddress.Project(ref), resourcePermissions.read).apply {
```
imsdu, line 165: We can include the authorize and the rev, label part in the fetch method no ?

imsdu (changes requested):

bogdanromanx (approved):
this needs to be added to the api changes

umbreak:
This already existed in [1.4.x](https://github.com/BlueBrain/nexus/blob/v1.4.x/delta/src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/TagRoutes.scala#L69). I've tested it myself against staging. It just wasn't documented

Issue #997, opened 2020-01-24
Filtering SSE events


umbreak:
The more SSE are being used, the more it makes sense to request a subset of events. Those endpoints would be helpful for clients:
- {endpoint}/v1/files/{org}/{project}/events
- {endpoint}/v1/resolvers/{org}/{project}/events
- {endpoint}/v1/schemas/{org}/{project}/events
- {endpoint}/v1/views/{org}/{project}/events
- {endpoint}/v1/storages/{org}/{project}/events
- {endpoint}/v1/resources/{org}/{project}/{schema}/events

PR #1840, opened 2020-11-30
Resolvers routes

imsdu:
Routes for resolvers, wiring comes in the next PR


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutes.scala
```diff
@@ -0,0 +1,206 @@
...
+    s: Scheduler,
+    cr: RemoteContextResolution,
+    ordering: JsonKeyOrdering
+) extends AuthDirectives(identities, acls)
+    with CirceUnmarshalling {
+
+  import baseUri.prefixSegment
+  implicit val resolverContext: ContextValue = Resolvers.context
+
+  private val Read  = resolverPermissions.read
```
umbreak, line 50: You can do that just with an alias in the imports:

```scala
import ch.epfl.bluebrain.nexus.delta.sdk.Permissions.resolvers.{read => Read, write => Write}
```

umbreak (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutes.scala
```diff
@@ -0,0 +1,206 @@
...
+                operationName(s"$prefixSegment/resolvers/events") {
+                  authorizeFor(AclAddress.Root, events.read).apply {
+                    lastEventId { offset =>
+                      emit(resolvers.events(offset))
+                    }
+                  }
+                }
+              }
+            },
+            (projectRef | projectRefFromUuidsLookup(projects)) { ref =>
```
umbreak, line 84: Where are the SSEs on:
- /resolvers/{org}/events
- /resolvers/{org}/{proj}/events

umbreak (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutes.scala
```diff
@@ -0,0 +1,206 @@
...
+              val projectAddress = AclAddress.Project(ref)
+              val authorizeRead  = authorizeFor(projectAddress, Read)
+              val authorizeWrite = authorizeFor(projectAddress, Write)
+              concat(
+                // List resolvers
+                (get & pathEndOrSingleSlash & extractUri & paginated & resolverSearchParams(ref)) {
+                  (uri, pagination, params) =>
+                    operationName(s"$prefixSegment/resolvers/{org}/{project}") {
```
umbreak, line 92: you don't need to repeat this directive in both cases. Before the contact you could have:
```scala
(pathEndOrSingleSlash & operationName(s"$prefixSegment/resolvers/{org}/{project}")) {
   concat(...)
}
```

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -28,19 +31,19 @@ trait Resolvers {
   /**
     * Create a new resolver where the id is either present on the payload or self generated
     * @param projectRef        the project where the resolver will belong
-    * @param resolverFields the payload to create the resolver
+    * @param payload the payload to create the resolver
     */
-  def create(projectRef: ProjectRef, resolverFields: ResolverFields)(implicit
+  def create(projectRef: ProjectRef, payload: Json)(implicit
```
umbreak, line 36: I would recommend to have both methods. Someone using the resolvers SDK that is not going through the API should not have to provide a json payload in order to create a resolver.

usually the `create (projectRef, json)` should then call the `create(projectRef, resolverFields)`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
```diff
@@ -49,9 +52,9 @@ trait Resolvers {
     * @param id             the resolver identifier to expand as the id of the resolver
     * @param projectRef        the project where the resolver will belong
     * @param rev            the current revision of the resolver
-    * @param resolverFields the payload to update the resolver
+    * @param payload the payload to update the resolver
     */
-  def update(id: IdSegment, projectRef: ProjectRef, rev: Long, resolverFields: ResolverFields)(implicit
+  def update(id: IdSegment, projectRef: ProjectRef, rev: Long, payload: Json)(implicit
```
umbreak, line 57: same

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdSourceParser.scala
```diff
@@ -18,51 +19,58 @@ import monix.bio.{IO, UIO}
...
-  def expandIri[R](segment: IdSegment, project: Project)(implicit
-      rejectionMapper: Mapper[JsonLdRejection, R]
-  ): IO[R, Iri] =
-    IO.fromOption(
-      segment.toIri(project.apiMappings, project.base),
-      rejectionMapper.to(InvalidId(segment.asString))
-    )
+      (_, expanded) <- expandSource(project, source)
+      _             <- checkAndSetSameId(iri, expanded)
+      decodedValue  <- IO.fromEither(expanded.to[A].leftMap(DecodingFailed))
```
umbreak, line 71: you are using the `expanded` here, while you should be using the one resulting from the method `checkAndSetSameId `:
```scala
for {
   (_, originalExpanded) <- expandSource(project, source)
   expanded              <- checkAndSetSameId(iri, originalExpanded)
   decodedValue          <- IO.fromEither(expanded.to[A].leftMap(DecodingFailed))
} yield decodedValue
```

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
imsdu: Only the second create method would fit as we need an id.
And it is a little more complicated than that as we need to generate the json source from the value.
So I thought that as we don't need it for now, we could postpone their implementation

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutes.scala
imsdu: We keep it for 1.6.0 according to this: #997

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
umbreak: Yes, we can postpone it.

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutes.scala
umbreak: I'm fine with doing this later. However it is already done for resources and schemas. 

It seems quite trivial to do it for the rest of resources since we are currently building the routes. It is one of the situations where doing it now will be probably faster than later.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
bogdanromanx: I think turning a json-ld document to a ResolverFields should be a serialization concern and it should not show in the contract. I would recommend the following signature(s):

```
def create(project: ProjectRef, id: Iri, value: ResolverValue, source: Json): ???
def create(project: ProjectRef, value: ResolverValue, source: Json): ???
...
```

imsdu: The thing is we need  a project to deserialize it properly and I follow the same pattern than the one with resources/schemas.
If we do it for resolvers, we should do it for the others no ?

umbreak: I don't see why this contract is better than the existing one. Serialization **is** a concern of the SDK since we need ids and other checks.

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutes.scala
imsdu: I would like to start the most important part, resource resolution asap.
This is nice to have but we can go back to this later.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resolvers.scala
bogdanromanx: I'm thinking about the default resolver that needs to be created for a project; I think it's nicer to have it as a value with specific fields rather than a json blob.

imsdu: Oh yes, I forgot that one.
So I will add now another method for it which takes a ResolverValue and generates the json source from it.

bogdanromanx: And in the case of plugins, it's a lot simpler to deal with these value types. If we want to also expose the serialization part in the sdk, we can always add a new method and even possibly overload the others:

```
def valueFrom(source: Json): IO[ParsingRejection, (Option[Iri], ResolverValue)]
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverValue.scala
```diff
@@ -68,6 +71,36 @@ object ResolverValue {
     override def tpe: ResolverType = ResolverType.CrossProject
   }
 
+  /**
+    * Generate the value payload from an id and a value
+    * @param id             the id of the resolver
+    * @param resolverValue  the value to encode as Json
+    */
+  def generatePayload(id: Iri, resolverValue: ResolverValue): Json = {
```
umbreak, line 79: why do you need something like this? Don't you have already a way to convert your `ResolverValue` to JsonLd using the `JsonLdDecoder[ResolverValue]`. Seems to me that the same has to be used

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverValue.scala
umbreak: I see you don't have it for `ResolverValue` but for `Resolver`. Then convert your `ResolverValue` to `Resolver` and use the compacted for

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverValue.scala
imsdu: It is the Resolver (and not the ResolverValue) which is exposed in the api. But you are right, some refactorings can be made to make things better.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/Resolver.scala
```diff
@@ -49,6 +64,28 @@ object Resolver {
...
+  implicit val resolverEncoder: Encoder.AsObject[Resolver] = {
+    Encoder.AsObject.instance {
+      case InProjectResolver(_, _, priority, _, _)                                                 =>
+        JsonObject(
+          "priority" -> priority.asJson
+        )
+      case CrossProjectResolver(_, _, resourceTypes, projects, identityResolution, priority, _, _) =>
+        JsonObject(
+          "priority"      -> priority.asJson,
+          "resourceTypes" -> resourceTypes.asJson,
```
umbreak, line 82: If you want this field to be compacted, you can do the following:
```scala
// Declare this somewhere on the top or bottom of the companion object
private val ctx = JsonLdContext(ContextValue.empty, base = Some(nxv.base))

// then
"resourceTypes" -> resourceTypes.map(ctx.compact(_, useVocab = false)).asJson

```

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverValue.scala
umbreak: Since you are talking about the Json value, does not matter if it is `Resolver` or `ResolverValue`, since the way it is serialized to the client should be the same, as far as I'm aware.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/Resolver.scala
imsdu: The resolver has more fields including the source and tags so they are not the same thing and it is not possible to generate a Resolver only from a ResolverValue and an Iri

bogdanromanx (dismissed):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/Resolver.scala
umbreak: For that purpose it can be generated with empty tags and source, since for the encoding they are discarded. 

Alternatively you can refactor some of this stuff. But basically this `Encoder.AsObject[Resolver]` generates the same Json as the other one

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/Resolver.scala
```diff
@@ -49,39 +54,31 @@ object Resolver {
...
       source: Json,
       tags: Map[Label, Long]
-  ) extends Resolver
+  ) extends Resolver {
+    override def priority: Priority = value.priority
+  }
 
   val context: ContextValue = ContextValue(contexts.resolvers)
 
   implicit val resolverEncoder: Encoder.AsObject[Resolver] = {
```
umbreak, line 79: shouldn't it be `private` or `private[resolver]` ? Usually we want to expose only the `JsonLdEncoder[Resolver]` and not the `[Encoder.AsObject[Resolver]]`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverValue.scala
```diff
@@ -101,10 +85,17 @@ object ResolverValue {
       .addContext(contexts.resolvers)
   }
 
-  @nowarn("cat=unused")
-  implicit val resolverValueEncoder: Encoder.AsObject[ResolverValue] = {
-    implicit val config: Configuration = Configuration.default.withDiscriminator(keywords.tpe)
-    deriveConfiguredEncoder[ResolverValue]
+  implicit val resolverValueEncoder: Encoder.AsObject[ResolverValue] = Encoder.AsObject.instance {
```
umbreak, line 88: shouldn't it be `private` or `private[resolver]` ?

umbreak (changes requested):

umbreak:
For the rest it looks good to me

umbreak (dismissed):

umbreak (approved):

PR #1843, opened 2020-12-01
Added SchemaImports resolution

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1825

TODO: 
- Implement `Resolvers.fetchSchema` and `Resolvers.fetchResource` or alternative, TBD
- Add some test where some resource is validated against a schema with imports fetched using the resolvers

### DIFF
--- a/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/SchemasRoutesSpec.scala
+++ b/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/SchemasRoutesSpec.scala
@@ -10,7 +10,8 @@ import ch.epfl.bluebrain.nexus.delta.kernel.utils.UrlUtils
 import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.{contexts, nxv}
 import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.JsonLdContext.keywords
 import ch.epfl.bluebrain.nexus.delta.sdk.Permissions.{events, resources, schemas}
-import ch.epfl.bluebrain.nexus.delta.sdk.generators.{ProjectGen, SchemaGen}
+import ch.epfl.bluebrain.nexus.delta.sdk.SchemaImports
+import ch.epfl.bluebrain.nexus.delta.sdk.generators.ProjectGen
 import ch.epfl.bluebrain.nexus.delta.sdk.model.Label
 import ch.epfl.bluebrain.nexus.delta.sdk.model.acls.{Acl, AclAddress}
 import ch.epfl.bluebrain.nexus.delta.sdk.model.identities.Identity.{Anonymous, Authenticated, Group, Subject}
@@ -56,24 +57,23 @@ class SchemasRoutesSpec
   )
   private val projectRef = project.value.ref
 
-  private val myId         = nxv + "myid"
-  private val myIdEncoded  = UrlUtils.encode(myId.toString)
-  private val myId2        = nxv + "myid2"
-  private val myId2Encoded = UrlUtils.encode(myId2.toString)
-  val payload              = jsonContentOf("resources/schema.json") deepMerge json"""{"@id": "$myId"}"""
-  val payloadNoId          = payload.removeKeys(keywords.id)
-  val schema               = SchemaGen.schema(myId, projectRef, payload)
-  val current              = SchemaGen.currentState(schema, subject = subject)
-  val payloadUpdated       = payloadNoId.replace("datatype" -> "xsd:integer", "xsd:double")
-  val schemaUpdated        = SchemaGen.schema(myId, projectRef, payloadUpdated)
+  private val myId           = nxv + "myid"
+  private val myIdEncoded    = UrlUtils.encode(myId.toString)
+  private val myId2          = nxv + "myid2"
+  private val myId2Encoded   = UrlUtils.encode(myId2.toString)
+  private val payload        = jsonContentOf("resources/schema.json") deepMerge json"""{"@id": "$myId"}"""
+  private val payloadNoId    = payload.removeKeys(keywords.id)
+  private val payloadUpdated = payloadNoId.replace("datatype" -> "xsd:integer", "xsd:double")
 
   private val (orgs, projs) =
     ProjectSetup.init(orgsToCreate = List(org), projectsToCreate = List(project.value)).accepted
 
+  private val schemaImports = SchemaImports(ResolversDummy(projs).accepted)
+
   private val acls = AclsDummy(PermissionsDummy(Set(schemas.write, schemas.read, events.read))).accepted
 
   private val routes =
-    Route.seal(SchemasRoutes(identities, acls, orgs, projs, SchemasDummy(orgs, projs).accepted))
+    Route.seal(SchemasRoutes(identities, acls, orgs, projs, SchemasDummy(orgs, projs, schemaImports).accepted))
 
   "A schema route" should {
 
--- a/delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/Vocabulary.scala
+++ b/delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/Vocabulary.scala
@@ -156,6 +156,7 @@ object Vocabulary {
 
     // Resource types
     val AccessControlList = nxv + "AccessControlList"
+    val Ontology          = nxv + "Ontology"
     val Organization      = nxv + "Organization"
     val Permissions       = nxv + "Permissions"
     val Project           = nxv + "Project"
--- a/delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/Vocabulary.scala
+++ b/delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/Vocabulary.scala
@@ -156,7 +156,6 @@ object Vocabulary {
 
     // Resource types
     val AccessControlList = nxv + "AccessControlList"
-    val Ontology          = nxv + "Ontology"
     val Organization      = nxv + "Organization"
     val Permissions       = nxv + "Permissions"
     val Project           = nxv + "Project"
--- a/delta/rdf/src/test/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/ExpandedJsonLdSpec.scala
+++ b/delta/rdf/src/test/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/ExpandedJsonLdSpec.scala
@@ -13,6 +13,7 @@ import scala.collection.immutable.VectorMap
 class ExpandedJsonLdSpec extends AnyWordSpecLike with Matchers with Fixtures {
 
   "An expanded Json-LD" should {
+    val example          = "http://example.com"
     val compacted        = jsonContentOf("compacted.json")
     val context          = compacted.topContextValueOrEmpty
     val expectedExpanded = jsonContentOf("expanded.json")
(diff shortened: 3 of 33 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverResolutionRejection.scala
```diff
@@ -0,0 +1,68 @@
...
+    * Rejection returned when attempting to fetch a resource using a resolver that doesn't exist.
+    *
+    * @param id           the resource identifier
+    * @param project      the project it belongs to
+    * @param resourceType the resource type
+    */
+  final case class ResourceNotFound(id: Iri, project: ProjectRef, resourceType: ResourceType)
+      extends ResolverResolutionRejection(s"$resourceType '$id' not found in project '$project'.")
+
+  final case class ProjectNotFound(projectRef: ProjectRef)
```
bogdanromanx, line 52: docs

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Schemas.scala
```diff
@@ -194,22 +195,25 @@ object Schemas {
...
 
-    @SuppressWarnings(Array("OptionGet"))
     // It is fine to do it unsafely since we have already computed the graph on evaluation previously in order to validate the schema.
-    def toGraphUnsafe(expanded: ExpandedJsonLd) =
-      expanded.toGraph.toOption.get
+    def toSchemaGraph(expanded: ExpandedJsonLd) =
+      expanded.filterType(nxv.Schema).toGraph.toOption.get
+
+    def toOntologyGraph(expanded: ExpandedJsonLd) =
+      expanded.filterType(nxv.Ontology).toGraph.toOption.get
```
bogdanromanx, line 206: as far as I know, we have schemas that also have the type Ontology so this strategy of splitting based on type will not work as it's implemented right now.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceType.scala
```diff
@@ -0,0 +1,40 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.model
+
+/**
+  * The enumeration of the platform resource types available
```
bogdanromanx, line 4: platform => system

delta/rdf/src/test/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/ExpandedJsonLdSpec.scala
```diff
@@ -142,20 +150,36 @@ class ExpandedJsonLdSpec extends AnyWordSpecLike with Matchers with Fixtures {
...
+      val multiRoot = jsonContentOf("/jsonld/expanded/input-multiple-roots.json")
+      val batmanIri = iri"$example/batman"
+      val name      = iri"$example/name"
+      val json      =
+        json"""[{"@id": "$batmanIri", "@type": ["$example/Person", "$example/Hero"], "$name": [{"@value": "Batman"}] }]"""
+
+      ExpandedJsonLd(multiRoot).accepted.filterType(iri"$example/Hero") shouldEqual
+        ExpandedJsonLd.expanded(json).rightValue
+
+      ExpandedJsonLd(multiRoot).accepted.filterType(iri"$example/Other") shouldEqual
```
bogdanromanx, line 181: I would also add a filterType test with `iri"$example/Person"` to check that it echoes back the entire document.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/SchemaImports.scala
```diff
@@ -0,0 +1,85 @@
...
+        rejectionRefs                          = schemaRejections ++ resourceRejections -- successRefs
+        _                                     <- if (rejectionRefs.nonEmpty) IO.raiseError(InvalidSchemaResolution(id, rejectionRefs)) else IO.unit
+      } yield (successRefs, schemaSuccess.values.map(_.expanded) ++ resourceSuccess.values.map(_.expanded))
+
+    def recurse(
+        resolved: Set[ResourceRef],
+        document: ExpandedJsonLd
+    ): IO[SchemaRejection, (Set[ResourceRef], ExpandedJsonLd)] = {
+      val imports   = document.cursor.downField(owl.imports).get[Set[ResourceRef]]
+      val toResolve = imports.getOrElse(Set.empty) -- resolved
```
bogdanromanx, line 49: if there's a decoding error (where the value cannot be decoded as a ResourceRef) it would not be reported; can that happen?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/SchemaImports.scala
```diff
@@ -0,0 +1,85 @@
...
+    */
+  def resolve(id: Iri, projectRef: ProjectRef, expanded: ExpandedJsonLd): IO[SchemaRejection, ExpandedJsonLd] = {
+
+    def fetchSchema(fetch: ResourceRef): IO[SchemaRejection, Schema]     = self.fetchSchema(projectRef, fetch)
+    def fetchResource(fetch: ResourceRef): IO[SchemaRejection, Resource] = self.fetchResource(projectRef, fetch)
+
+    def lookupFromSchemasAndResources(toResolve: Set[ResourceRef]) =
+      for {
+        (schemaRejections, schemaSuccess)     <- lookupInBatch(toResolve, fetchSchema)
+        resourcesToResolve                     = toResolve -- schemaSuccess.keySet
```
bogdanromanx, line 37: we should probably reject results that don't have the type Ontology.

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/Vocabulary.scala
```diff
@@ -156,6 +156,7 @@ object Vocabulary {
 
     // Resource types
     val AccessControlList = nxv + "AccessControlList"
+    val Ontology          = nxv + "Ontology"
```
bogdanromanx, line 159: this is wrong... it should be in the `owl` namespace with a different prefix.

bogdanromanx (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/SchemaImports.scala
umbreak: Since the @context is in place, telling that `owl:imports` must have @id values, I don't think this can happen here

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Schemas.scala
umbreak: I didn't find anything like that in neuroshapes. However, I've made the change

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/SchemasRoutesSpec.scala
```diff
@@ -56,24 +57,23 @@ class SchemasRoutesSpec
...
+  private val myId2          = nxv + "myid2"
+  private val myId2Encoded   = UrlUtils.encode(myId2.toString)
+  private val payload        = jsonContentOf("resources/schema.json") deepMerge json"""{"@id": "$myId"}"""
+  private val payloadNoId    = payload.removeKeys(keywords.id)
+  private val payloadUpdated = payloadNoId.replace("datatype" -> "xsd:integer", "xsd:double")
 
   private val (orgs, projs) =
     ProjectSetup.init(orgsToCreate = List(org), projectsToCreate = List(project.value)).accepted
 
+  private val schemaImports = SchemaImports(ResolversDummy(projs).accepted)
```
imsdu, line 71: I guess this one is ok but we will need to take care of circular dependencies

umbreak: Yep, we'll have to address this in the future.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverResolutionRejection.scala
```diff
@@ -0,0 +1,73 @@
...
+object ResolverResolutionRejection {
+
+  /**
+    * Rejection returned when a subject intends to retrieve a resource using a resolver at a specific revision,
+    * but the provided revision does not exist.
+    *
+    * @param provided the provided revision
+    * @param current  the last known revision
+    */
+  final case class RevisionNotFound(provided: Long, current: Long)
```
imsdu, line 29: Unless we want a report of what the failed resolutions when resolving a resource, we don't need this because we can fail first but another resource can have this id on a different project and may have this revision no ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverResolutionRejection.scala
```diff
@@ -0,0 +1,73 @@
...
+        s"Revision requested '$provided' not found, last known revision is '$current'."
+      )
+
+  /**
+    * Rejection returned when a subject intends to retrieve a resource using a resolver at a specific tag,
+    * but the provided tag does not exist.
+    *
+    * @param tag the provided tag
+    */
+  final case class TagNotFound(tag: Label) extends ResolverResolutionRejection(s"Tag requested '$tag' not found.")
```
imsdu, line 40: Same

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverResolutionRejection.scala
umbreak: Imagine you have just one resolver in a project, the default in project resolver (which is the normal case).
In this case you try to fetch a resource using that resolver and 4 things can happen (the same as the current logic):
- The revision provided is wrong
- The tag provided is wrong
- The project does not exist
- The resource id does not exist

If you have multiple resolvers, you will still have to save the last error and issue it back to the client

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/SchemaImports.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+    * Resolve the ''imports'' from the passed ''expanded'' document and recursively from the resolved documents.
+    *
+    * @param id         the schema id
+    * @param projectRef the project where the schema belongs to
+    * @param expanded   the schema expanded form
+    * @return a "fat-schema" with all the imports resolved
+    */
+  def resolve(id: Iri, projectRef: ProjectRef, expanded: ExpandedJsonLd): IO[SchemaRejection, ExpandedJsonLd] = {
```
imsdu, line 29: We also need the current caller here

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/SchemaImports.scala
umbreak: These details should be added once `Resolvers.fetchSchema` and `Resolvers.fetchResource` or an alternative is implemented.

It does not have much to do with `SchemaImport` lookup resolution. Let's add the necessary bits and pieces afterwards

bogdanromanx (approved):

imsdu (approved):

PR #1845, opened 2020-12-02
JsonLdEncoder fixes

umbreak:
Added methods:
- **JsonLdEncoder.computeFromCirce**: Computes the compacted/expanded form from the provided Json derived from Circe `Encoder`
- **JsonLdEncoder.compcatedFromCirce**: Uses the provided Json derived from the Circe `Encoder` as the compacted form

Most of the times (except for SSE events, which is an edge case where we want the source intact) we actually want the Json-LD algorithm to kick in, since that will ensure our contexts are correct.

Fixes:
- Removed `nxv` CURIE from all contexts (curies are dangerous since they'll compact things that we might not want).
- Fixed certain contexts
- Removed `_compacted` and `expanded` and other unnecessary fields from SSE


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/marshalling/RdfRejectionHandler.scala
```diff
@@ -323,10 +323,10 @@ object RdfRejectionHandler {
     HttpResponseFields(_ => StatusCodes.BadRequest)
 
   implicit private[routes] def compactFromCirceRejection[A <: Rejection: Encoder.AsObject]: JsonLdEncoder[A] =
-    JsonLdEncoder.fromCirce(id = bnode, iriContext = contexts.error)
+    JsonLdEncoder.computeFromCirce(id = bnode, ctx = ContextValue(contexts.error))
```
imsdu, line 326: We could have an overloaded computeFromCirce method that would take care of introducing the context value

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/context/ContextValue.scala
```diff
@@ -24,12 +24,12 @@ final case class ContextValue private[jsonld] (value: Json) {
     */
   def merge(that: ContextValue): ContextValue =
     (value.asArray, that.value.asArray, value.asString, that.value.asString) match {
-      case (Some(arr), Some(thatArr), _, _) => arrOrObj(removeEmpty(arr ++ thatArr))
-      case (_, Some(thatArr), _, _)         => arrOrObj(removeEmpty(value +: thatArr))
-      case (Some(arr), _, _, _)             => arrOrObj(removeEmpty(arr :+ that.value))
-      case (_, _, Some(str), Some(thatStr)) => arrOrObj(removeEmpty(Seq(str.asJson, thatStr.asJson)))
-      case (_, _, Some(str), _)             => arrOrObj(removeEmpty(Seq(str.asJson, that.value)))
-      case (_, _, _, Some(thatStr))         => arrOrObj(removeEmpty(Seq(value, thatStr.asJson)))
+      case (Some(arr), Some(thatArr), _, _) => arrOrObj(removeEmptyAndDup(arr ++ thatArr))
```
imsdu, line 27: Making the ContextValue with an ADT with:
```
case object EmptyContext
case class   ContextList 
```
would make it easier no ?
We only really need json on the contextObj method, for the rest, it seems bring to be bothersome as json is something a little large

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/marshalling/RdfRejectionHandler.scala
umbreak: There were too many overloads, I found it more annoying than actually helping

delta/sdk/src/main/resources/contexts/metadata.json
```diff
@@ -17,11 +17,12 @@
       "@type": "@id"
     },
     "_instant": "https://bluebrain.github.io/nexus/vocabulary/instant",
-    "_subject": "https://bluebrain.github.io/nexus/vocabulary/subject",
+    "_subject": "https://bluebrain.github.io/nexus/vocabulary/metadata/subject",
```
bogdanromanx, line 20: any reason for this change?

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/context/ContextValue.scala
umbreak: I would not touch this part of the logic now, since it is sensitive. A `ContextValue` can be (from the point of view of Json)
- String
- Empty
- Object
- List

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/permissions/PermissionsEventSpec.scala
```diff
@@ -60,7 +60,7 @@ class PermissionsEventSpec extends AnyWordSpecLike with Matchers with TestHelper
               "@value" : 1
             }
           ],
-          "https://bluebrain.github.io/nexus/vocabulary/subject" : [
+          "https://bluebrain.github.io/nexus/vocabulary/metadata/subject" : [
```
bogdanromanx, line 63: same as above

delta/sdk/src/main/resources/contexts/metadata.json
umbreak: Yes, because of collisions with the `subject` on ACLs and resolvers and others.

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/permissions/PermissionsEventSpec.scala
umbreak: See above

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/marshalling/RdfRejectionHandler.scala
imsdu: The contexts defined in the Vocabulary could be a ContextValue if we never use them as a Iri

delta/sdk/src/main/resources/contexts/metadata.json
bogdanromanx: I see. How did we handle this in <= v1.4?

umbreak: It was wrongly handled :)

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/marshalling/RdfRejectionHandler.scala
umbreak: They are used as [iris too](https://github.com/BlueBrain/nexus/blob/master/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala#L43).

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/context/ContextValue.scala
imsdu: Yep, I understand this is sensitive and could be done later, no worries.
The String and Object cases can be expressed as the List case to simplify things no ?
That is why I had only two cases in the ADT

umbreak: But then you'll need a list with another ADT on it with 3 cases.

bogdanromanx (approved):

imsdu (approved):

Issue #1846, opened 2020-12-02
SSE events missing fields


umbreak:
1.4.x events looks as follows:
```json
{
  "@context": "https://bluebrain.github.io/nexus/contexts/resource.json",
  "@type": "Updated",
  "_resourceId": "https://staging.nexus.ocp.bbp.epfl.ch/v1/resources/bbp_test/studio_data2/_/CampaignReportCollectionNexusTask_entity_managemen_https___staging__RasterTask_6338922bf6_activity",
  "_source": {
    "@context": [
      "https://bbp.neuroshapes.org",
      "https://bluebrain.github.io/nexus/contexts/resource.json"
    ],
    "@type": "CampaignAnalysis",
    "endedAtTime": "2020-03-18T11:05:15.660971"
  },
  "_types": [
    "https://neuroshapes.org/CampaignAnalysis"
  ],
  "_projectUuid": "c909bd38-6b50-4981-862d-4beab46376b1",
  "_organizationUuid": "7798cc81-a818-4ea4-aa85-cf5361ebcac4",
  "_rev": 5,
  "_instant": "2020-03-18T11:05:15.944046Z",
  "_subject": "https://staging.nexus.ocp.bbp.epfl.ch/v1/realms/bbp/users/ivaska"
}
```

We are missing the `_types` fields and also the `_projectUuid` and `_organizationUuid`.

umbreak:
We should also unify the `@id` and `_resourceId`. All events should have a field called:
- permissions: `_permissionId`
- acls: `_aclId`
- realms: `_realmId`
- organizations: `_organizationId`
- projects: `_projectId`
- resources: `_resourceId`
- resolvers: `_resolverId`
- schemas: `_schemaId`
...

PR #1848, opened 2020-12-02
Removed functions from ResourceF

umbreak:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceUris.scala
```diff
@@ -0,0 +1,175 @@
...
+    */
+  def outgoingShortForm(implicit base: BaseUri): Option[Uri]
+
+}
+
+object ResourceUris {
+
+  final private case class WithNavigation private (relativeAccessUri: Uri, relativeAccessUriShortForm: Uri)
+      extends ResourceUris {
+    override def incoming(implicit base: BaseUri): Option[Uri]          = Some(accessUri / "incoming")
```
imsdu, line 63: We only use these links during ResourceF serialization no ?
We could have an encoder taking a BaseUri and generate the JsonObject with the applicable links for each one.
A bit like we do, for subjects in ResourceF

imsdu (changes requested):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewState.scala
```diff
@@ -102,11 +100,11 @@ object ElasticSearchViewState {
...
     }
 
-    override def toResource(mappings: ApiMappings, base: ProjectBase): Option[ElasticSearchViewResource] =
+    override def toResource(mappings: ApiMappings, base: ProjectBase): Option[ElasticSearchViewResource] = {
       Some(
         ResourceF(
-          id = accessUrl.andThen(_.iri),
-          accessUrl = accessUrl.andThen(_.shortForm(mappings, base)),
+          id = id,
+          uris = ResourceUris(s"views/$project", id)(mappings, base),
```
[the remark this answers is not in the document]
umbreak: I don't understand

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala
```diff
@@ -105,21 +63,19 @@ object ResourceF {
 
   implicit final private def resourceFUnitEncoder(implicit base: BaseUri): Encoder.AsObject[ResourceF[Unit]] =
     Encoder.AsObject.instance { r =>
-      val incoming =
-        r.incoming.apply(base).map(inc => JsonObject("_incoming" -> inc.asJson)).getOrElse(JsonObject.empty)
-      val outgoing =
-        r.outgoing.apply(base).map(out => JsonObject("_outgoing" -> out.asJson)).getOrElse(JsonObject.empty)
+      val incoming = r.uris.incomingShortForm.fold(JsonObject.empty)(in => JsonObject("_incoming" -> in.asJson))
```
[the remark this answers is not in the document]
umbreak: Why to make it more complicated? this is very explicit already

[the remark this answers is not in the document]
imsdu: To have the same approach as the one we have with createdBy/updatedBy and the identity decoder and have definition and serialization in a single file

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceUris.scala
umbreak: `ResourceUris` has nothing to do with serialization. As the class says
> A collection of resource uris

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/utils/RouteFixtures.scala
```diff
@@ -224,49 +211,42 @@ trait RouteFixtures extends TestHelpers {
...
       rev: Long,
       deprecated: Boolean,
       createdBy: Subject,
       updatedBy: Subject,
-      am: ApiMappings = ApiMappings.empty,
-      base: Iri = nxv.base,
       additionalMetadata: Json = Json.obj()
-  ): Json =
+  ): Json = {
+    val incoming = resourceUris.incomingShortForm.fold(Json.obj())(in => Json.obj("_incoming" -> in.asJson))
```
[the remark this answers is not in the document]
umbreak: All the routes are already tested in `ResourceUrisSpec`

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewState.scala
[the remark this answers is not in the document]
imsdu: Concatenation of the prefix and the projectRef could be defined in an apply method in ResourceUris more than in a method in a State class.

delta/app/src/main/resources/kamon.conf
```diff
@@ -62,17 +62,6 @@ kamon {
   }
 
   modules {
-
-    host-metrics {
```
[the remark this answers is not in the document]
umbreak: Yes. I introduced them to fix some issue with Mac Os 11, but @bogdanromanx fixed it already

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala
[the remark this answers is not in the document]
umbreak: Identities serialization is used all over the place, that's why it is provided in `Identity`. This is a small enough class with a small enough responsibility.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceUris.scala
umbreak: incoming and outgoing links are also used and stored in the elasticsearch plugin

delta/app/src/test/resources/permissions/eventstream-0-5.txt
```diff
@@ -1,20 +1,20 @@
-data:{"@context":["https://bluebrain.github.io/nexus/contexts/metadata.json","https://bluebrain.github.io/nexus/contexts/permissions.json"],"@type":"PermissionsAppended","_instant":"1970-01-01T00:00:00Z","_rev":1,"_subject":"http://localhost/v1/anonymous","permissions":["acls/read"]}
+data:{"@context":["https://bluebrain.github.io/nexus/contexts/metadata.json","https://bluebrain.github.io/nexus/contexts/permissions.json"],"@id":"http://localhost/v1/permissions","@type":"PermissionsAppended","_instant":"1970-01-01T00:00:00Z","_rev":1,"_subject":"http://localhost/v1/anonymous","permissions":["acls/read"]}
```
bogdanromanx, line 1: we've discussed this a long time ago, events should not have an `@id` field:

- the id is not that useful as there's no way to retrieve an event by id
- the id of the resource should not be the id of the event

In the case of "kg" resources we've used the `_resourceId` field to encode the corresponding resource. We should keep that if we want to specify the corresponding resource id.

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/IriOrBNode.scala
```diff
@@ -176,6 +176,18 @@ object IriOrBNode {
     override val asIri: Option[Iri] = Some(this)
 
     override val asBNode: Option[BNode] = None
+
+    /**
+      * Returns a new absolute Iri resolving the current relative [[Iri]] with the passed absolute [[Iri]]
+      */
+    def resolvedAgainst(iri: Iri): Iri =
```
bogdanromanx, line 183: this seems to fail silently if the provided iri is not absolute; I'm wondering if it would make sense to have a `RelativeIri` wrapper with a safe constructor to avoid these situations.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceUris.scala
```diff
@@ -0,0 +1,182 @@
...
+  /**
+    * @return the outgoing [[Uri]] in a short form
+    */
+  def outgoingShortForm(implicit base: BaseUri): Option[Uri]
+
+}
+
+object ResourceUris {
+
+  final private case class WithNavigation private (relativeAccessUri: Uri, relativeAccessUriShortForm: Uri)
```
bogdanromanx, line 61: there's no need for a private ctor if the class itself is private

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceUris.scala
```diff
@@ -0,0 +1,182 @@
...
+    val relativeShortForm = relative / ctx.compact(id, useVocab = false)
+    WithNavigation(relative / id.toString, relativeShortForm)
+  }
+
+  /**
+    * Constructs [[ResourceUris]] from a relative [[Uri]].
+    *
+    * @param relative the relative base [[Uri]]
+    */
+  private def apply(relative: Uri): ResourceUris =
```
bogdanromanx, line 100: I would recommend keeping this method available for use.

bogdanromanx (changes requested):

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/IriOrBNode.scala
umbreak: It is the same logic than the akka http `Uri`. It does not fail silently, there is nothing to resolve if the iri is already resolved. That's the desired behaviour.

delta/app/src/test/resources/permissions/eventstream-0-5.txt
umbreak: SSEs are anyway messed up in several other ways. I'm not sure if this is in the scope of the current PR. For some events we had `@id` (realms, acls, orgs, projects), for some we had `_resourceId` (resources, schemas, resolvers) and for some we had nothing (permissions).

I think this should be tacked in the jointly with the issue: https://github.com/BlueBrain/nexus/issues/1846

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/IriOrBNode.scala
bogdanromanx: and are we happy with the possibility of ending up with a relative iri?

umbreak: The tests will tell you :) In our code that won't happen cause we are always using the `BaseUri` for this. I don't feel like introducing a `AbsoluteIri` thing because that change is going to cascade all over the place now

bogdanromanx (approved):

imsdu (approved):

PR #1849, opened 2020-12-03
Initialized storage plugin

umbreak:


imsdu (approved):

bogdanromanx (approved):

bogdanromanx (changes requested):
Requires also including it in the app universal mapping (see elasticsearch example).

PR #1850, opened 2020-12-04
Added storage Command, Event, State and STM logic

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1495


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/Algorithm.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import cats.syntax.all._
+
+import java.security.MessageDigest
+import scala.util.Try
+
+/**
+  * A digest algorithm
+  */
+final case class Algorithm private (value: String) extends AnyVal
```
imsdu, line 11: DigestAlgoritm to make it clear ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/Storage.scala
```diff
@@ -0,0 +1,87 @@
...
+      source: Json
+  ) extends Storage
+
+  /**
+    * A storage that stores and fetches files from an S3 compatible service
+    */
+  final case class S3Storage(
+      id: Iri,
+      project: ProjectRef,
+      default: Boolean,
```
imsdu, line 59: We could have the StorageValue here no ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageRejection.scala
```diff
@@ -0,0 +1,180 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.{ProjectRef, ProjectRejection}
+import io.circe.syntax._
+import io.circe.{Encoder, JsonObject}
+
+/**
+  * Enumeration of Storage rejection types.
+  *
+  * @param reason a descriptive message as to why the rejection occurred
+  */
+sealed abstract class StorageRejection(val reason: String) extends Product with Serializable
```
imsdu, line 24: There is a default property for storage so we should have a rejection when we try to create a second default one no ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageValue.scala
```diff
@@ -0,0 +1,87 @@
...
+    * @param writePermission the permission required in order to upload a file to this storage
+    * @param maxFileSize     the maximum allowed file size (in bytes) for uploaded files
+    */
+  final case class DiskStorageValue(
+      default: Boolean,
+      algorithm: Algorithm,
+      volume: Path,
+      readPermission: Permission,
+      writePermission: Permission,
+      maxFileSize: Long
```
imsdu, line 35: We could have a refined type to make sure it is positive and also to put an upper-value that is reasonable

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageValue.scala
```diff
@@ -0,0 +1,87 @@
...
+    * @param readPermission  the permission required in order to download a file from this storage
+    * @param writePermission the permission required in order to upload a file to this storage
+    * @param maxFileSize     the maximum allowed file size (in bytes) for uploaded files
+    */
+  final case class RemoteDiskStorageValue(
+      default: Boolean,
+      algorithm: Algorithm,
+      endpoint: Uri,
+      credentials: Option[String],
+      folder: String,
```
imsdu, line 80: Is it not a path ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/Storages.scala
```diff
@@ -0,0 +1,110 @@
...
+
+    event match {
+      case e: StorageCreated    => created(e)
+      case e: StorageUpdated    => updated(e)
+      case e: StorageTagAdded   => tagAdded(e)
+      case e: StorageDeprecated => deprecated(e)
+    }
+  }
+
+  private[storage] def evaluate(
```
imsdu, line 57: When we already have a default storage for a project, shouldn't we fail ?

imsdu (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageValue.scala
umbreak: No, it is a single value. It could probably be a `Label`

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/Storages.scala
umbreak: This logic, as for when we have already a resolver with the same priority in a project, cannot be handled in the state machine

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageValue.scala
umbreak: The upper value is driven by configuration. So since we have to check dynamically for the upper value, I would rather check for both than having a custom case class for this.

imsdu: So this check could be part of the STM logic no ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/Storages.scala
imsdu: I haven't done it with the priority thing with resolvers because we said it could be done later.
For realms for exemple, we check for duplicates in the STM.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/Storage.scala
umbreak: I did the same way than the elastic search plugin. For now the `Storage` and the `StorageValue` are decoupled.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/Storages.scala
umbreak: It does not apply anymore. As we decided, this will be done in the `Storages` logic synchronously.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageRejection.scala
umbreak: There is no rejection for this, as we discussed on slack

imsdu (approved):

PR #1854, opened 2020-12-07
Added files Command, Event, State and STM logic

umbreak:


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/file/Files.scala
```diff
@@ -0,0 +1,119 @@
...
+    event match {
+      case e: FileCreated                   => created(e)
+      case e: FileUpdated                   => updated(e)
+      case e: FileComputedAttributesUpdated => updatedAttributes(e)
+      case e: FileTagAdded                  => tagAdded(e)
+      case e: FileDeprecated                => deprecated(e)
+    }
+  }
+
+  private[file] def evaluate(
```
imsdu, line 62: Where do we check the right to write on the storage / the size of the file ?
It should be here no ?

imsdu (changes requested):

bogdanromanx (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/file/model/FileCommand.scala
```diff
@@ -0,0 +1,116 @@
...
+    *
+    * @param id         the file identifier
+    * @param project    the project the file belongs to
+    * @param storage    the reference to the remote storage used
+    * @param mediaType  the media type of the file
+    * @param bytes      the size of the file file in bytes
+    * @param digest     the digest information of the file
+    * @param subject    the identity associated to this command
+    */
+  final case class UpdateFileComputedAttributes(
```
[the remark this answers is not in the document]
umbreak: I changed the name to make more concrete. This does not update any random `FileAttributes`, just the computed ones, as the docs explain

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/file/model/FileEvent.scala
```diff
@@ -0,0 +1,131 @@
...
+    * @param project   the project the file belongs to
+    * @param storage   the reference to the remote storage used
+    * @param mediaType the media type of the file
+    * @param bytes     the size of the file file in bytes
+    * @param digest    the digest information of the file
+    * @param rev       the last known revision of the file
+    * @param instant   the instant this event was created
+    * @param subject   the identity associated to this event
+    */
+  final case class FileComputedAttributesUpdated(
```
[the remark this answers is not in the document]
umbreak: same

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/file/Files.scala
umbreak: For files I'm afraid the routes are going to have this logic. Most likely in the routes we need to fetch the storage because we need it for the directive `withSizeLimit(storage.maxFileSize)`. And since we have the storage, we also check the permissions in the routes:
```scala
withSizeLimit(storage.maxFileSize)  & authorizeFor(projectPath, storage.writePermission) {
   fileUpload("file") { case (metadata, byteSource) =>
      // call the files bundle here
   }
}
```

If you guys have other ideas about how to do this in the `Files` stm without a big port of akka http functionality from those directives, I'm willing to do it inside `Files` stm

umbreak: If it is a stream, you might not always know the size of the file having the `HttpEntity`, since it can be chuncked. The size in that case is computed. So yeah I don't see an easy and clean way to do it inside `Files`. it seems much cleaner to do it in the routes in this case, as shown in the example.

imsdu: It seems to be done automatically by throwing an exception when you read the stream:
https://github.com/akka/akka-http/blob/master/akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java#L102

umbreak: Yes, but while [consuming the stream](https://github.com/akka/akka-http/blob/04ccca8d5e622f13f1711fc3f08a3581a6073293/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala#L631), which is when calling `fileUpload`

bogdanromanx (changes requested):
the deprecation status of the storage is still not addressed for creates and updates

umbreak:
This is going to be addressed in the routes as well:
```scala
parameter("storage".as[IdSegment].?) { storageIdOpt =>
	storage(storageIdOpt) { storage => // This directive checks the storage and that the storage is not deprecated
		withSizeLimit(storage.maxFileSize)  & authorizeFor(projectPath, storage.writePermission) {
		   fileUpload("file") { case (metadata, byteSource) =>
		      // call the files bundle here
		   }
		}
	}
}
```

The deprecation of the storage should be checked **before** we consume the payload, otherwise is a waste of time.

bogdanromanx (approved):

PR #1870, opened 2020-12-09
Added storage implementation

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1526

TODO (in a separate PR): encrypt/decrypt credentials


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/config/AggregateConfig.scala
```diff
@@ -0,0 +1,18 @@
...
+import ch.epfl.bluebrain.nexus.sourcing.processor.{EventSourceProcessorConfig, StopStrategyConfig}
+
+/**
+  * Aggregate configuration.
+  *
+  * @param stopStrategy     the stop strategy configuration for this aggregate [[EventDefinition]]
+  * @param snapshotStrategy the snapshot strategy configuration for this aggregate [[EventDefinition]]
+  * @param processor        the event source processor config
+  */
+//TODO: ported from service module, we might want to avoid this duplication
```
imsdu: I think it will be better do it soon

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragesConfig.scala
```diff
@@ -0,0 +1,116 @@
...
+    * @param digestAlgorithm algorithm for checksum calculation
+    * @param readPermission  the default permission required in order to download a file from a disk storage
+    * @param writePermission the default permission required in order to upload a file to a disk storage
+    * @param showLocation    flag to decide whether or not to show the absolute location of the files in the metadata response
+    * @param maxFileSize     the default maximum allowed file size (in bytes) for uploaded files
+    */
+  final case class DiskStorageConfig(
+      volume: Path,
+      digestAlgorithm: DigestAlgorithm,
+      readPermission: Permission,
```
imsdu: the last 4 ones could be included in a common subsetting as they are common to every storage.
And if we do the same thing with the equivalent values in StorageValue, it should make it easy combine them where needed

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageState.scala
```diff
@@ -81,17 +85,60 @@ object StorageState {
...
-        case RemoteDiskStorageValue(default, algorithm, endpoint, cred, folder, read, write, maxSize) =>
-          RemoteDiskStorage(id, project, default, algorithm, endpoint, cred, folder, read, write, maxSize, tags, source)
+        case DiskStorageValue(default, volume, read, write, maxSize)                                       =>
+          DiskStorage(
+            id = id,
+            project = project,
+            default = default,
+            algorithm = cfg.disk.digestAlgorithm,
+            volume = volume,
+            readPermission = read.getOrElse(cfg.disk.readPermission),
```
imsdu: It could be more interesting to pass the config to the STM in the next function so as to have it in the state instead of overwriting it here ?

imsdu (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageState.scala
```diff
@@ -81,17 +85,60 @@ object StorageState {
...
+          )
+        case S3StorageValue(default, bucket, endpoint, accessKey, secretKey, region, read, write, maxSize) =>
+          S3Storage(
+            id = id,
+            project = project,
+            default = default,
+            algorithm = cfg.amazonUnsafe.digestAlgorithm,
+            bucket = bucket,
+            endpoint = endpoint.orElse(cfg.amazonUnsafe.defaultEndpoint),
+            accessKey = accessKey.orElse(if (endpoint.forall(endpoint.contains)) cfg.amazonUnsafe.accessKey else None),
```
imsdu: If someone removes the aws config because it is not relevant at some point, we should be able to fetch the state even if it is deprecated

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/Storages.scala
```diff
@@ -1,25 +1,503 @@
...
+
+  private def expandIri(segment: IdSegment, project: Project): IO[InvalidStorageId, Iri] =
+    JsonLdSourceParser.expandIri(segment, project, InvalidStorageId.apply)
+}
 
 object Storages {
 
+  private[storage] type StoragesAggregate =
+    Aggregate[String, StorageState, StorageCommand, StorageEvent, StorageRejection]
+  private[storage] type StoragesCache     = KeyValueStore[(ProjectRef, Iri), StorageResource]
```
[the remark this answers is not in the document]
umbreak: I changed this in last commit

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragesConfig.scala
umbreak: It is easier for deriving the pureconfig like this, and I don't have any other reason why to group it

umbreak: It is easier for deriving the pureconfig like this, and I don't have any other reason why to group it

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/config/AggregateConfig.scala
umbreak: Sure. We talked about it with bogdan this morning in the meeting. We'll do it when we find a common pattern to goup things

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageState.scala
umbreak: I don't know what you mean

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/Storages.scala
```diff
@@ -55,22 +533,59 @@ object Storages {
...
+        config.amazon.as(StorageType.S3Storage) ++
+        config.remoteDisk.as(StorageType.RemoteDiskStorage)
+
     def validate(id: Iri, value: StorageValue) =
-      access(value) >> validateFileSize(id, value.maxFileSize)
+      validateStorageType(id, value.tpe) >> validatePermissions(value) >> access(value) >> validateFileSize(id, value)
 
-    def validateFileSize(id: Iri, size: Long) =
-      if (size <= 0 || size > maxFileSize) IO.raiseError(InvalidMaxFileSize(id, size, maxFileSize)) else IO.unit
+    def validatePermissions(value: StorageValue) =
```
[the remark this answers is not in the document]
umbreak: I don't know what you mean

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageState.scala
imsdu: 1) Someone configure delta with s3, create storages relying on default values
2) S3 is not used anymore, the s3 config is removed
3) the amazonUnsafe will fail

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageState.scala
umbreak: I tried this but in the there was a lot of extra code involved, since I had to pass the whole `storage: Storage`

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/Storages.scala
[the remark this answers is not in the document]
imsdu: Something like:
```
val storagePermissionSet = readPermission.toSet ++ writePermission
if(storagePermissionSet.subsetOf(allPermissions))
  success
else
  fail
```

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageState.scala
imsdu: We can't have something like this in the STM ?
```
def next(storageConfig: StorageConfig)(state: StorageState, event: StorageEvent)
```
And maybe have a special state for the case of RemoteStorage/S3 storage which has been enabled and then disabled ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/model/StorageState.scala
umbreak: I understand your concern. I addressed that by having yet another  case class: `StorageFields`.
1. StorageFields: used as client input, to serialize json-ld into a case class and in `StorageCommand`
2. StorageValue: used for `StorageEvent`. It contains all the information from the `StorageFields` plus the resolved configuration
3. Storage: It contains `StorageValue` plus some extra fields like source, tags, id and project

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/serialization/KryoSerializerInit.scala
```diff
@@ -0,0 +1,33 @@
...
+import ch.epfl.bluebrain.nexus.delta.plugins.storages.serialization.KryoSerializerInit.PathSerializer
+import com.esotericsoftware.kryo.io.{Input, Output}
+import com.esotericsoftware.kryo.{Kryo, Serializer}
+import io.altoo.akka.serialization.kryo.DefaultKryoInitializer
+import io.altoo.akka.serialization.kryo.serializer.scala.ScalaKryo
+
+import java.nio.file.Path
+
+//TODO: partially ported from service module, we might want to avoid this duplication
+class KryoSerializerInit extends DefaultKryoInitializer {
```
bogdanromanx, line 12: I think this will collide with the service one; the serializer has a single initialisation class.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/Storages.scala
```diff
@@ -0,0 +1,632 @@
...
+    projects: Projects
+)(implicit rcr: RemoteContextResolution, uuidF: UUIDF) {
+
+  /**
+    * Create a new storage where the id is either present on the payload or self generated
+    *
+    * @param projectRef the project where the storage will belong
+    * @param payload    the payload to create the storage
+    */
+  def create(projectRef: ProjectRef, payload: Json)(implicit caller: Subject): IO[StorageRejection, StorageResource] = {
```
bogdanromanx, line 63: can we refer to the "payload" as "source" instead?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/model/Storage.scala
```diff
@@ -0,0 +1,93 @@
...
+      value: RemoteDiskStorageValue,
+      tags: Map[Label, Long],
+      source: Json
+  ) extends Storage {
+    override val default: Boolean = value.default
+  }
+
+  val context: ContextValue = ContextValue(contexts.storage)
+
+  implicit private val storageEncoder: Encoder[Storage] = Encoder.instance(_.source)
```
bogdanromanx, line 90: if the source tolerates optional fields the source will not be an accurate encoding result

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/model/StorageEvent.scala
```diff
@@ -1,4 +1,4 @@
-package ch.epfl.bluebrain.nexus.delta.plugins.storage.model
+package ch.epfl.bluebrain.nexus.delta.plugins.storages.storage.model
```
bogdanromanx, line 1: I would recommend a different package structure, where the default package is (notice the singular):
```
ch.epfl.bluebrain.nexus.delta.plugins.storage
```
... and the model contains the file/storage sub-packages such that it follows the SDK structure:
```
ch.epfl.bluebrain.nexus.delta.plugins.storage.model.files
ch.epfl.bluebrain.nexus.delta.plugins.storage.model.storages
```

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/model/StorageValue.scala
```diff
@@ -0,0 +1,78 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.storages.storage.model
+
+import akka.http.scaladsl.model.Uri
+import ch.epfl.bluebrain.nexus.delta.sdk.model.Label
+import ch.epfl.bluebrain.nexus.delta.sdk.model.identities.AuthToken
+import ch.epfl.bluebrain.nexus.delta.sdk.model.permissions.Permission
+
+import java.nio.file.Path
+
+sealed trait StorageValue extends Product with Serializable {
```
bogdanromanx, line 10: other common fields could be extracted here, like algorithm etc.

delta/plugins/storage/src/test/resources/akka-test.conf
```diff
@@ -0,0 +1,47 @@
...
+      "ch.epfl.bluebrain.nexus.delta.plugins.storages.file.model.FileEvent"           = "circe"
+      "java.nio.file.Path"                                                            = "kryo"
+      "ch.epfl.bluebrain.nexus.sourcing.projections.StreamSupervisor$Stop$"           = "kryo"
+    }
+    testkit.typed.throw-on-shutdown-timeout = false
+  }
+
+  cluster {
+    min-nr-of-members = 1
+    sharding.state-store-mode = ddata
```
bogdanromanx, line 37: add `jmx.multi-mbeans-in-same-jvm = on` to avoid noise on the tests

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/StoragesSpec.scala
```diff
@@ -0,0 +1,661 @@
...
+
+    val storages = (for {
+      eventLog         <- EventLog.postgresEventLog[Envelope[StorageEvent]](EventLogUtils.toEnvelope).hideErrors
+      (orgs, projects) <- projectSetup
+      storages         <- Storages(storageConfig, eventLog, perms, orgs, projects, access)
+    } yield storages).accepted
+
+    "creating a storage" should {
+
+      "succeed with the id present on the payload" in {
```
bogdanromanx, line 392: maybe a test to for changing the default?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/syntax/UriSyntax.scala
```diff
@@ -61,16 +61,19 @@ final class UriOps(private val uri: Uri) extends AnyVal {
...
-      uri.copy(path = uri.path + segment.drop(1))
-    else if (uriEndsWithSlash)
-      uri.copy(path = uri.path + segment)
-    else if (segmentStartsWithSlash)
-      uri.copy(path = uri.path / segment.drop(1))
-    else
-      uri.copy(path = uri.path / segment)
+    if (segment.trim.isEmpty) uri
+    else {
+      lazy val segmentStartsWithSlash = segment.startsWith("/")
```
bogdanromanx, line 66: I don't think the lazy is necessary for these two values as the synchronization overload could actually make it slower.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/StoragesConfig.scala
```diff
@@ -0,0 +1,116 @@
...
+    * @param readPermission        the default permission required in order to download a file from a remote disk storage
+    * @param writePermission       the default permission required in order to upload a file to a remote disk storage
+    * @param showLocation          flag to decide whether or not to show the absolute location of the files in the metadata response
+    * @param maxFileSize           the default maximum allowed file size (in bytes) for uploaded files
+    */
+  final case class RemoteDiskStorageConfig(
+      defaultEndpoint: Uri,
+      defaultEndpointPrefix: String,
+      defaultCredentials: Option[AuthToken],
+      digestAlgorithm: DigestAlgorithm,
```
bogdanromanx, line 108: this is provided by the remote storage, so do we need to configure it?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/StoragesConfig.scala
```diff
@@ -0,0 +1,116 @@
...
+      readPermission: Permission,
```
bogdanromanx, line 109: should these other fields be prefixed with "default" as well?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/StoragesConfig.scala
```diff
@@ -0,0 +1,116 @@
...
+    * @param amazon        configuration for the s3 compatible storage
+    * @param remoteDisk    configuration for the remote disk storage
+    */
+  @SuppressWarnings(Array("OptionGet"))
+  final case class StorageTypeConfig(
+      disk: DiskStorageConfig,
+      amazon: Option[S3StorageConfig],
+      remoteDisk: Option[RemoteDiskStorageConfig]
+  ) {
+    lazy val amazonUnsafe: S3StorageConfig             = amazon.get
```
bogdanromanx, line 46: this is used in the code, so it can throw an error... if that is the case, we should maybe use default values or handle the error downstream

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/Storages.scala
```diff
@@ -0,0 +1,632 @@
...
+      initialState = Initial,
+      next = next,
+      evaluate = evaluate(access, permissions, config.storageTypeConfig),
+      tagger = (event: StorageEvent) =>
+        Set(
+          moduleType,
+          s"${Projects.moduleType}=${event.project}",
+          s"${Organizations.moduleType}=${event.project.organization}"
+        ),
+      snapshotStrategy = config.aggregate.snapshotStrategy.combinedStrategy(
```
bogdanromanx, line 487: I don't think there's a need to snapshot storages; the expected number of events is quite low.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/Storages.scala
```diff
@@ -0,0 +1,632 @@
...
+  )(implicit
+      uuidF: UUIDF,
+      clock: Clock[UIO],
+      scheduler: Scheduler,
+      as: ActorSystem[Nothing],
+      rcr: RemoteContextResolution
+  ): UIO[Storages] =
+    for {
+      agg     <- aggregate(config, access, permissions)
+      index    = cache(config)
```
bogdanromanx, line 428: this should be also delayed

bogdanromanx (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/model/StorageValue.scala
umbreak: When needed. So far not needed.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/serialization/KryoSerializerInit.scala
umbreak: Yes, and it is not a problem. Maybe I can move it to the tests, since this is just used in order to pass the tests (the path kryo serialization is already covered in the service)

bogdanromanx: it would be better to move to the tests then, yes

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/StoragesSpec.scala
umbreak: This is already on the tests. The SSEs test verify that, the test on the update verifies that (because the revision gets affected), and others

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storages/storage/StoragesConfig.scala
umbreak: If you see the state machine, it is used in a way that it can't throw.
Besides that, the name says it is unsafe, which tells you not to use it unless you know what you're doing.

bogdanromanx (approved):

imsdu (approved):

PR #1871, opened 2020-12-09
Make cache durable and loaded at startup from disk

### LINKED ISSUE #1806: Ensure cache consistency when the service boots up for realm, resolvers and storages
Realms have an unicity constraint on the openIdConfig uri that cannot be enforced during the service bootstrap.
Resolvers have an unicity constraint on the priority that cannot be enforced during the service bootstrap.

Replay the current events by tag for these resources when the service starts to populate the cache and bind the routes only after this operation is complete.

imsdu:
Fixes #1806

### DIFF
--- a/delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resolvers/ResolversImpl.scala
+++ b/delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resolvers/ResolversImpl.scala
@@ -25,7 +25,7 @@ import ch.epfl.bluebrain.nexus.delta.sdk.model.search.SearchResults.UnscoredSear
 import ch.epfl.bluebrain.nexus.delta.sdk.model.{Envelope, IdSegment, Label}
 import ch.epfl.bluebrain.nexus.delta.sdk.utils.UUIDF
 import ch.epfl.bluebrain.nexus.delta.service.config.AggregateConfig
-import ch.epfl.bluebrain.nexus.delta.service.resolvers.ResolversImpl.{ResolversAggregate, ResolversCache}
+import ch.epfl.bluebrain.nexus.delta.service.resolvers.ResolversImpl.{ResolverKey, ResolversAggregate, ResolversCache}
 import ch.epfl.bluebrain.nexus.delta.service.syntax._
 import ch.epfl.bluebrain.nexus.sourcing._
 import ch.epfl.bluebrain.nexus.sourcing.processor.EventSourceProcessor.persistenceId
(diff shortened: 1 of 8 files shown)


delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resolvers/ResolversImpl.scala
```diff
@@ -180,8 +180,10 @@ final class ResolversImpl(
 
 object ResolversImpl {
 
+  final private[resolvers] case class ResolverKey(project: ProjectRef, iri: Iri)
```
umbreak, line 183: This seems to be unrelated to the current PR, isn't it?

I would suggest any improvement to the current code to be made in a separate PR. If you want to do this you should do it too for Schemas and Resources

umbreak (changes requested):

tests/src/test/resources/application.conf
```diff
@@ -1,6 +1,6 @@
 tests {
-    delta-uri = "http://localhost:8080/v1"
-    realm-uri = "http://"${"keycloak:8080"}"/auth/realms"
```
bogdanromanx, line 3: this is required if you want to do local debugging

bogdanromanx (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resolvers/ResolversImpl.scala
imsdu: I did it because I had to declare the cache key in the serialization bindings

tests/src/test/resources/application.conf
imsdu: Resolver creation was failing because of this as it wasn't able to fetch the openId config

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resolvers/ResolversImpl.scala
bogdanromanx: this is required to be able to serialize the cache key

umbreak (approved):

tests/src/test/resources/application.conf
bogdanromanx: that's very strange; the docker compose plugin sets the `keycloak:8080` env variable.

imsdu: Yep, but we use the realm-uri for realm payload for the config url so the `keycloak:8080` env variable resolves to localhost:xxxx which can't be reached by the delta instance as it is the address exposed outside the docker compose network. Inside it is `keycloak:8080`

bogdanromanx (approved):

PR #1883, opened 2020-12-10
Added encryption of sensitive storage fields

umbreak:
This includes the following:
- accessKey and secretKey in `S3Storage`
- credentials in `RemoteDiskStorage`

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
@@ -7,6 +7,7 @@ import cats.syntax.all._
 import ch.epfl.bluebrain.nexus.delta.kernel.RetryStrategy
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.Storages._
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.StoragesConfig.StorageTypeConfig
+import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.EncryptionState.{Decrypted, Encrypted}
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.StorageCommand.{CreateStorage, DeprecateStorage, TagStorage, UpdateStorage}
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.StorageEvent.{StorageCreated, StorageDeprecated, StorageTagAdded, StorageUpdated}
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.StorageRejection._
@@ -51,7 +52,8 @@ final class Storages private (
     eventLog: EventLog[Envelope[StorageEvent]],
     cache: StoragesCache,
     orgs: Organizations,
-    projects: Projects
+    projects: Projects,
+    crypto: Crypto
 )(implicit rcr: RemoteContextResolution, uuidF: UUIDF) {
 
   /**
@@ -374,7 +376,7 @@ final class Storages private (
 
   private def stateAt(project: ProjectRef, iri: Iri, rev: Long) =
     EventLogUtils
-      .fetchStateAt(eventLog, persistenceId(moduleType, identifier(project, iri)), rev, Initial, next)
+      .fetchStateAt(eventLog, persistenceId(moduleType, identifier(project, iri)), rev, Initial, next(crypto))
       .leftMap(RevisionNotFound(rev, _))
 
   private def identifier(project: ProjectRef, id: Iri): String =
@@ -389,7 +391,7 @@ object Storages {
   private[storage] type StoragesAggregate =
     Aggregate[String, StorageState, StorageCommand, StorageEvent, StorageRejection]
   private[storage] type StoragesCache     = KeyValueStore[StorageKey, StorageResource]
-  private[storage] type StorageAccess     = StorageValue => IO[StorageNotAccessible, Unit]
+  private[storage] type StorageAccess     = StorageValue[Decrypted] => IO[StorageNotAccessible, Unit]
   final private[storage] case class StorageKey(project: ProjectRef, iri: Iri)
 
   /**
@@ -426,7 +428,7 @@ object Storages {
     for {
       agg     <- aggregate(config, access, permissions)
       index   <- UIO.delay(cache(config))
-      storages = apply(agg, eventLog, index, orgs, projects)
+      storages = apply(agg, eventLog, index, orgs, projects, config.storageTypeConfig.encryption.crypto)
       _       <- UIO.delay(startIndexing(config, eventLog, index, storages))
     } yield storages
 
@@ -435,9 +437,10 @@ object Storages {
       eventLog: EventLog[Envelope[StorageEvent]],
       index: StoragesCache,
       orgs: Organizations,
-      projects: Projects
+      projects: Projects,
+      crypto: Crypto
   )(implicit rcr: RemoteContextResolution, uuidF: UUIDF) =
-    new Storages(agg, eventLog, index, orgs, projects)
+    new Storages(agg, eventLog, index, orgs, projects, crypto)
 
   private def cache(config: StoragesConfig)(implicit as: ActorSystem[Nothing]): StoragesCache = {
     implicit val cfg: KeyValueStoreConfig      = config.keyValueStore
@@ -476,7 +479,7 @@ object Storages {
     val definition = PersistentEventDefinition(
       entityType = moduleType,
       initialState = Initial,
-      next = next,
+      next = next(config.storageTypeConfig.encryption.crypto)(_, _),
       evaluate = evaluate(access, permissions, config.storageTypeConfig),
       tagger = (event: StorageEvent) =>
         Set(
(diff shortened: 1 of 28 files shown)


imsdu (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageValue.scala
```diff
@@ -1,20 +1,24 @@
...
 import io.circe.generic.extras.Configuration
 import io.circe.generic.extras.semiauto.deriveConfiguredEncoder
 import io.circe.syntax._
+import io.circe.Encoder
 
 import java.nio.file.Path
 import scala.annotation.nowarn
 
-sealed trait StorageValue extends Product with Serializable {
+sealed trait StorageValue[A <: EncryptionState] extends Product with Serializable {
```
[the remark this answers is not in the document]
umbreak: I tried going back and forth between the two.

However the other approach generates too much code and repetition while not having an obvious advantage. 
The current way works well since `Storage` is force to have a `StorageValue[Decrypted]` and `StorageEvent` is forced to have a `StorageValue[Encrypted]`.

In terms of serialization, each serializer should take care of what it wants to do with the values. In this case `StorageEvent` serializer takes care of its serialization and `Storage` takes care of its own.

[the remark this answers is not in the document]
umbreak: The code and the tests makes sure that having a `StorageValue[Decrypted]` already provides the correctly decrypted fields, so there is no reason to have a concrete type for those fields.

Same goes for `StorageValue[Encrypted]`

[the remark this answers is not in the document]
imsdu: The fact that secrets are encoded as String is a problem because they can be displayed on the logs by our code but also akka code and others and we should do wrap them anyway.
If we just want encryption at rest, it should not generate too much code as it will be handled by the encoder/decoder.
Where would it be verbose ?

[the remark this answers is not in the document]
umbreak: I don't understand the problem with the logs. The concern with the logs or akka displaying credentials can be easily fixed by overriding the `toString`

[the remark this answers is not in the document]
imsdu: Absolutely but my point is that we should do it with StorageFields, StorageConfig where this can be solved by wrapping the sensitive strings.
We could also use this wrapper to properly encode/decode the token/secrets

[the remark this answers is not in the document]
umbreak: I get it

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
```diff
@@ -544,13 +573,15 @@ object Storages {
...
+    def validateAndReturnValue(id: Iri, fields: StorageFields): IO[StorageRejection, StorageValue[Encrypted]] =
       for {
-        value <- IO.fromOption(fields.toValue(config), InvalidStorageType(id, fields.tpe, allowedStorageTypes))
-        _     <- validatePermissions(fields)
-        _     <- access(value)
-        _     <- validateFileSize(id, fields.maxFileSize, value.maxFileSize)
-      } yield value
+        value     <- IO.fromOption(fields.toValue(config), InvalidStorageType(id, fields.tpe, allowedStorageTypes))
+        encrypted <- IO.fromEither(value.encrypt(config.encryption.crypto))
+        _         <- IO.fromEither(encrypted.decrypt(config.encryption.crypto))
```
bogdanromanx, line 580: why are we attempting to decrypt it? is it to ensure it can be decrypted?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/EncryptionState.scala
```diff
@@ -0,0 +1,31 @@
...
+  /**
+    * An decrypted state
+    */
+  final case object Decrypted extends EncryptionState
+
+  type Encrypted = Encrypted.type
+  type Decrypted = Decrypted.type
+
+  implicit val encryptedStateEncoder: Encoder[Encrypted] = Encoder.instance(_ => Json.Null)
+  implicit val encryptedStateDecoder: Decoder[Encrypted] = Decoder.decodeOption[Unit].as(Encrypted)
```
bogdanromanx, line 27: use `Decoder.const(Encrypted)`

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageRejection.scala
```diff
@@ -203,6 +214,11 @@ object StorageRejection {
       val tpe = ClassUtils.simpleName(r)
       val obj = JsonObject(keywords.tpe -> tpe.asJson, "reason" -> r.reason.asJson)
       r match {
+        case InvalidEncryptionSecrets(tpe, details)  =>
+          logger.warn(
```
bogdanromanx, line 218: I think this warrants an Error level

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageValue.scala
bogdanromanx, line 19: I would recommend a doc that explains the need to the generic type arg

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageValue.scala
```diff
@@ -65,35 +99,97 @@ object StorageValue {
...
+  implicit private[model] val storageValueEncoder: Encoder[StorageValue[Decrypted]] = {
     implicit val config: Configuration      = Configuration.default.withDiscriminator(keywords.tpe)
     implicit val pathEncoder: Encoder[Path] = Encoder.encodeString.contramap(_.toString)
 
     Encoder.encodeJsonObject.contramapObject { storage =>
-      deriveConfiguredEncoder[StorageValue].encodeObject(storage).add(keywords.tpe, storage.tpe.iri.asJson)
+      deriveConfiguredEncoder[StorageValue[Decrypted]]
+        .encodeObject(storage)
+        .add(keywords.tpe, storage.tpe.iri.asJson)
+        .remove("credentials")
```
bogdanromanx, line 190: maybe make these fields constants to avoid typos

bogdanromanx (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
umbreak: Yes, since the next function will decrypt it unsafely, we have to check in the eval function if that can be done

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/EncryptionState.scala
umbreak: No, it does not work for null values, I've already tried

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageValue.scala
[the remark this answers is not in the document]
umbreak: I've fixed that adding a `Secret` case class and used it everywhere where there could be sensitive data involved (fields, config, source)

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageValue.scala
umbreak: I've removed most of the usages with the introduction of the `Secret` class

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
```diff
@@ -496,19 +509,46 @@ object Storages {
...
   }
 
-  private[storage] def next(
+  private def encryptSource(source: DecryptedSecret[Json], value: StorageValue[Encrypted]): EncryptedSecret[Json] =
+    source.mapEncrypt(changeSourceSensitiveValues(value, _))
+
+  private def decryptSource(source: EncryptedSecret[Json], value: StorageValue[Decrypted]): DecryptedSecret[Json] =
+    source.mapDecrypt(changeSourceSensitiveValues(value, _))
+
+  private def changeSourceSensitiveValues[A <: EncryptionState](value: StorageValue[A], source: Json): Json =
```
imsdu, line 518: We could just encrypt the whole source payload to make this simpler and not changing the incoming payload

bogdanromanx (approved):

imsdu (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
umbreak: We don't want this, because it will make it less readable for debugging purposes. You still want to be able to read easily the rest of the fields

imsdu: We have StorageValue for debugging purposes

umbreak: But I want to debug what the client posted, not what is in the value (then I don't know if the client posted it or not)

umbreak: Besides, in 99% of the cases no client is going to provide credentials or accessKey/secretKey through the API. 
Instead they are going to rely on the configuration values, which makes the encryption of the whole json an overkill.

PR #1885, opened 2020-12-11
Add resolver based context resolution

### LINKED ISSUE #1857: Expand resources / schemas / resolvers using the resolution logic for fetching context references

imsdu:
Fixes #1857


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdSourceParser.scala
```diff
@@ -40,12 +40,13 @@ trait JsonLdSourceParser {
...
     */
   def decode[A: JsonLdDecoder, R](project: Project, source: Json)(implicit
+      context: ContextValue,
       uuidF: UUIDF,
       rcr: RemoteContextResolution,
       rejectionMapper: Mapper[JsonLdRejection, R]
   ): IO[R, (Iri, A)] = {
     for {
-      (_, expanded) <- expandSource(project, source)
+      (_, expanded) <- expandSource(project, source.addContext(context))
```
umbreak, line 49: You removed this line from the `Impl` just to add it here. I don't find it beneficial in terms of readability since this is done through an implicit `ContextValue` that won't get picked up automatically but it will have to be imported in the implementation itself.

I find it more obfuscating than instructing

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverContextResolution.scala
```diff
@@ -0,0 +1,71 @@
...
+    (iri: Iri) =>
+      rcr
+        .resolve(iri)
+        .onErrorFallbackTo(
+          resolveResource(ResourceRef(iri), projectRef, caller)
+            .bimap(
+              report =>
+                RemoteContextNotAccessible(
+                  iri,
+                  s"Resolution via static resolution and via resolvers failed in $projectRef",
```
umbreak, line 35: can you wrap this variable in single quotes? `'$projectRef'`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverContextResolution.scala
```diff
@@ -0,0 +1,71 @@
...
+                  Some(report.asJson)
+                ),
+              result => result.compacted.ctx.contextObj.asJson
```
umbreak, line 38: I think you should use the source, instead of the compacted (more efficient)

```scala
result.source.topContextValueOrEmpty.contextObj.asJson
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdSourceParser.scala
imsdu: We already add context to the SearchResults encoder, via an implicit value, I thought it was similar and good enough.

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdSourceParser.scala
```diff
@@ -61,11 +62,12 @@ trait JsonLdSourceParser {
     * @return a tuple with the resulting @id iri and the decode value
     */
   def decode[A: JsonLdDecoder, R](project: Project, iri: Iri, source: Json)(implicit
```
umbreak, line 64: if you pass here as implicit the `ResolverResolution`, you don't need to have the extra line

```scala
implicit val rcr: RemoteContextResolution = contextResolution(projectRef)
```
in every implementation. You can just do it directly here, since you already have the project too

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdSourceParser.scala
umbreak: It isn't great for `SearchResults` too. And that happens in the routes and not on the Impl

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resolvers/ResolverContextResolution.scala
imsdu: Ok, I thought that was one of the pros of having the compacted resource.

umbreak:
I think `JsonLdSourceParser ` is growing quite a lot with implicits. Why don't we just make it a class and in its constructors it can have all those fields as parameters:

```scala
final class JsonLdSourceParser[R](implicit rcr: RemoteContextResolution, rejectionMapper: Mapper[JsonLdRejection, R], uuidF: UUIDF, contextResolution: ResolverContextResolution) { 
   // rest simplified
}
```

Then on our implementations we just have this parser as a dependency, and it will be simpler to use

bogdanromanx (commented):
looks good to me

umbreak (approved):
I would try the refactoring I mentioned in the previous comment on `JsonLdSourceParser `, since this object has grown enough to be a class :)

However I'm happy if that is done in a subsequent PR

bogdanromanx (approved):

PR #1893, opened 2020-12-16
Moved HttpClient to sdk module

umbreak:
Some other minor refactorings too


delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/marshalling/CirceUnmarshallingSpec.scala
```diff
@@ -1,12 +1,12 @@
 package ch.epfl.bluebrain.nexus.delta.routes.marshalling
```
imsdu, line 1: Move the test to sdk

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/marshalling/RdfMarshallingSpec.scala
```diff
@@ -9,7 +9,7 @@ import ch.epfl.bluebrain.nexus.delta.SimpleResource.{context, contextIri}
 import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.nxv
 import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution
 import ch.epfl.bluebrain.nexus.delta.rdf.utils.JsonKeyOrdering
-import ch.epfl.bluebrain.nexus.delta.routes.marshalling.RdfMediaTypes._
+import ch.epfl.bluebrain.nexus.delta.rdf.RdfMediaTypes._
```
imsdu, line 12: Maybe this one too

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClient.scala
```diff
@@ -0,0 +1,96 @@
...
+  * Http client based on the akka http model.
+  */
+trait HttpClient {
+
+  /**
+    * Execute the argument request and unmarshal the response into an A using a [[Decoder]].
+    *
+    * @param req the request to execute
+    */
+  def to[A: Decoder](req: HttpRequest)(implicit A: ClassTag[A]): IO[HttpClientError, A] =
```
bogdanromanx, line 27: I think we should include the base method here to make it more usable:

```
def execute(req: HttpRequest): Resource[IO[HttpClientError, *], HttpResponse]
```

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/marshalling/RdfMarshallingSpec.scala
umbreak: that's for jsonLd stuff. For now this is still defined in the app.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClient.scala
bogdanromanx: and if we change the return types to have `Resource[IO[HttpClientError, *], A]` instead, we can get rid of the discard bytes, since the bytes can automatically be discarded after handling the response.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClient.scala
```diff
@@ -0,0 +1,96 @@
...
+import monix.bio.{IO, Task}
+import monix.execution.Scheduler
+
+import scala.concurrent.TimeoutException
+import scala.reflect.ClassTag
+
+/**
+  * Http client based on the akka http model.
+  */
+trait HttpClient {
```
bogdanromanx, line 20: you probably need to disable coverage measurement for this class since there aren't any direct tests for it.

bogdanromanx (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClient.scala
umbreak: Let's do this later, when there is a need for it

bogdanromanx (approved):

PR #1894, opened 2020-12-17
Add resolver subresource endpoints

### LINKED ISSUE #1858: Create resolver sub-resource endpoints
```
/resolvers/{org}/{proj}/{resolver_id}/{resource_to_resolve_id}
/resolvers/{org}/{proj}/_/{resource_to_resolve_id}
/resolvers/{org}/{proj}/{resolver_id}/{resource_to_resolve_id}?tag={tag}
/resolvers/{org}/{proj}/_/{resource_to_resolve_id}?tag={tag}
/resolvers/{org}/{proj}/{resolver_id}/{resource_to_resolve_id}?rev={rev}
/resolvers/{org}/{proj}/_/{resource_to_resolve_id}?rev={rev}
```

imsdu:
Fixes #1858


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutes.scala
```diff
@@ -158,6 +168,9 @@ final class ResolversRoutes(identities: Identities, acls: Acls, projects: Projec
                           }
                         )
                       }
+                    },
+                    iri { resourceIri =>
```
umbreak, line 172: why is here an iri? It should be an `idSegment` since it is the same as the other API operations related with ids

umbreak (changes requested):

umbreak (changes requested):
I find it a bit confusing in some signatures during resolve resolution that sometimes we have `IdSegment` and sometimes `Iri`. 

I guess it make sense, but I cannot really follow up why in some cases we have one and in others another, unless I would dig very deep in the code.

bogdanromanx (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutes.scala
imsdu: It can't be an IdSegment because you don't have the mappings and the base to expand it properly as an Iri.

delta/app/src/test/resources/resolvers/schema-resolved-resolver-resolution-report.json
```diff
@@ -0,0 +1,20 @@
+{
+  "@context": "https://bluebrain.github.io/nexus/contexts/resolvers.json",
+  "Resource": {
```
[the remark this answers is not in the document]
imsdu: For rejections, this part is embedded in a reports object, should I do the same ?
Or change both to something else ?

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutes.scala
umbreak: Why not? you use the mappings and the base of the current project

imsdu: The resource may be in another project

umbreak: yes, but the API is centered on the original project, from the API perspective that's what matters

umbreak (approved):

bogdanromanx (approved):

PR #1906, opened 2020-12-18
Added support for RemoteDiskStorage file upload/download

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1863


delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/remote/RemoteDiskStorageAccessSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{BaseUri, Label}
+import ch.epfl.bluebrain.nexus.delta.sdk.syntax._
+import ch.epfl.bluebrain.nexus.testkit.{IOValues, TestHelpers}
+import monix.execution.Scheduler
+import org.scalatest.DoNotDiscover
+import org.scalatest.concurrent.Eventually
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+@DoNotDiscover
```
imsdu, line 18: To delete no ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClient.scala
```diff
@@ -47,52 +46,56 @@ trait HttpClient {
 
 object HttpClient {
 
+  type AkkaSource = Source[ByteString, Any]
```
imsdu, line 49: This is not only used in the HttpClient

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/remote/RemoteDiskStorageAccess.scala
```diff
@@ -1,14 +1,24 @@
...
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.StorageValue.RemoteDiskStorageValue
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.StorageAccess
+import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.remote.client.RemoteDiskStorageClient
 import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.sdk.model.identities.AuthToken
 import monix.bio.IO
+import monix.execution.Scheduler
 
-object RemoteDiskStorageAccess extends StorageAccess {
+class RemoteDiskStorageAccess(implicit as: ActorSystem, sc: Scheduler) extends StorageAccess {
```
imsdu, line 14: We could create only one instance of this client and pass it along to the access, fetch and save operations so that they don't need to know about the ActorSystem or the Scheduler which are only related to the client.

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/remote/client/RemoteStorageClientSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{BaseUri, Label}
+import ch.epfl.bluebrain.nexus.testkit.IOValues
+import monix.execution.Scheduler
+import org.scalatest.DoNotDiscover
+import org.scalatest.concurrent.Eventually
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+@DoNotDiscover
+class RemoteStorageClientSpec
```
imsdu, line 27: A test where the remote storage is down
A test where we don't have right access to see if we have a meaningful message because it seems to happen in prod

imsdu (changes requested):

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/remote/RemoteDiskStorageAccessSpec.scala
umbreak: No. This is correct, since all the suite gets called from `RemoteStorageSpec`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClient.scala
umbreak: I know. It is used anywhere under the sdk. It seems a good place to put it.

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/remote/client/RemoteStorageClientSpec.scala
umbreak: I cannot test the right access easily, since it will have to validate a token.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/remote/RemoteDiskStorageAccess.scala
umbreak: If you see the signatures, the `ActorSystem` is still needed for `S3DiskStorageAccess`. Scheduler is passed anyway to the `Storages` (notice I didn't have to change the current places where `evaluate` is being called.

So that seems to be the simplest solution.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClient.scala
imsdu: There is the package object in the sdk ?

bogdanromanx (approved):

imsdu (approved):

PR #1910, opened 2020-12-21
Schemas of resources can be fetched through resolvers

### LINKED ISSUE #1901: Creation of a resource depending on a schema resolved by a cross-project resolver fails
*`schema1` is a schema of org/proj
* org2/proj2 has a cross-project resolver pointing to it,

Creating a resource in org2/proj2 fails as `schema1` can't be found whereas it should be resolved

imsdu:
Fixes #1901


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceRejection.scala
```diff
@@ -98,6 +98,12 @@ object ResourceRejection {
         s"Resource '$id' failed to validate against the constraints defined in schema '$schema'"
       )
 
+  final case class InvalidSchemaRejection(
```
umbreak, line 101: Is there a reason why you decided to not wrap schema errors here, like we do for projects or orgs?

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceRejection.scala
imsdu: The errors we get here are now errors related to resolution via resolvers, not directly schemas.
The project/org related ones are already risen by fetching the project.
The only one that is left is related to schema deprecation and it was weird and verbose to raise a wrapped one in resources state machine to convert it back later.

umbreak (approved):

PR #1911, opened 2020-12-21
Fix tag routes

### LINKED ISSUE #1903: Tagging a resource fails
Tagging a resource fails with the message:
```
{
  "@type" : "MalformedQueryParamRejection",
  "reason" : "The query parameter 'rev' was malformed.",
  "details" : "the provided query parameter should not be present"
}
```
As the rev query parameter must be present, maybe a route collision ?

imsdu:
Fixes #1903 

Fixes:
* Tag route
* Tag regexp
* Tag creation http code

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/TagLabel.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/TagLabel.scala
@@ -1,0 +1,46 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.model
+
+import cats.syntax.all._
+import ch.epfl.bluebrain.nexus.delta.sdk.error.FormatError
+import ch.epfl.bluebrain.nexus.delta.sdk.error.FormatError.IllegalTagFormatError
+import io.circe.{Decoder, Encoder}
+
+/**
+  * A safe representation of a tag label
+  *
+  * @param value the string representation of the tag label
+  */
+final case class TagLabel private (value: String) {
+  override def toString: String = value
+}
+
+object TagLabel {
+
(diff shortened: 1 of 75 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/TagLabel.scala
```diff
@@ -0,0 +1,46 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.error.FormatError
+import ch.epfl.bluebrain.nexus.delta.sdk.error.FormatError.IllegalTagFormatError
+import io.circe.{Decoder, Encoder}
+
+/**
+  * A safe representation of a tag label
+  *
+  * @param value the string representation of the tag label
+  */
+final case class TagLabel private (value: String) {
```
umbreak, line 13: cannot be `extends Anyval` ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/TagLabel.scala
```diff
@@ -0,0 +1,46 @@
...
+  override def toString: String = value
+}
+
+object TagLabel {
+
+  private[sdk] val maxLength = 32
```
umbreak, line 19: Is that the only restriction? I would not like to allow tags with no-ascii characters...that could lead to undesired behaviour. I think I slight modification from Label should actually be enough:
```scala
  private[sdk] val regex = "[a-zA-Z0-9_-[.]]{1,32}".r
```

umbreak (changes requested):

wwajerowicz (approved):

umbreak (approved):

PR #1913, opened 2020-12-22
Merge contexts of ResourceF and its value when serializing

### LINKED ISSUE #1900: Resources types is not compacted according to its context
After creating a resource with the payload:
```
{
  "@context": {
    "nxv": "https://bluebrain.github.io/nexus/vocabulary/", 
    "other": "https://some.other.prefix.com/"
  },
  "@id": "https://dev.nexus.test.com/simplified-resource/xxx",
  "@type": "nxv:TestResource",
  "other:priority": 12,
  "other:projects": [
    "testProject",
    "testProject2"
  ]
}
```
When we later fetch it, the type is not compacted anymore as it was in v.1.4:
```
{
    "@context" : [
      {
        "nxv" : "https://bluebrain.github.io/nexus/vocabulary/"
        "other" : "https://some.other.prefix.com/"
      },
      "https://bluebrain.github.io/nexus/contexts/metadata.json"
    ],
    "@id" : "https://dev.nexus.test.com/simplified-resource/1",
    "@type" : "https://bluebrain.github.io/nexus/vocabulary/TestResource", <-- type is not compacted
    "_constrainedBy" : "https://dev.nexus.test.com/test-schema",
    "_createdBy" : "http://delta:8080/v1/realms/resourcesbrwredrxlrljkked/users/eaudhjeilkuprwyt",
    "_deprecated" : false,
    "_incoming" : "http://delta:8080/v1/resources/5xfjrp24pl03xrn/4omn7an4dnhepdc/test-schema/test-resource:1/incoming",
    "_outgoing" : "http://delta:8080/v1/resources/5xfjrp24pl03xrn/4omn7an4dnhepdc/test-schema/test-resource:1/outgoing",
    "_project" : "http://delta:8080/v1/projects/5xfjrp24pl03xrn/4omn7an4dnhepdc",
    "_rev" : 1,
    "_self" : "http://delta:8080/v1/resources/5xfjrp24pl03xrn/4omn7an4dnhepdc/test-schema/test-resource:1",
    "_updatedBy" : "http://delta:8080/v1/realms/resourcesbrwredrxlrljkked/users/eaudhjeilkuprwyt",
    "other:priority" : 5,
    "other:projects" : [
      "testProject2",
      "testProject"
    ]
  }```

### PR DESCRIPTION
**Pull request by imsdu**
Fixes #1900

### CODE REVIEW

File: delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala [hunk 1]
```diff
@@ -84,9 +88,44 @@ object ResourceF {
...
+    JsonLdEncoder.computeFromCirce(
+      _.id.resolvedAgainst(base.endpoint.toIri),
+      ContextValue(contexts.metadata).merge(valueContext)
+    )
+
   implicit final def resourceFUnitJsonLdEncoder(implicit base: BaseUri): JsonLdEncoder[ResourceF[Unit]] =
-    JsonLdEncoder.computeFromCirce(_.id.resolvedAgainst(base.endpoint.toIri), ContextValue(contexts.metadata))
+    resourceFUnitJsonLdEncoder(ContextValue.empty)
+
+  implicit def resourceFAJsonLdEncoder[A](implicit
```
umbreak, line 102: why are you reimplementing the `compose ` logic customized here? Why not to do it directly in `JsonLdEncoder.compose` ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala
```diff
@@ -84,9 +88,44 @@ object ResourceF {
       r.void.asJsonObject deepMerge r.value.asJsonObject
     }
 
+  private def resourceFUnitJsonLdEncoder(
+      valueContext: ContextValue
```
umbreak, line 92: What is this `ContextValue`? This is not always necessary, nor useful. If the `JsonLdEncoder[ResourceF[Unit]]` is implemented correctly, why you need this here?

umbreak (changes requested):
I would rather fix the issue once and for all making the necessary changes in `JsonLdEncoder` instead of hacking around the issue.

Wasn't working what I recommended?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala
imsdu: Because we can't create new JsonLdEncoders to create new ones like in your recommendation as they keep using the previous context and not the merged ones so it did not fix the bug.
It is the only place where compose is used in the code.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala
imsdu: It is just a private method used to create the encoder from other methods

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala
umbreak: Then there must be something else to be fixed. As far as I can see, most of the implementations of `JsonLdEncoder` in on the core and plugins codebase use `JsonLdEncoder.computeFromCirce`. In this case it should work, because when [compacting we do rely on the call to the context method](https://github.com/BlueBrain/nexus/blob/master/delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/encoder/JsonLdEncoder.scala#L152).

When changing the JsonLdEncoder on `compose` overriding the `context` method, should affect all those JsonLdEncoder instances built with `computeFromCirce`.

umbreak: for those cases, overriding the JsonLdContext in the `compose` function with something on the following lines should work:
```scala
...
val AA = A.withContext(context(value))
val BB = B.withContext(context(value))
...
```

imsdu: No because the self.compact calls the self.context and not the merge one we introduce.
And not all depend on Circe so we can't generalize it (Schemas and Resources for example).
So far only ResourceF is concerned.

imsdu: And apart from this cornercase, JsonLdEncoder is working fine so we don't need to change its interface to address this issue (and less on a branch supposed to fix a bug)

umbreak: `Schemas` and `Resources ` are exceptions and do not need that anyway, so it isn't a concern.

For the rest, I'm gonna do some checks myself and let you know.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala
```diff
@@ -84,9 +88,44 @@ object ResourceF {
...
+
+      override def context(value: ResourceF[A]): ContextValue =
+        A.context(value.value).merge(ContextValue(contexts.metadata))
 
-  implicit def resourceFAJsonLdEncoder[A: JsonLdEncoder](implicit base: BaseUri): JsonLdEncoder[ResourceF[A]] =
-    JsonLdEncoder.compose(rf => (rf.value, rf.void, rf.id.resolvedAgainst(base.endpoint.toIri)))
+      override def compact(
+          value: ResourceF[A]
+      )(implicit opts: JsonLdOptions, api: JsonLdApi, rcr: RemoteContextResolution): IO[RdfError, CompactedJsonLd] = {
+        val rfUnitJsonLdEncoder = resourceFUnitJsonLdEncoder(context(value))
```
umbreak, line 114: we could call it `metdataEncoder`, since ResourceF[Unit] is just metadata

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceF.scala
```diff
@@ -84,9 +88,44 @@ object ResourceF {
       r.void.asJsonObject deepMerge r.value.asJsonObject
     }
 
+  private def resourceFUnitJsonLdEncoder(
+      valueContext: ContextValue
+  )(implicit base: BaseUri): JsonLdEncoder[ResourceF[Unit]] =
+    JsonLdEncoder.computeFromCirce(
+      _.id.resolvedAgainst(base.endpoint.toIri),
+      ContextValue(contexts.metadata).merge(valueContext)
```
umbreak, line 96: if this method is called just within 

```scala
val rfUnitJsonLdEncoder = resourceFUnitJsonLdEncoder(context(value))
```

We don't need to add again  the contexts.metadata, since `context` method already has it. We can just use `valueContext` directly

umbreak (changes requested):

umbreak (approved):

umbreak (approved):

PR #1914, opened 2020-12-23
Blazegraph state machine.

wwajerowicz:


delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/model/BlazegraphViewType.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.nxv
+
+/**
+  * Enumeration of Blazegraph view types.
+  */
+sealed trait BlazegraphViewType extends Product with Serializable {
+
+  /**
```
umbreak, line 11: Can you also add

```scala
def tpe: Iri
def types: Set[Iri] = Set(tpe, nxv + "SparqlView")
```

I think the `tpe` method is used in some places

imsdu (dismissed):

umbreak (changes requested):

umbreak (approved):

PR #1934, opened 2021-01-08
Migration module for core features

### LINKED ISSUE #1539: Migration tool update for v1.5.0

imsdu:
Fixes #1539

Scan all events from the delta v1.4 keyspace and submit it to delta v1.5 modules to migrate them to the new model


delta/app/src/main/resources/app.conf
```diff
@@ -61,7 +61,7 @@ app {
       # the initial contact points
       contact-points = ["127.0.0.1:9042"]
       # the default keyspace
-      keyspace = "delta"
+      keyspace = "delta_1_5"
```
umbreak, line 64: why this is modified? If you need to modify it for migrations, you can do it somewhere else.

delta/migration/src/main/scala/ch/epfl/bluebrain/nexus/migration/Migration.scala
```diff
@@ -0,0 +1,514 @@
...
+                  updateResource(fixId(fixedSource, id))
+                }
+                .toTask
+                .void
+            case Deprecated(id, _, _, _, _, _, _)                                          =>
+              resources.deprecate(IriSegment(id), projectRef, None, cRev).toTask.void
+            // Tagging
+            case TagAdded(id, _, _, _, targetRev, tag, _, _)                               =>
+              // No information on resource type in tag event :'(
+              resources
```
umbreak, line 398: I guess for now it is alright, but to do it for all resource types we will have to do it in another way, folding over a list or something like that.

umbreak (commented):
Overall LGTM

delta/app/src/main/resources/app.conf
imsdu: We will need a new keyspace for the migrated data.
And you can't rename keyspaces/tables with Cassandra.
Or do we want to export data and reimport it after data is migrated to keep delta ?

delta/migration/src/main/scala/ch/epfl/bluebrain/nexus/migration/Migration.scala
imsdu: Yep, I have taken the easy path for now as we haven't decided how we will deal with storages/files/views.
I guess traits implemented by the plugins ?

delta/app/src/main/resources/app.conf
umbreak: but you can override configuration manually on the `MigrationModule`

imsdu: When you restart and run Delta afterwards, you will still need to run it on the new keyspace.
Only migration points to the old keyspace delta as I would like to avoid to run v1.5 by mistake on v1.4 data.

umbreak: I see. Well I guess is fine

delta/migration/src/main/scala/ch/epfl/bluebrain/nexus/migration/Migration.scala
umbreak: Yeah I guess so. So far we didn't need it for anything else, since the logic is handled by the routes. But we might need it for migrations

umbreak (dismissed):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/Main.scala
```diff
@@ -48,7 +48,17 @@ object Main extends BIOApp {
       (appConfig, mergedConfig) <- AppConfig.load(configNames, classLoader).handleError
       _                         <- initializeKamon(mergedConfig)
       pluginsContexts            = pluginsDef.map(_.remoteContextResolution)
-      modules                    = DeltaModule(appConfig, mergedConfig, classLoader, pluginsContexts)
+      modules                   <-
+        if (sys.env.isDefinedAt("MIGRATE_DATA"))
+          UIO.delay {
+            log.info("Starting Delta in normal mode")
```
bogdanromanx, line 54: the logs should be reversed.

bogdanromanx (approved):

umbreak (dismissed):

bogdanromanx (approved):

PR #1938, opened 2021-01-12
Added file routes

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1859

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
@@ -3,7 +3,7 @@ package ch.epfl.bluebrain.nexus.delta.plugins.storage.files
 import akka.actor.typed.ActorSystem
 import akka.actor.{ActorSystem => ClassicActorSystem}
 import akka.http.scaladsl.model.ContentTypes.`application/octet-stream`
-import akka.http.scaladsl.model.{ContentType, HttpEntity, HttpHeader, Uri}
+import akka.http.scaladsl.model.{ContentType, HttpEntity, Uri}
 import akka.persistence.query.Offset
 import cats.effect.Clock
 import cats.syntax.all._
@@ -15,14 +15,14 @@ import ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model.FileEvent._
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model.FileRejection._
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model.FileState.{Current, Initial}
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model._
+import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.Storages
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.StorageRejection.StorageIsDeprecated
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.{DigestAlgorithm, Storage, StorageType}
-import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.Storages
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.utils.EventLogUtils
 import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
 import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.ContextValue
 import ch.epfl.bluebrain.nexus.delta.sdk._
-import ch.epfl.bluebrain.nexus.delta.sdk.directives.DeltaDirectives.{unacceptedMediaTypeRejection, FileData}
+import ch.epfl.bluebrain.nexus.delta.sdk.directives.FileResponse
 import ch.epfl.bluebrain.nexus.delta.sdk.jsonld.ExpandIri
 import ch.epfl.bluebrain.nexus.delta.sdk.model._
 import ch.epfl.bluebrain.nexus.delta.sdk.model.acls.AclAddress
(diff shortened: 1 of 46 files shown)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -203,43 +206,49 @@ final class Files(
...
     * @param project   the project where the storage belongs
+    * @param headers   the HTTP headers to decide whether or not the content can be fetched (using its accept headers)
     */
   def fetchContent(
       id: IdSegment,
-      project: ProjectRef
-  )(implicit caller: Caller): IO[FileRejection, AkkaSource] =
-    fetchContent(id, project, None).named("fetchFileContent", moduleType)
+      project: ProjectRef,
+      headers: Seq[HttpHeader] = Seq.empty
```
imsdu, line 214: We could check headers in the route no so as not to handle this in this layer ?
And if at one point we fetch files from another plugin for example, http headers won't be relevant.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/Digest.scala
```diff
@@ -21,4 +23,9 @@ object Digest {
     * A digest that does not yield a value because it is still being computed
     */
   final case object NotComputedDigest extends Digest
+
+  implicit val digestEncoder: Encoder.AsObject[Digest] = Encoder.encodeJsonObject.contramapObject {
+    case ComputedDigest(algorithm, value) => JsonObject("_algorithm" -> algorithm.asJson, "_value" -> value.asJson)
+    case NotComputedDigest                => JsonObject("_value" -> "".asJson)
```
imsdu, line 29: Shouldn't it be JNull ? Unless it was implemented this way in v1.4 ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileEvent.scala
```diff
@@ -124,4 +144,51 @@ object FileEvent {
...
+  ): JsonLdEncoder[FileEvent] = {
+    implicit val subjectEncoder: Encoder[Subject] = Identity.subjectIdEncoder
+
+    implicit val encoder: Encoder.AsObject[FileEvent] = Encoder.encodeJsonObject.contramapObject { event =>
+      val storageAndType       = event match {
+        case created: FileCreated => Some(created.storage -> created.storageType)
+        case updated: FileUpdated => Some(updated.storage -> updated.storageType)
+        case _                    => None
+      }
+      implicit val storageType = storageAndType.map(_._2).getOrElse(StorageType.DiskStorage)
```
imsdu, line 176: Is there way to get rid of the default value ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/DeltaDirectives.scala
```diff
@@ -20,29 +19,40 @@ import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.JsonLd
...
       `application/ld+json`,
       `application/json`,
       `application/n-triples`,
       `text/vnd.graphviz`
     )
 
+  /**
+    * Filename, file contentType and file data
+    */
+  type FileData = (String, ContentType, AkkaSource)
```
imsdu, line 51: A case class in the model package would be clearer

imsdu (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
umbreak: Yes we could, but that made the actual logic quite complicated in the routes.
If they are not relevant is not important, that's why they have a default value

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/Digest.scala
umbreak: It was indeed like this in 1.4.x. It returned a value as an empty string

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileEvent.scala
umbreak: Not really. But it doesn't matter much. This is just used here.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/DeltaDirectives.scala
umbreak: I didn't want to bother wrapping things up, since it isn't broadly used, but I can do that

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
umbreak: I've just added methods to handle the case of calling files.fetch from somewhere outside of the routes where headers are not important.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutes.scala
```diff
@@ -0,0 +1,266 @@
...
+                        }
+                      }
+                    }
+                  }
+                },
+                // Create a file without id segment
+                (post & pathEndOrSingleSlash & noParameter("rev") & parameter("storage".as[IdSegment].?)) { storage =>
+                  extractRequestEntity { entity =>
+                    operationName(s"$prefixSegment/files/{org}/{project}") {
+                      emit(Created, files.create(storage, ref, entity))
```
wwajerowicz, line 111: In all the other routes, we do the authorization in the routes, but for files we do it in Files, it's a bit inconsistent.

wwajerowicz (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutes.scala
umbreak: Yes it is. But there is a good reason for it. File authorization depends on storage and not on a hardcoded value. 

All the logic about fetching a storage, expanding its IdSegment, checking if it is deprecated, etc it is done inside the `Files` bundle. So is the authorization.

This leaves the routes less verbose, while the logic is concentrated in the bundle.

imsdu (approved):

Issue #1941, opened 2021-01-13
File link not implemented


umbreak:
Missing implementation of [linking a file](https://bluebrainnexus.io/docs/delta/api/current/kg-files-api.html#link-an-existing-file-using-post)

PR #1942, opened 2021-01-13
File attributes with no digest are updated asynchronously for remote disk storage uploads

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1922

There are some missing tests, but I'll add them when this issue is implemented https://github.com/BlueBrain/nexus/issues/1941 (since it is somewhat related: there are no empty digests if we don't do a moveFile)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/StorageFileRejection.scala
```diff
@@ -38,6 +38,14 @@ object StorageFileRejection {
           s"File cannot be fetched from path '$path' for unexpected reasons. Details '$details'"
         )
 
+    /**
+      * Rejection performing this operation because the storage does not support it
+      */
+    final case class UnsupportedOperation(tpe: StorageType)
```
imsdu, line 44: The type and the message don't match

imsdu (changes requested):

imsdu (approved):

wwajerowicz (approved):

PR #1946, opened 2021-01-14
Blazegraph api

wwajerowicz:


delta/plugins/blazegraph/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViewsSpec.scala
```diff
@@ -1,274 +1,456 @@
...
       includeMetadata = false,
       includeDeprecated = false,
-      Permission.unsafe("my/permission")
+      Permission.unsafe("views/query")
+    )
+    val indexingSource = jsonContentOf("indexing-view-source.json")
+
+    val updatedIndexingValue  = indexingValue.copy(resourceTag = Some(TagLabel.unsafe("v1.5")))
+    val updatedIndexingSource =
+      jsonContentOf("indexing-view-source.json").mapObject(_.add("resourceTag", Json.fromString("v1.5")))
```
umbreak, line 69: indexingSource.mapObject

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/serialization/EventSerializer.scala
```diff
@@ -0,0 +1,76 @@
...
+  override def manifest(o: AnyRef): String = o match {
+    case _: BlazegraphViewEvent => blazegraphViewsEventManifest
+    case _                      =>
+      throw new IllegalArgumentException(
+        s"Unknown event type '${o.getClass.getCanonicalName}', expected BlazegraphViewEvent"
+      )
+  }
+
+  override def toBinary(o: AnyRef): Array[Byte] = o match {
+    case e: BlazegraphViewEvent => e.asJson.noSpaces.getBytes(StandardCharsets.UTF_8)
```
umbreak, line 38: we should probably use a printer:
```scala
   private val printer: Printer = Printer. noSpaces.copy(dropNullValues = true)
   printer.print(e.asJson)
```
To avoid printing null values (we should do that in all `EventSerializer`, in this case we will gain few bytes here and there)

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
```diff
@@ -1,17 +1,313 @@
...
+    */
+  def create(project: ProjectRef, source: Json)(implicit
+      subject: Subject
+  ): IO[BlazegraphViewRejection, BlazegraphViewResource] = {
+    for {
+      p                <- projects.fetchActiveProject(project)
+      (iri, viewValue) <- sourceDecoder(p, source)
+      res              <- eval(CreateBlazegraphView(iri, project, viewValue, source, subject), p)
+    } yield res
+  }.named("createView", moduleType)
```
umbreak, line 70: Maybe `createBlazegraphView` instead? It would be useful to have different spans for different views, since from the API perspective we would not always be able to differentiate them

umbreak (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/model/BlazegraphViewRejection.scala
```diff
@@ -40,8 +43,10 @@ object BlazegraphViewRejection {
     * Rejection returned when attempting to update a view that doesn't exist.
     *
     * @param id the view id
+    * @param project the project it belongs to
     */
-  final case class ViewNotFound(id: Iri) extends BlazegraphViewRejection(s"View '$id' not found.")
+  final case class ViewNotFound(id: Iri, project: ProjectRef)
+      extends BlazegraphViewRejection(s"View '$id' not found in project '$project]'.")
```
imsdu, line 49: Blazegraph view would disembiguate the message with Elastic/Composite views
There is also an extra ] to remove

imsdu (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
```diff
@@ -1,17 +1,313 @@
...
+      }
+      .named("fetchViewBy", moduleType)
+
+  /**
+    * List views.
+    *
+    * @param pagination the pagination settings
+    * @param params     filtering parameters for the listing
+    */
+  def list(
```
umbreak, line 234: I don't think we need this method. Listings from the API are going to be provided by elasticsearch plugin.

However it might be useful for other purposes, outside of the API.

umbreak (changes requested):

umbreak (approved):

imsdu (approved):

Issue #1869, opened 2020-12-08
Snapshotting fails for the different states because of serialization


imsdu:
The different CurrentState of the different entities are missing in the serialization-bindings (we will have to add some tests too to ensure we don't break compatibility when adding more fields to it)
```
2020-12-08 16:34:38 WARN  a.s.Serialization(akka://delta) - Using the Java serializer for class [ch.epfl.bluebrain.nexus.delta.sdk.model.realms.RealmState$Current] which is not recommended because of performance implications. Use another serializer or disable this warning using the setting 'akka.actor.warn-about-java-serializer-usage'
2020-12-08 16:34:38 WARN  a.s.DisabledJavaSerializer - Outgoing message attempted to use Java Serialization even though `akka.actor.allow-java-serialization = off` was set! Message type was: [class ch.epfl.bluebrain.nexus.delta.sdk.model.realms.RealmState$Current]
2020-12-08 16:34:38 WARN  c.e.b.n.s.p.EventSourceProcessor - Failed to save snapshot given metadata [SnapshotMetadata(realm-realmkodmldqdetdpyaqc, 4, 0, None)] due to: Attempted to serialize message using Java serialization while `akka.actor.allow-java-serialization` was disabled. Check WARNING logs for more details.
```

imsdu:
Not relevant anymore as we are going away from the akka-persistence library

PR #1949, opened 2021-01-18
Disable snapshotting when deprecating

imsdu:
As long as #1869 is not fixed, snapshotting is now disabled.


bogdanromanx (approved):

umbreak:
Can you please add a TODO?
```scala
// TODO: We want to allow snapshotting but we have to provide serailization-bindings for State.Current
```

umbreak (changes requested):
Can you please add a TODO?
```scala
// TODO: We want to allow snapshotting but we have to provide serailization-bindings for State.Current
```

imsdu:
> Can you please add a TODO?
> 
> ```scala
> // TODO: We want to allow snapshotting but we have to provide serailization-bindings for State.Current
> ```

There is an open issue, it is ok as it is no ?

umbreak:
Yeah...as long as I trust opened issues, it is going to be way faster to look for all TODOs in the code before we release 1.5.0 than to go issue by issue.

umbreak (approved):

PR #1953, opened 2021-01-20
Storage plugin bootstrap

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1864


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/PluginsInfoRoutes.scala
```diff
@@ -0,0 +1,65 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.search.SearchResults._
+import ch.epfl.bluebrain.nexus.delta.sdk.plugin.PluginInfo
+import ch.epfl.bluebrain.nexus.delta.sdk.{Acls, Identities}
+import io.circe.Encoder
+import io.circe.generic.semiauto.deriveEncoder
+import kamon.instrumentation.akka.http.TracingDirectives.operationName
+import monix.bio.UIO
+import monix.execution.Scheduler
+
+class PluginsInfoRoutes(identities: Identities, acls: Acls, info: List[PluginInfo])(implicit
```
imsdu, line 25: For now, it is ok but in the end it would be great to have this information merged with the other application information/health status

umbreak: yeah, maybe. Right now the `/version` endpoint shows version information about the services (cassandra, elasticsearch, blazegraph and remote storage service), which is not the same as the plugins.

But maybe we could have in them in the same endpoint, under a `plugins` json object. To be discussed

umbreak: I've added an issue for it https://github.com/BlueBrain/nexus/issues/1954

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePluginConfig.scala
```diff
@@ -0,0 +1,24 @@
...
+import pureconfig.{ConfigReader, ConfigSource}
+
+final case class StoragePluginConfig(storages: StoragesConfig, files: FilesConfig)
+
+object StoragePluginConfig {
+
+  /**
+    * Converts a [[Config]] into an [[StoragePluginConfig]]
+    */
+  def load(config: Config): StoragePluginConfig =
```
imsdu, line 16: Not a priority, could be done in 1.6 for example but we could handle config loading like for AppConfig

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/plugin/PluginsLoader.scala
```diff
@@ -0,0 +1,92 @@
...
+      pluginClassLoader <- IO.delay(new PluginClassLoader(jar.toURI.toURL, parentClassLoader)).hideErrors
+      pluginDefClasses  <- IO.delay(loadPluginDefClasses(pluginClassLoader)).hideErrors
+      pluginDef         <- pluginDefClasses match {
+                             case pluginDef :: Nil =>
+                               IO.pure(Some(pluginClassLoader.create[PluginDef](pluginDef, println)()))
+                             case Nil              =>
+                               logger.warn(s"Jar file '$jar' does not contain a 'PluginDef' implementation.")
+                               IO.pure(None)
+                             case multiple         =>
+                               IO.raiseError(MultiplePluginDefClassesFound(jar, multiple.toSet))
```
wwajerowicz, line 48: If we log and ignore when there is no `PluginDef` found, we should do the same when there are multiple.

wwajerowicz (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/PluginDef.scala
```diff
@@ -11,17 +12,34 @@ trait PluginDef {
...
     * Plugin information
     */
   def info: PluginInfo
 
+  /**
+    * The priority of this plugin.
+    * This value will decide the order in which this plugin takes is executed compared to the rest of the plugins.
+    * It affects Routes ordering and classpath ordering
+    */
+  def priority: Int
```
imsdu, line 30: If we add priority, we can't handle it in the code, we will have to handle them via configuration as it really depends on which plugins are here

delta/plugins/storage/src/test/resources/application.conf
```diff
@@ -1,6 +1,6 @@
 include required(classpath("akka-test.conf"))
 # All application specific configuration and defaults should reside here
-app {
+storage-plugin {
```
imsdu, line 3: We could just include storage.conf and override what is needed

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/context/RemoteContextResolution.scala
```diff
@@ -47,6 +47,18 @@ trait RemoteContextResolution {
...
       case _                     => Set.empty
     }
+
+  /**
+    * Merges the current [[RemoteContextResolution]] with the passed ones
+    */
+  def merge(others: RemoteContextResolution*): RemoteContextResolution =
+    new RemoteContextResolution {
+      override def resolve(iri: Iri): Result[Json] = {
+        others.map(_.resolve(iri)).toList.foldLeft(self.resolve(iri)) { (acc, c) =>
```
imsdu, line 57: It would be better to use foldLeftM which could provide short-circuiting when we succeed

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/ClasspathResourceUtils.scala
```diff
@@ -21,10 +21,10 @@ trait ClasspathResourceUtils {
     * @return the content of the referenced resource as an [[InputStream]] or a [[ClasspathResourceError]] when the
     *         resource is not found
     */
-  def ioStreamOf(resourcePath: String): IO[ClasspathResourceError, InputStream] =
+  def ioStreamOf(resourcePath: String, classLoader: ClassLoader): IO[ClasspathResourceError, InputStream] =
```
imsdu, line 24: the classloader could be an implicit no ? As we should only use one in a given context

imsdu (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/plugin/PluginsLoader.scala
umbreak: I think it is different. There could be any random jar file in the directory that gets scanned which we don't care about. That's fine to issue a warning and continue.

But if there is a jar that contains an implementation of `PluginDef`it is meant to be used as a plugin. And if it contains multiple, it is an error that should not let the application start, but fail.

build.sbt
```diff
@@ -523,8 +527,11 @@ lazy val storagePlugin = project
...
     moduleName                 := "delta-storage-plugin",
     libraryDependencies       ++= Seq(
-      akkaHttpXml,
-      alpakkaS3,
+      akkaHttpXml exclude ("com.typesafe.akka", "akka-http_2.13"),
+      alpakkaS3 excludeAll (
+        ExclusionRule(organization = "com.typesafe.akka", name = "akka-stream_2.13"),
+        ExclusionRule(organization = "com.typesafe.akka", name = "akka-http_2.13")
+      ),
       "io.kamon"       %% "kamon-akka-http" % kamonVersion % Provided,
```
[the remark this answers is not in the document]
umbreak: I've tried that but it does not work, since it removes totally the dependency even on the compile scope and then obviously I get errors loke:
```
[error] /Users/dmontero/github/nexus/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePlugin.scala:4:34: Unused import
[error] import akka.http.scaladsl.server.Route
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/plugin/PluginDef.scala
umbreak: Priority is now read from configuration present in each plugin and it can be overridden by system properties

wwajerowicz (approved):

imsdu (approved):

PR #1962, opened 2021-01-22
Allow getting cache per project (or whatever composite key we want)

imsdu:
Allows to speed up retrieving resolvers/storages per project as we don't need to scan them all anymore.

### DIFF
--- a/delta/app/src/main/resources/app.conf
+++ b/delta/app/src/main/resources/app.conf
@@ -220,7 +220,7 @@ app {
     key-value-store {
       ask-timeout = 15 seconds
       consistency-timeout = 5 seconds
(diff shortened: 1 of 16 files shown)


umbreak (approved):

bogdanromanx (approved):
does the restoration on service restart work with this implementation?

imsdu:
> does the restoration on service restart work with this implementation?
Looks like it, when I restarted, I did not get any error.

bogdanromanx:
It would be useful to exercise the functionality (ideally with a large number of projects) to see if the data will be available since the Map of distributed caches are indexed per node in memory; so until an entry is added to the cache, the distributed cache will not be created, thus not materialized from disk.

PR #1963, opened 2021-01-22
Backporting sourcing modifications

imsdu:


umbreak (approved):

delta/sourcing/src/main/resources/scripts/cassandra.ddl
```diff
@@ -148,6 +148,6 @@ CREATE TABLE IF NOT EXISTS delta.projections_failures
     value          text,
     error_type     varchar,
     error          text,
-    PRIMARY KEY (projection_id, offset, persistence_id, sequence_nr)
+    PRIMARY KEY (projection_id, timestamp, persistence_id, sequence_nr)
 )
-            WITH CLUSTERING ORDER BY (offset ASC);
\ No newline at end of file
+            WITH CLUSTERING ORDER BY (timestamp ASC);
```
umbreak, line 153: Is the clustering order any useful at all? Do we use it?
The current combination `PRIMARY KEY / WITH CLUSTER ORDER` does not make any sense.

We could have a compound primary key, if you want ordering in any useful way:
```
PRIMARY KEY ((projection_id, persistence_id), timestamp)
   WITH CLUSTERING ORDER BY (timestamp ASC);
```
In this case, when doing:
```
SELECT * WHERE (projection_id=XXX AND persistence_id=XXX);
```
the results will be ordered by timestamp the given projection_id & persistence_id

Another option could be:
```
PRIMARY KEY ((projection_id), persistence_id, timestamp)
   WITH CLUSTERING ORDER BY (persistence_id ASC, timestamp ASC);
```
In this case, when doing:
```
SELECT * WHERE (projection_id=XXX);
```
the results will be ordered by persistence_id and timestamp the given projection_id

In case we don't need ordering, just remote it.

More info: https://www.datastax.com/blog/we-shall-have-order

umbreak (changes requested):

umbreak (approved):

PR #1971, opened 2021-01-27
Added ElasticSearch client

umbreak:


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ServiceDescription.scala
```diff
@@ -0,0 +1,24 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.client
+
+import io.circe.Decoder
+
+/**
+  * Information about the deployed service
+  *
+  * @param version service version
+  */
+final case class ServiceDescription(version: String) {
```
imsdu, line 10: The service description could be in the sdk no ? As we also have one for Blazegraph ?

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
```diff
@@ -0,0 +1,260 @@
...
+    else indices.map(sanitize(_, allowWildCard = true)).mkString(",")
+
+  /**
+    * Replaces the characters ' "\<>|,/?' in the provided index with '_' and drops all '_' prefixes.
+    * The wildcard (*) character will be only dropped when ''allowWildCard'' is set to false.
+    *
+    * @param index the index name to sanitize
+    * @param allowWildCard flag to allow wildcard (*) or not.
+    */
+  private def sanitize(index: String, allowWildCard: Boolean): String = {
```
imsdu, line 201: Instead of sanitizing here, we could maybe pass something other than a string when providing an index.
So that the caller has to behave himself a minimum when passing an index and no magic happens here

imsdu (changes requested):

imsdu (approved):

PR #1974, opened 2021-01-28
Global event log

wwajerowicz:


delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/Journal.scala
```diff
@@ -18,9 +19,12 @@ import monix.bio.{IO, Task, UIO}
   */
 private[testkit] class Journal[Id, E <: Event] private (
     events: IORef[Vector[Envelope[E]]],
+    eventsByTag: IORef[Map[String, Vector[Envelope[E]]]],
```
umbreak, line 22: Do we need `eventsByTag` as an extra parameter here? Couldn't we derive it from `events`?
Since we can access each `E.eventType`, from a `Vector[Envelope[E]]` we can generate the Map

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/EventResolver.scala
```diff
@@ -0,0 +1,24 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.eventlog
+
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.ExpandedJsonLd
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{Event, ResourceF}
+import monix.bio.Task
+
+/**
+  * Resolver which allows to fetch the latest state for an [[Event]].
+  */
+trait EventResolver {
```
umbreak, line 10: A resolver again in a class name. We have a lot of *Resolver* already. 
What about `EventState`

And then instead of `latestStateAsExpandedJsonLd` we just have

```scala
trait EventState {
   def eventType: String
   def apply(e: Event): Task[Option[ResourceF[ExpandedJsonLd]]]
}
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/GlobalEventLog.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+    * Get stream of all [[Event]]s for a project.
+    *
+    * @param project  the project reference
+    * @param offset   the offset to start from
+    */
+  def events(project: ProjectRef, offset: Offset): IO[ProjectNotFound, Stream[Task, Envelope[Event]]]
+
+  /**
```
umbreak, line 27: It is missing:

```scala
def events(org: Label, offset: Offset): IO[ProjectNotFound, Stream[Task, Envelope[Event]]]
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/GlobalEventLog.scala
```diff
@@ -0,0 +1,34 @@
...
+    * Fetch the latest for the event as expanded JSON-LD.
+    *
+    * @param event  the event for which to fetch latest state
+    */
+  def latestStateAsExpandedJsonLd(event: Event): Task[Option[ResourceF[ExpandedJsonLd]]]
```
umbreak, line 32: `latestState` probably is enough, or `expandedLatestState`

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
```diff
@@ -506,8 +506,9 @@ object BlazegraphViews {
       evaluate = evaluate(validateP, validateRef),
       tagger = (event: BlazegraphViewEvent) =>
         Set(
+          Event.eventTag,
           moduleType,
-          s"${Projects.moduleType}=${event.project}",
+          Projects.projectTag(event.project),
           s"${Organizations.moduleType}=${event.project.organization}"
```
umbreak, line 512: Probably we want the same here: `Organizations.orgTag(event.project.organization)`

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/model/BlazegraphViewEvent.scala
```diff
@@ -43,6 +44,11 @@ sealed trait BlazegraphViewEvent extends Event {
     * @return the revision that the event generates
     */
   def rev: Long
+
+  /**
+    * @return the type of this event
+    */
+  def eventType: String = BlazegraphViews.moduleType
```
umbreak, line 51: Maybe we can have a custom type for the moduleType (eventType) concept?

Something like 

```scala 
trait ModuleType {
   def name: String
}

object ModuleType {
   final case object ProjectModule extends ModuleType("project")
   // others
}
```

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
```diff
@@ -21,14 +22,17 @@ import ch.epfl.bluebrain.nexus.delta.sdk.model.resources.ResourceRejection._
...
+import com.typesafe.scalalogging.Logger
 import fs2.Stream
 import io.circe.Json
 import monix.bio.{IO, Task, UIO}
 
 /**
   * Operations pertaining to managing resources.
   */
-trait Resources {
+trait Resources extends EventResolver {
```
umbreak, line 33: Why `EventResolver` is only implemented here and not in `Resolvers`, `Schemas`, `Storages`, `BlazegraphViews`, `ElasticSearchViews`?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
```diff
@@ -216,6 +220,21 @@ trait Resources {
...
+  override def eventType: String = Resources.moduleType
+
+  override def latestStateAsExpandedJsonLd(e: Event): Task[Option[ResourceF[ExpandedJsonLd]]] = e match {
+    case rEv: ResourceEvent =>
+      fetch(IriSegment(rEv.id), rEv.project, None)
+        .map(res => Some(res.map(_.expanded)))
+        .onErrorHandleWith(_ => Task.pure(None))
+
+    case ev =>
+      logger.warn(
```
umbreak, line 232:
```scala
Task.delay(logger.warn(...)).as(None)
```

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/EventResolver.scala
```diff
@@ -0,0 +1,24 @@
...
+    * The type of [[Event]] this resolver resolves
+    */
+  def eventType: String
+
+  /**
+    * Fetch the latest state for a given [[Event]]
+    *
+    * @param e the event for which to fetch the latest state.
+    */
+  def latestStateAsExpandedJsonLd(e: Event): Task[Option[ResourceF[ExpandedJsonLd]]]
```
umbreak, line 22: expandedLatestState

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
wwajerowicz: I wanted to keep this PR small, the rest can be implemented in another PR.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/GlobalEventLog.scala
imsdu: Where do we need to use this method ?
In the tests, we use it with it after getting the stream of events ?
Maybe we could pass directly the envelope with the event which already contains the event type ?
This way we would not have to touch the Event contract ?

imsdu (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/model/BlazegraphViewEvent.scala
imsdu: I don't know if using the module type is the best idea in the long run as these are two different things.
Sure for now, it fits as we have a one-to-one relationship but this may change.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/GlobalEventLog.scala
wwajerowicz: We will need it in Blazegraph indexing to get the latest state for a given event. The eventType in `Envelope` is a specific event type(e.g. `ResourceCreated`), but here we need an event type which encompasses all event of a given type(all `ResourceEvent`s).

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/Journal.scala
```diff
@@ -18,9 +19,12 @@ import monix.bio.{IO, Task, UIO}
...
     semaphore: IOSemaphore,
-    entityType: String
-)(implicit idLens: Lens[E, Id]) {
+    entityType: String,
+    tagger: E => Set[String]
+)(implicit idLens: Lens[E, Id])
+    extends EventLog[Envelope[E]] {
```
imsdu, line 27: If we have an dummy/test implementation for Aggregate and KeyValueStore, the Dummy and the Service implementation will get quite similar :)

imsdu (changes requested):

umbreak (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/eventlog/ExpandedGlobalEventLog.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  override def events(offset: Offset): Stream[Task, ResourceF[ExpandedJsonLd]] = exchange(
+    eventLog.eventsByTag(Event.eventTag, offset)
+  )
+
+  override def events(
+      project: ProjectRef,
+      offset: Offset
+  ): IO[ProjectNotFound, Stream[Task, ResourceF[ExpandedJsonLd]]] =
+    projects.fetch(project).map(_ => exchange(eventLog.eventsByTag(Projects.projectTag(project), offset)))
```
umbreak, line 36: `.as` instead of `.map`

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/eventlog/ExpandedGlobalEventLog.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  override def events(org: Label, offset: Offset): IO[OrganizationNotFound, Stream[Task, ResourceF[ExpandedJsonLd]]] =
+    orgs.fetch(org).map(_ => exchange(eventLog.eventsByTag(Organizations.orgTag(org), offset)))
```
umbreak, line 39: same

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/GlobalEventLog.scala
```diff
@@ -0,0 +1,35 @@
...
+import fs2.Stream
+import monix.bio.{IO, Task}
+
+trait GlobalEventLog[T] {
+
+  /**
+    * Get stream of all [[Event]]s
+    * @param offset the offset to start from
+    */
+  def events(offset: Offset): Stream[Task, T]
```
umbreak, line 17: I changed to to `apply` because it seems enough for me. You're not fetching events anymore, you're fetching `T`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/EventExchange.scala
```diff
@@ -0,0 +1,49 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{Event, ResourceF}
+import com.typesafe.scalalogging.Logger
+import monix.bio.Task
+
+import scala.reflect.ClassTag
+
+/**
+  * Resolver which allows to fetch the latest state for an [[Event]].
+  */
+trait EventExchange[E <: Event] {
```
umbreak, line 13: This way you decided it to do it, instead of the one I provided as an example has some drawbacks:
1. You have to cast twice (one when calling appliesTo and another when calling toExpanded or toCompacted)
3. You have to provide many more methods on the same class.
3. You have to define the generic method `toCompacted` and `toExpanded` implementing this logger which usually it should't be required at all, since the logging already happens on the GlobalEventLog
3. You have this generic type which gets passed along everywhere (event though it is useless cause it ends up being `_`), instead of an abstract type which does not pollute the rest of the code.

Having an intermediate class `EventExchangeApply` (or whatever you wanna call it) solves most of the issues and keeps the class `EventExchange` with a single responsibility.

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/model/BlazegraphViewEvent.scala
umbreak, line 51: we don't need this

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewEvent.scala
```diff
@@ -28,6 +29,11 @@ sealed trait ElasticSearchViewEvent extends Event {
     * @return the view unique identifier
     */
   def uuid: UUID
+
+  /**
+    * @return the type of this event
+    */
+  def eventType: String = ElasticSearchViews.moduleType
```
umbreak, line 36: we don't need this

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ResourcesDummy.scala
```diff
@@ -220,9 +220,35 @@ object ResourcesDummy {
       contextResolution: ResolverContextResolution
   )(implicit clock: Clock[UIO], uuidF: UUIDF): UIO[ResourcesDummy] =
     for {
-      journal <- Journal(moduleType)
+      journal <- Journal(
+                   moduleType,
+                   1L,
+                   (ev: ResourceEvent) =>
+                     Set("event", Projects.projectTag(ev.project), Organizations.orgTag(ev.project.organization))
```
umbreak, line 227: isn't it missing also the `Resources.moduleType`?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Resources.scala
```diff
@@ -364,4 +366,24 @@ object Resources {
...
+  /**
+    * Create an instance of [[EventExchange]] for [[ResourceEvent]].
+    * @param resources  resources operation bundle
+    */
+  def eventExchange(resources: Resources): EventExchange[ResourceEvent] = new EventExchange[ResourceEvent] {
+    override protected def fetchExpanded(event: ResourceEvent): Task[ResourceF[ExpandedJsonLd]] =
+      resources
+        .fetch(IriSegment(event.id), event.project, None)
+        .map(res => res.map(_.expanded))
+        .hideErrorsWith(rej => new IllegalArgumentException(rej.reason))
```
umbreak, line 379: I'm not sure about converting all errors to `IllegalArgumentException`, but I don't have a quick solution in mind

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/EventExchange.scala
imsdu: If we have something like this:
```scala
EventExchange[E <: Event, S](fetch: E => IO[SomeError, State])(implicit c: ClassTag, jsonldEncoder: JsonLdEncoder[S])
```
Wouldn't that make things simpler and remove the need of a trait as we would have all what we need ?

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/eventlog/ExpandedGlobalEventLog.scala
```diff
@@ -0,0 +1,74 @@
...
+    eventLog: EventLog[Envelope[Event]],
+    projects: Projects,
+    orgs: Organizations,
+    eventExchanges: EventExchangeCollection
+) extends GlobalEventLog[ResourceF[ExpandedJsonLd]] {
+
+  private val logger: Logger = Logger[ExpandedGlobalEventLog]
+
+  override def events(offset: Offset): Stream[Task, ResourceF[ExpandedJsonLd]] = exchange(
+    eventLog.eventsByTag(Event.eventTag, offset)
```
imsdu, line 29: We could add a TODO here to add an operation to just keep the last event for a given persistenceId (and go back to it later)

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/GlobalEventLog.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import akka.persistence.query.Offset
+import ch.epfl.bluebrain.nexus.delta.sdk.model.Label
+import ch.epfl.bluebrain.nexus.delta.sdk.model.organizations.OrganizationRejection.OrganizationNotFound
+import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectRef
+import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectRejection.ProjectNotFound
+import fs2.Stream
+import monix.bio.{IO, Task}
+
+trait GlobalEventLog[T] {
```
imsdu, line 11: IndexingLog ?

imsdu (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/eventlog/ExpandedGlobalEventLog.scala
umbreak: I don't see the signature to that `EventExchange` more useful than the current. The JsonLdEncoder[S] part is misleading there, since in many cases you don't want to perform any encoding (it is already present in the `State`).

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/GlobalEventLog.scala
umbreak: We are going to use that for SSEs too, that's why we added the generic `T`, so `IndexingLog` is not adequate.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/GlobalEventLog.scala
wwajerowicz: I would rename it to `stream`, seems to give more information than just `apply`.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/EventExchange.scala
umbreak: I don't see the signature to that `EventExchange` more useful than the current. The `JsonLdEncoder[S]` part is misleading, since in many cases you don't want to perform any encoding (it is already present in the State).

imsdu: I should not have put state but as it is the resource built from the state, my bad.
And we don't know at this level that it works that way and we should not care, it is up to the JsonLdEncoder that we already use for the different types of resources to take care of it.

wwajerowicz: @umbreak 
re 1. the cast in `appliesTo` is only done once, when building the index, 
re2.  I find it cleaner to have those methods in a single class rather than spread through two
re3 logging in 2 places notifies about two possible errors: no `EventExchange` for a given event and wrong event passed to event exchange(the latter shouldn't happen though).
re4 I agree that having `[_ <: Event]` everywhere is quite annoying, but I personally find the version with  `EventExchangeApply` quite difficult to follow.

umbreak: I'm fine if you wanna keep this version. The please try the following:
- Try to use an abstract type instead of the generic type. It should work and you should be able to get rid of the `[_ <: Event]`
- Remove the logging from here. This is a log level implementation, which returns a Task[Option] which is just fine. The logging, if necessary, should be done in the higher level place where this is used. And I believe is not even necessary, since this is already handled in the globaleventlog

umbreak (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/eventlog/GlobalEventLog.scala
imsdu: We are not sure if this can be applied to SSEs or if another approach will be better.

Issue #1356, opened 2020-07-29
Prevent multiple resolvers from having the same priority in the scope of a project


umbreak:
In the case multiple `CrossProjectResolver` resources with the same `priority` value exists for the same project, the resolution of resources might not be deterministic, since the [ordering of resolvers is done just by priority](https://github.com/BlueBrain/nexus/blob/master/delta/src/main/scala/ch/epfl/bluebrain/nexus/kg/cache/ResolverCache.scala#L60).

That can be fixed having something like

```scala
  implicit private val ordering: Ordering[Resolver] = Ordering.by[Resolver, Int](_.priority).orElseBy(_.ref)
```

Ideally we would prefer it ordered by label, but we cannot access it easily at that level...so UUID would do the trick.

An alternative solution to the problem is to only allow creation of resolvers with different priorities within the scope of a project.

pafonta:
Hello @umbreak,

Thank you for having documented this issue.

As a user, I would prefer the solution of "only allow creation of resolvers with different priorities within the scope of a project".

bogdanromanx:
We will prevent the creation of resolvers with the same priority in the same project.

umbreak:
When creating/updating a resolver, check for the previously existing resolvers priorities. If there is a resolver with the same priority, reject the request

PR #1979, opened 2021-01-29
Prevented multiple resolvers from having the same priority in the scope of a project

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1356


delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ResolversDummy.scala
```diff
@@ -155,6 +160,24 @@ class ResolversDummy private (
         _          <- cache.setToCache(res)
       } yield res
     }
+
+  private def verifyPriorityUniqueness(
```
imsdu, line 164: It belongs more to the state machine to do this check by providing a function to search resolvers.

imsdu: We also need to disable it for migration by checking if the env variable MIGRATE_DATA is defined

imsdu (changes requested):

wwajerowicz (approved):

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ResolversDummy.scala
umbreak: In the state machine we don't have access to the cache. Besides that, for consistency with the storages, this is done [in the Impl section](https://github.com/BlueBrain/nexus/blob/master/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala#L80)

The migration part is not tacked in this PR

imsdu: For Realms, it is handled in the state machine for example. And for others, we don't pass the cache but other modules and so on.
For the migration part, it can be handled elsewhere, we just need not to forget about it.

imsdu (dismissed):

PR #1990, opened 2021-02-02
Fixed clustered bootstrapping

bogdanromanx:
The actor system name must match the actor system name from the seed list.


imsdu (changes requested):
The MigrationModule is mostly a copy-paste from the DeltaModule so we need to change it also there

imsdu (approved):

PR #1993, opened 2021-02-03
Added support for keeping track of projects statistics

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/1984
Fixes https://github.com/BlueBrain/nexus/issues/1995

- Added support for storing an `A` in the `ProjectionProgress`
- Reverted to using a single `StreamSupervisor`. The storing of the state is handled explicitly through the stream and Distributed Data
- Added support for generating keyspace/tables on system boot if config properties are set.
- Added GH Actions to timeout at 20min (default is 360min)


delta/app/src/main/resources/app.conf
```diff
@@ -183,6 +183,14 @@ app {
     pagination = ${app.defaults.pagination}
     # the projects indexing config
     indexing = ${app.defaults.indexing}
+
+    # configuration for the persistence of progress of projections
+    persist-progress-config {
+      # the maximum number of items in a batch before persisting progress
+      max-batch-size = 5
+      # the maximum time allowed to pass between persisting the progress
+      max-time-window = 50 millis
```
bogdanromanx, line 192: the default values can be set to be higher... e.g. batch: 10, window: 5 seconds

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ProjectsStatistics.scala
```diff
@@ -0,0 +1,77 @@
...
+import monix.bio.Task
+import monix.execution.Scheduler
+import ch.epfl.bluebrain.nexus.delta.kernel.RetryStrategy.logError
+
+trait ProjectsStatistics {
+
+  /**
+    * Retrieve the current statistics in terms of number of existing events and the latest offset (per project)
+    */
+  def get(): Task[ProjectStatisticsCollection]
```
bogdanromanx, line 26: any reason for Task instead of UIO?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/Event.scala
```diff
@@ -23,6 +24,11 @@ trait Event extends Product with Serializable {
     * @return the subject that performed the action that resulted in emitting this event
     */
   def subject: Subject
+
+  /**
+    * @return Some(project) when an event belongs to a project, None otherwise
+    */
+  def belongsTo: Option[ProjectRef]
```
wwajerowicz, line 31: Maybe we could create a subtype of events like `ProjectBasedEvent` with just `def project: ProjectRef` rather than add this field to all the events, even those that don't belong to a project.

bogdanromanx (changes requested):

wwajerowicz (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/Event.scala
bogdanromanx: I'd prefer the method to the sub-type, honestly...

umbreak: I'm fine with this, but I didn't find a good name for it. `ProjectBasedEvent` might be alright...but it is not very convincing either.

delta/app/src/main/resources/app.conf
umbreak: That can be problematic. The value is stored on the state **after** it is persisted on the DB. The higher this value the bigger the delay from the actual stream.

We could find ways to fix this (either branching the stream and doing some complicated broadcast & merge) or indeed using Distributed Data

bogdanromanx: > The value is stored on the state after it is persisted on the DB.
The value should be added to the state for each element... and persisting to the DB should happen every now and again.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/Event.scala
bogdanromanx: how about `ProjectScopedEvent`?

delta/app/src/main/resources/app.conf
umbreak: I changed it. Now I'm using Distributed Data

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ProjectsStatistics.scala
```diff
@@ -0,0 +1,91 @@
...
+      KeyValueStore.distributed[ProjectRef, ProjectStatistics]("ProjectsStatistics", (_, stats) => stats.count)
+
+    def buildStream(
+        progress: ProjectionProgress[ProjectStatisticsCollection]
+    ): Stream[Task, ProjectStatisticsCollection] = {
+      val initial = SuccessMessage(progress.offset, "", 1, progress.value, Vector.empty)
+      stream(progress.offset)
+        .collect { case env @ Envelope(event: ProjectScopedEvent, _, _, _, _, _) => env.toMessage.as(event.project) }
+        .mapAccumulate(initial) { (acc, msg) => (msg.as(acc.value.incrementCount(msg.value, msg.offset)), msg.value) }
+        .evalMap { case (acc, projectRef) => cache.put(projectRef, acc.value.value(projectRef)).as(acc) }
```
imsdu, line 72: Before writing to the cache, we could debounce to put less pressure on it

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ProjectsStatistics.scala
umbreak: That will prevent from writing to every cache entry. A debounce will take the last `projectRef` and only that will be updated.

wwajerowicz (approved):

imsdu (approved):

bogdanromanx (approved):

PR #2028, opened 2021-02-12
Indexing coordinator

umbreak:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/ViewLens.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import ch.epfl.bluebrain.nexus.sourcing.projections.ProjectionId.ViewProjectionId
+
+import java.util.UUID
+
+/**
+  * A lens on a View
+  * @tparam A the view type
+  */
+trait ViewLens[A] {
```
imsdu, line 11: Should not we just have a trait defined in the sdk that would be needed to be implemented by the different views in the different plugins rather than having too defined a Lens ?
Plus we only need basic information here

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/ViewLens.scala
umbreak: I didn't want to couple those things _yet_.
Maybe we end up doing it this way, maybe not.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingState.scala
```diff
@@ -0,0 +1,10 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.indexing
+
+import ch.epfl.bluebrain.nexus.sourcing.projections.stream.StreamSupervisor
+
+sealed private[indexing] trait IndexingState extends Product with Serializable
+
+private[indexing] object IndexingState {
+  final case object Initial                                         extends IndexingState
+  final case class Current(rev: Long, supervisor: StreamSupervisor) extends IndexingState
```
imsdu, line 9: It may be another PR but it would be nice to know when it has been started for example

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingState.scala
```diff
@@ -0,0 +1,10 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.indexing
+
+import ch.epfl.bluebrain.nexus.sourcing.projections.stream.StreamSupervisor
+
+sealed private[indexing] trait IndexingState extends Product with Serializable
+
+private[indexing] object IndexingState {
+  final case object Initial                                         extends IndexingState
```
imsdu, line 8: We could have a different state for Stop to make a distinction between views that have never been started and the ones that ran at least once

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamCoordinator.scala
```diff
@@ -0,0 +1,74 @@
...
+  def apply[V: ViewLens](
+      entityType: String,
+      buildStream: BuildStream[V],
+      projection: Projection[Unit],
+      config: EventSourceProcessorConfig,
+      retryStrategy: RetryStrategy[Throwable]
+  )(implicit V: ClassTag[V], as: ActorSystem[Nothing], sc: Scheduler): Task[IndexingStreamCoordinator[V]] = {
+
+    def start(view: V): Task[IndexingState] = {
+      val stream = projection.progress(view.projectionId).flatMap(buildStream(view, _))
```
wwajerowicz, line 57: We also need an option to restart indexing the view from the beginning for the cases when we just want to reindex a view without updating it.

imsdu (changes requested):

wwajerowicz (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingState.scala
umbreak: This information is already present in the progress offset, which as per your request is handled separately in a distributed data cache

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingState.scala
umbreak: Again, this information is going to be obtained using the progress

bogdanromanx (dismissed):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamCoordinator.scala
```diff
@@ -0,0 +1,74 @@
...
+object IndexingStreamCoordinator {
+
+  type BuildStream[V] = (V, ProjectionProgress[Unit]) => Task[Stream[Task, Unit]]
+
+  private[indexing] type Agg = Aggregate[String, IndexingState, IndexingCommand, IndexingState, Throwable]
+
+  /**
+    * Construct a [[IndexingStreamCoordinator]] relying on the underlying transient [[ShardedAggregate]]
+    */
+  def apply[V: ViewLens](
```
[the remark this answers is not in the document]
umbreak: If we need extra work to use cluster sharding, we don't want to use it. The balancing benefits do not outperform the extra hussle.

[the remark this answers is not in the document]
bogdanromanx: all cluster singletons run on the oldest node; this affects the performance of the entire system... but let's discuss in #2031 if you're not confident this would be a good change.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingState.scala
imsdu: I meant when the current StreamSupervisor started, this part is not in progress offset.

wwajerowicz (dismissed):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingState.scala
umbreak: We don't care about when it started, we care about the progress.

If we ever care about when it started, I will add it

wwajerowicz (approved):

bogdanromanx (approved):

PR #2032, opened 2021-02-12
Introduced project initialization hook for core and plugins

### LINKED ISSUE #1907: Create a default in-project resolver after project creation

bogdanromanx:
The change also preconfigures projects with a default InProject
resolver with a fixed id.

Fixes #1907


delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/OrganizationsDummy.scala
```diff
@@ -32,15 +33,19 @@ final class OrganizationsDummy private (
     journal: OrganizationsJournal,
     cache: OrganizationsCache,
     semaphore: IOSemaphore,
-    applyOwnerPermissions: ApplyOwnerPermissionsDummy
+    scopeInitializations: Set[ScopeInitialization]
```
imsdu, line 36: Organizations know too much, it should not know that there are ScopeInitialization behind the scene or decide if it should run in parallel or not.
From its point of view it is just a function that takes the fresh org and return an `IO[OrganizationInitializationFailed, Unit]` (and it should make testing easier).
Maybe this how we run them should be delegated to a ScopeInitializationRunner or something ?

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ProjectsDummy.scala
```diff
@@ -30,23 +31,28 @@ final class ProjectsDummy private (
...
+                    CreateProject(
+                      ref,
+                      fields.description,
+                      fields.apiMappings,
+                      fields.baseOrGenerated(ref),
+                      fields.vocabOrGenerated(ref),
+                      caller
+                    )
+                  )
+      _        <- IO.parTraverseUnordered(scopeInitializations)(_.onProjectCreation(resource.value, caller))
```
imsdu, line 52: Same thing like with organizations, the hook should be simpler from there

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/organizations/OrganizationsImpl.scala
```diff
@@ -138,16 +136,29 @@ object OrganizationsImpl {
...
+            orgs
+              .fetch(envelope.event.label)
+              .redeemCauseWith(
+                _ => IO.unit,
+                { resource =>
+                  index.put(resource.value.label, resource) >>
+                    IO.when(!resource.deprecated) {
+                      IO.parTraverseUnordered(si)(_.onOrganizationCreation(resource.value, resource.createdBy))
+                        .void
+                        .redeemCause(_ => (), identity)
```
imsdu, line 157: We should have some logs somewhere to know when it fails and keep failing

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/organizations/OrganizationsImpl.scala
```diff
@@ -138,16 +136,29 @@ object OrganizationsImpl {
...
-          )
+          .mapAsync(config.indexing.concurrency) { envelope =>
+            orgs
+              .fetch(envelope.event.label)
+              .redeemCauseWith(
+                _ => IO.unit,
+                { resource =>
+                  index.put(resource.value.label, resource) >>
+                    IO.when(!resource.deprecated) {
+                      IO.parTraverseUnordered(si)(_.onOrganizationCreation(resource.value, resource.createdBy))
```
imsdu, line 155: We should only try this when getting a creation event.
And like in the synchronous bit, I feel like we should to hide more details from this part

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/OwnerPermissionsDummy.scala
```diff
@@ -0,0 +1,55 @@
...
+import monix.bio.IO
+
+/**
+  * Apply the owner permissions at org/project creation
+  *
+  * @param acls               an instance of the acl module
+  * @param ownerPermissions   the owner permissions to be present at project creation
+  * @param serviceAccount     the service account to apply owner permissions when needed
+  */
+class OwnerPermissionsDummy(acls: Acls, ownerPermissions: Set[Permission], serviceAccount: ServiceAccount)
```
imsdu, line 20: As it takes an acl (which can be a dummy), we should not need one for applying owner permissions itself no ?

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/utils/OwnerPermissionsScopeInitialization.scala
```diff
@@ -0,0 +1,77 @@
...
+        .fetch(AclAddress.Organization(organization.label))
+        .void // discard value as the check is only done to see if there have been ACLs already set on this path
+        .onErrorHandleWith { _: AclRejection.AclNotFound =>
+          val str = s"Owner permissions for organization '${organization.label}' have not been set, applying them now."
+          UIO.delay(logger.info(str)) >>
+            acls
+              .append(Acl(AclAddress.Organization(organization.label), subject -> ownerPermissions), 0L)
+              .void // discard return value
+              .onErrorHandleWith {
+                case _: AclRejection.IncorrectRev => // concurrent update to the acls of the organization, assuming a success
```
imsdu, line 45: The rejection NothingToBeUpdated may also happen

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/utils/OwnerPermissionsScopeInitialization.scala
```diff
@@ -0,0 +1,77 @@
...
+        .fetch(AclAddress.Project(project.ref))
+        .void // discard value as the check is only done to see if there have been ACLs already set on this path
+        .onErrorHandleWith { _: AclRejection.AclNotFound =>
+          val str = s"Owner permissions for project '${project.ref}' have not been set, applying them now."
+          UIO.delay(logger.info(str)) >>
+            acls
+              .append(Acl(AclAddress.Project(project.ref), subject -> ownerPermissions), 0L)
+              .void // discard return value
+              .onErrorHandleWith {
+                case _: AclRejection.IncorrectRev => // concurrent update to the acls of the project, assuming a success
```
imsdu, line 68: Same

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/utils/OwnerPermissionsScopeInitialization.scala
```diff
@@ -0,0 +1,77 @@
...
+    extends ScopeInitialization {
+
+  private val logger: Logger          = Logger[OwnerPermissionsScopeInitialization]
+  implicit private val caller: Caller = serviceAccount.caller
+
+  override def onOrganizationCreation(
+      organization: Organization,
+      subject: Subject
+  ): IO[ScopeInitializationFailed, Unit] =
+    if (MigrationState.isRunning) IO.unit
```
imsdu, line 33: If we had this `ScopeInitializationRunner` (or whatever it may be called), we could also disable this operations in a single point

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/utils/ResolverScopeInitialization.scala
```diff
@@ -0,0 +1,64 @@
...
+  override def onProjectCreation(project: Project, subject: Subject): IO[ScopeInitializationFailed, Unit] = {
+    if (MigrationState.isRunning) UIO.unit
+    else
+      resolvers
+        .fetch(defaultInProjectResolverId, project.ref)
+        .void // discard value
+        .onErrorHandleWith {
+          case _: ResolverRejection.ResolverNotFound =>
+            // if the default resolver is not found attempt to create it or fail with ScopeInitializationFailed
+            logger.info(s"Default InProject resolver not found for project ${project.ref}, creating it now.")
```
imsdu, line 41: A warn level would be best

imsdu (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/utils/ResolverScopeInitialization.scala
bogdanromanx: This also happens to be the happy path when the project gets created; a level of WARN seems to much.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/utils/OwnerPermissionsScopeInitialization.scala
bogdanromanx: and append for 0L revision cannot end in NothingToBeUpdated.

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/OwnerPermissionsDummy.scala
bogdanromanx: I'm not sure I understand your comment...

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/organizations/OrganizationsImpl.scala
bogdanromanx: Each ScopeIntialization implementation records these logs.

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/OwnerPermissionsDummy.scala
imsdu: `OwnerPermissionsScopeInitialization` and `OwnerPermissionsDummy` have the same dependencies, we should not need two implementations

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/OrganizationsDummy.scala
bogdanromanx: I think introducing a Runner makes things more complicated... I don't see any benefit for adding an arbitrary function that gets invoked. The behaviour is explicit: organizations and projects have an initialization hook allows for arbitrary behaviour to be registered as per construction. Unless we would like to prioritize the execution of certain initialization behaviours (no usecase for that so far), I see the abstraction adding some unnecessary complexity.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/organizations/OrganizationsImpl.scala
bogdanromanx: good point, we should only execute for *Created events.

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/OrganizationsDummy.scala
imsdu: To ensure we run it the same way in synchronous mode and healing mode and to make tests easier without having to worry what a ScopeInitialization is ?

bogdanromanx: Given the fact that there are only 2 places where this code is used, the customization of the tracing span and aligning errors I think it would be too much to generalize/abstract.

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/OwnerPermissionsDummy.scala
bogdanromanx: it's a module dependency problem... we need two implementations in order to maintain the dummy acl behaviour without bringing more deps on the sdkTestkit.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsImpl.scala
```diff
@@ -160,20 +160,35 @@ object ProjectsImpl {
...
-            projects.fetch(envelope.event.project).redeemCauseWith(_ => IO.unit, res => index.put(res.value.ref, res))
-          )
+          .mapAsync(config.indexing.concurrency) { envelope =>
+            projects
+              .fetch(envelope.event.project)
+              .redeemCauseWith(
+                _ => IO.unit,
+                { resource =>
+                  index.put(resource.value.ref, resource) >>
+                    IO.when(!resource.deprecated && envelope.event.isInstanceOf[ProjectCreated]) {
```
umbreak, line 183: Why the two checks? it is enough with the second. Although it is pretty ugly. I would rather introduce a method on `ProjectEvent`

```scala
def isCreated: Boolean  =
self match {
   case _: ProjectCreated => true
   case _ => false
```

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsImpl.scala
```diff
@@ -160,20 +160,35 @@ object ProjectsImpl {
...
+                      IO
+                        .parTraverseUnordered(si)(_.onProjectCreation(resource.value, resource.createdBy))
+                        .void
+                        .redeemCause(_ => (), identity)
```
umbreak, line 187: `.onErrorFallbackTo(IO.unit)`

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/utils/OwnerPermissionsScopeInitialization.scala
umbreak: Again migrations code creeping in production code....

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/utils/ResolverScopeInitialization.scala
umbreak: 1. Why you try to fetch it and then create it? Just create it and check if it fails with already exists or incorrect revision and ignore those errors
2. We don't need any logging there, it is just an expected an normal behaviour. We never log those kinds of things.

umbreak (changes requested):
It is missing the same thing for storages

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/utils/ResolverScopeInitialization.scala
imsdu: It may fail also if meanwhile, another resolver has been created with priority 1 :)

umbreak: Yeah, So ignore all those errors

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsImpl.scala
```diff
@@ -160,20 +159,35 @@ object ProjectsImpl {
     KeyValueStore.distributed(moduleType, clock)
   }
 
+  @SuppressWarnings(Array("IsInstanceOf"))
```
umbreak, line 162: THis is not needed anymore

umbreak (approved):

umbreak (dismissed):

PR #2037, opened 2021-02-15
Add a LRU cache for schemas to improve resource validation performance

### LINKED ISSUE #1533: Implement caching for schema models and shacl engines to speed up resource validation

### LINKED ISSUE #2017: Perform validation of a resource inside the Schema state actor
We have to change the `EventSourceProcessor` to allow for execution of arbitrary `IO` on the current state. 

This is very important in the case of Schemas in order to avoid serializing the `Schema` instance. This serialization is expensive because the flatten schema Model is big and the validation triples added to the shape are many (almost 4000 triples).

Once this is added, validation will be performed on the Schema state actor instead of the resource state actor.

imsdu:
Fixes #1533


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/cache/KeyValueStore.scala
```diff
@@ -103,8 +106,21 @@ trait KeyValueStore[K, V] {
...
-  def get(key: K): UIO[Option[V]] =
-    entries.map(_.get(key))
+  def get(key: K): UIO[Option[V]]
+
+  /**
+    * @param key the key
+    * @param op the computation yielding the value to associate with `key`, if
+    *           `key` is previously unbound.
+    */
+  def getOrElseUpdate[E](key: K, op: => IO[E, V]): IO[E, V] =
```
umbreak, line 116: this operation is not atomic. We should at least mention that in the docs

delta/service/src/main/resources/service.conf
```diff
@@ -6,11 +6,11 @@ service {
     "org.apache.jena.rdf.model.Model"                                                      = "kryo"
     "ch.epfl.bluebrain.nexus.delta.rdf.graph.Graph"                                        = "kryo"
     "ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode$Iri"                                     = "kryo"
-    "ch.epfl.bluebrain.nexus.delta.rdf.shacl.ShaclShapesGraph"                             = "kryo"
     "ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceF"                                    = "kryo"
     "ch.epfl.bluebrain.nexus.delta.sdk.model.acls.AclAddress"                              = "kryo"
     "ch.epfl.bluebrain.nexus.delta.sdk.model.Label"                                        = "kryo"
     "ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectRef"                          = "kryo"
+    "ch.epfl.bluebrain.nexus.delta.sdk.model.schemas.Schema"                               = "kryo"
```
umbreak, line 13: Can you also add
```
    "ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectStatisticsCollection$ProjectStatistics" = "kryo"
```

umbreak (changes requested):

umbreak (approved):

PR #2046, opened 2021-02-17
Fix module definition and config reader

imsdu:


wwajerowicz (dismissed):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePluginModule.scala
```diff
@@ -125,7 +126,9 @@ object StoragePluginModule extends ModuleDef {
     )
   }
 
-  make[StorageScopeInitialization]
+  make[StorageScopeInitialization].from { (storages: Storages, serviceAccount: ServiceAccount) =>
```
umbreak, line 129: This is probably not needed if it is just used on `ScopeInitialization`. You can very well just have:
```scala
many[ScopeInitialization].add{...}
```

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/MigrationModule.scala
```diff
@@ -76,6 +86,42 @@ class MigrationModule(appCfg: AppConfig, config: Config)(implicit classLoader: C
...
+      val projection                            = appCfg.database.flavour match {
+        case Postgres  => Projection.postgres(appCfg.database.postgres, ProjectStatisticsCollection.empty)
+        case Cassandra => Projection.cassandra(appCfg.database.cassandra, ProjectStatisticsCollection.empty)
+      }
+      projection.flatMap { p =>
+        ProjectsStatistics(appCfg.projects, p, eventLog.eventsByTag(Event.eventTag, _))
+      }
+  }
+
+  make[ResolverScopeInitialization].from { (resolvers: Resolvers, serviceAccount: ServiceAccount) =>
```
umbreak, line 113: same here. If this is just used to power `ScopeInitialization` it can be done directly there with the .add clause

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/MigrationModule.scala
```diff
@@ -76,6 +86,42 @@ class MigrationModule(appCfg: AppConfig, config: Config)(implicit classLoader: C
...
+    new ResolverScopeInitialization(resolvers, serviceAccount)
+  }
+
+  make[OwnerPermissionsScopeInitialization].from { (acls: Acls, appCfg: AppConfig, serviceAccount: ServiceAccount) =>
```
umbreak, line 117: same

umbreak (changes requested):

umbreak (commented):
I think it is fine. You can change what I suggested but it is not fully necessary either

umbreak (approved):

PR #2049, opened 2021-02-17
Allow caching of projection progress.

wwajerowicz:


delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphIndexingCoordinator.scala
```diff
@@ -84,6 +86,7 @@ private class IndexingStream(
   )(implicit sc: Scheduler): IO[Nothing, Stream[Task, Unit]] =
     for {
       _     <- client.createNamespace(namespace, indexProperties).hideErrorsWith(illegalArgument)
+      _     <- cache.remove(projectionId)
```
umbreak, line 89: wouldn't it be more like `cache.put(projectionId, initialProgress)` ?

umbreak (commented):
Needs to be done for elasticsearch too

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphIndexingCoordinator.scala
wwajerowicz: cache.put(projectionId, initialProgress) would only work if there's no value in the cache yet, otherwise when we restart the view, the initialProgress will be earlier than the current and the update will be ignored by the KeyValueStore

umbreak: then:
```
   _     <- cache.remove(projectionId)
   _     <- cache.put(projectionId, initialProgress)
```

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/projections/ProjectionStream.scala
```diff
@@ -138,6 +138,56 @@ object ProjectionStream {
...
+        initial: ProjectionProgress[A],
+        persistProgress: (ProjectionId, ProjectionProgress[A]) => Task[Unit],
+        persistErrors: (ProjectionId, Vector[Message[A]]) => Task[Unit],
+        cacheProgress: (ProjectionId, ProjectionProgress[A]) => Task[Unit],
+        config: PersistProgressConfig
+    ): Stream[Task, A] =
+      stream
+        .accumulateProgress(initial)
+        .evalMap { case (progress, message) =>
+          cacheProgress(projectionId, progress).as((progress, message))
```
imsdu, line 186: That makes a lot of writes in the cache and Distributed Data is eventually consistent, we really need to persist for every event processed by every view ?

imsdu (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphIndexingCoordinator.scala
wwajerowicz: I think 
```scala
      _     <- cache.remove(projectionId)
      _     <- cache.put(projectionId, initialProgress)
```
would make most sense.

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/projections/ProjectionStream.scala
bogdanromanx: this would be a problem indeed; if we need more consistency we would have to switch to the previous way to store the value in memory.

wwajerowicz: If we don't want to persist every message,  we would have to chunk and flatten the stream one more time.
We could also cache the progress in the same place where we record it in the projection , but then the delay would be the same as for the projection.

umbreak: If the progress has already been computed, we can use `debounce(5.seconds)` or something before `.accumulateProgress(initial)`

wwajerowicz: if we use `Ref[Task, A] wouldn't  it we have to run the stream (and indexing) for each view on each node?

umbreak: Yes. That's the penalty to pay for a faster cache.

wwajerowicz: I think `debounce` drops elements, which would mean that some messages are dropped and, in case of errors and warnings, not persisted.

umbreak: otherwise you can do another `groupWithin`, take the latest from the cunks and index it into the cache and then flatten the chunks back into the stream

umbreak (changes requested):

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/projections/ProjectionStream.scala
wwajerowicz: yes, I guess that will be more efficient than writing to cache for every message.

bogdanromanx (approved):

umbreak (approved):

PR #2060, opened 2021-02-19
Archive plugin STM

### LINKED ISSUE #1496: Archive Plugin: Define STM

bogdanromanx:
Fixes #1496

### DIFF
--- a/delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
+++ b/delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
@@ -1,0 +1,77 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.archive.model
+
+import cats.Order
+import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceRef
+import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectRef
+
+import java.nio.file.Path
+
+/**
+  * Enumeration of archive references.
+  */
+sealed trait ArchiveReference extends Product with Serializable {
+
+  /**
+    * @return the referenced resource id optionally qualified with a tag or a revision
+    */
+  def ref: ResourceRef
+
+  /**
+    * @return the parent project of the referenced resource
+    */
+  def project: ProjectRef
+
+  /**
+    * @return the target location in the archive
+    */
+  def path: Path
+
+  /**
+    * @return the archive reference type
+    */
+  def tpe: ArchiveReferenceType
+}
+
+object ArchiveReference {
+
+  /**
+    * An archive resource reference.
+    *
+    * @param ref            the referenced resource id
+    * @param project        the parent project of the referenced resource
+    * @param path           the target location in the archive
+    * @param representation the format in which the resource should be represented
+    */
+  final case class ResourceReference(
+      ref: ResourceRef,
+      project: ProjectRef,
+      path: Path,
+      representation: ArchiveResourceRepresentation
+  ) extends ArchiveReference {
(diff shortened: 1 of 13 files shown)


delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+    * An archive resource reference.
+    *
+    * @param ref            the referenced resource id
+    * @param project        the parent project of the referenced resource
+    * @param path           the target location in the archive
+    * @param representation the format in which the resource should be represented
+    */
+  final case class ResourceReference(
```
umbreak, line 45: isn't it missing the field [originalSource](https://bluebrainnexus.io/docs/delta/api/current/kg-archives-api.html#archive-format) ?

umbreak (changes requested):

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
bogdanromanx: The resource reference has a representation field that can be: source / expanded / compacted / ntriples or dot.
All are mutually exclusive so this makes for a better model.

Same thing with the ResourceRef field: it can be Latest / Tag / Rev...

bogdanromanx: The wire representation will include that field and possibly a new one like `format`.

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
```diff
@@ -0,0 +1,77 @@
...
+import cats.Order
+import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceRef
+import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectRef
+
+import java.nio.file.Path
+
+/**
+  * Enumeration of archive references.
+  */
+sealed trait ArchiveReference extends Product with Serializable {
```
imsdu, line 12: I thought Archives could contain every possible resource out there including views ?

bogdanromanx: that's correct... what creates this confusion?

imsdu: I understood ResourceReference for data resources only

bogdanromanx: I don't know how to name it better to convey that all resource types are included: schemas, resolvers, storages, views etc.

imsdu: Maybe documenting a little more ?

bogdanromanx: yeah, but can I do that in a different PR to avoid a full re-build?

imsdu: No problem

umbreak (approved):

PR #2075, opened 2021-02-23
Added view routes (part 1)

umbreak:
TODO:
- statistics routes
- progress routes
- listings for other resource types routes
- listings for all views routes
- testing of query routes (_search)
- SSEs routes


delta/sdk/src/main/resources/contexts/metadata.json
```diff
@@ -62,6 +62,9 @@
     "Storage": "https://bluebrain.github.io/nexus/vocabulary/Storage",
     "DiskStorage": "https://bluebrain.github.io/nexus/vocabulary/DiskStorage",
     "S3Storage": "https://bluebrain.github.io/nexus/vocabulary/S3Storage",
-    "RemoteDiskStorage": "https://bluebrain.github.io/nexus/vocabulary/RemoteDiskStorage"
+    "RemoteDiskStorage": "https://bluebrain.github.io/nexus/vocabulary/RemoteDiskStorage",
```
wwajerowicz, line 65: I think at some point we need to find a way to split the metadata context between plugins and core. Context in `sdk` shouldn't know about types from plugins.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/RawResult.scala
```diff
@@ -0,0 +1,49 @@
...
+
+    override def graph(
+        value: RawResult
+    )(implicit opts: JsonLdOptions, api: JsonLdApi, rcr: RemoteContextResolution): IO[RdfError, Graph] =
+      IO.raiseError(ConversionError("An ElasticSearch response cannot be converted to Graph", "to graph"))
+
+    override def compact(
+        value: RawResult
+    )(implicit opts: JsonLdOptions, api: JsonLdApi, rcr: RemoteContextResolution): IO[RdfError, CompactedJsonLd] =
+      IO.pure(CompactedJsonLd.unsafe(BNode.random, ContextValue.empty, value.value))
```
wwajerowicz, line 47: ES query results are not JSON-LD and I don't think we should treat them as such. I don't really see the need for this `JsonLdEncoder`.

wwajerowicz (changes requested):

delta/sdk/src/main/resources/contexts/metadata.json
umbreak: Yes. That can be done having in the pluginDef some extraMetadata whatever that gets merged with the original metadata. Let's do that later though

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/RawResult.scala
umbreak: This `JsonLdEncoder` odes exactly that, return pure json without any modifications. And as you can see it won't work with expanded nor other weird things.

Why is there? Because it is much more convenient to keep using the current `emit` which will deal with the error channel and the content negotiation without having to code again for it

delta/sdk/src/main/resources/contexts/metadata.json
wwajerowicz: Yes, that's something for later.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/RawResult.scala
wwajerowicz: Won't it return an exception when someone will try to request expanded JSON-LD or other of those formats and interfere with regular content negotiation?

umbreak: If you try to use this endpoint having Accept: `application/n-triples` it will trigger the exception, yes. However this exception says quite clearly that `An ElasticSearch response cannot be converted to N-Triples` so I don't find it wrong.

If you thing that's bad I can add some special case to handle `Json` on the `emit`. But it seems an unnecessary overhead to me

wwajerowicz: We can do it this way, but we just have to make sure that we return `406 Not Acceptable` when someone requests n-triples etc.

umbreak: I've handled it the proper way, thought the `DeltaDirectives`

wwajerowicz (approved):

bogdanromanx (approved):

Issue #2077, opened 2021-02-23
Add linking operation for S3Storage


imsdu:

imsdu:
Fixed with #2078

Issue #2062, opened 2021-02-19
Create a S3 storage should fail if the bucket does not exist


imsdu:
The problem comes from the integration tests as now, checks are not performed in the same order as before.

imsdu:
Fixed with #2078

Issue #2061, opened 2021-02-19
Creating a local storage should fail if the volume does not exist


imsdu:
The problem comes from the integration tests as now, checks are not performed in the same order as before.

imsdu:
Fixed with #2078

PR #2078, opened 2021-02-23
Fix storage bugs and make storage integration tests run

### LINKED ISSUE #2056: When attempting to fetch file metadata, the file itself is returned not the metadata

imsdu:
Fixes #2056 #2061 #2062 #2077


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3StorageLinkFile.scala
```diff
@@ -0,0 +1,55 @@
...
+
+class S3StorageLinkFile(storage: S3Storage)(implicit as: ActorSystem) extends LinkFile {
+
+  import as.dispatcher
+
+  private val fileNotFoundException = new IllegalArgumentException("File not found")
+
+  override def apply(key: Uri.Path, description: FileDescription): IO[StorageFileRejection, FileAttributes] = {
+    val attributes    = S3Attributes.settings(storage.value.toAlpakkaSettings)
+    val location: Uri = s"${storage.value.address(storage.value.bucket)}/$key"
```
umbreak, line 27: why is used the string conversion here and not the actual Uri conversion?
```scala
import ch.epfl.bluebrain.nexus.delta.rdf.syntax.sdk.syntax._

val location: Uri = storage.value.address(storage.value.bucket) / key
```

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/routes/StoragesRoutes.scala
```diff
@@ -259,7 +259,7 @@ object StoragesRoutes {
       case IncorrectRev(_, _)                => StatusCodes.Conflict
       case WrappedProjectRejection(rej)      => rej.status
       case WrappedOrganizationRejection(rej) => rej.status
-      case StorageNotAccessible(_, _)        => StatusCodes.Forbidden
+      case StorageNotAccessible(_, _)        => StatusCodes.BadRequest
```
umbreak, line 262: Why? If it is not accessible it is as if the client wouldn't have permissions

umbreak (changes requested):

bogdanromanx (dismissed):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3StorageLinkFile.scala
imsdu: I forgot about it

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/routes/StoragesRoutes.scala
imsdu: `StorageNotAccessible` is raised when volumes / bucket are invalid no ? It is not another rejection when there are no permissions ?

umbreak: Now it seems to me more like a NotFound.
But yeah, bad request is fine I guess

umbreak (approved):

PR #2087, opened 2021-02-25
Archive creation / fetch logic

bogdanromanx:


delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
```diff
@@ -1,24 +1,195 @@
...
+
+  // TODO: implement download function
+  //def download(id: IdSegment, projectRef: ProjectRef): IO[ArchiveRejection, Source[ByteString, Any]] = ???
+
+  private def eval(cmd: CreateArchive, project: Project): IO[ArchiveRejection, ArchiveResource] =
+    (for {
+      result    <- aggregate.evaluate(identifier(cmd.project, cmd.id), cmd).mapError(_.value)
+      (am, base) = project.apiMappings -> project.base
+      resource  <- IO.fromOption(result.state.toResource(am, base, cfg.ttl), UnexpectedInitialState(cmd.id, project.ref))
+    } yield resource).named("evaluateArchiveCommand", moduleType)
```
umbreak, line 136: this span is interesting, but it is not present in any other module. We should be consistent so either remove it add it everywhere else

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
```diff
@@ -28,19 +199,19 @@ object Archives {
...
       else IO.unit
     }
 
     def checkAbsolutePaths: IO[ArchiveRejection, Unit] = {
-      val nonAbsolute = command.resources.value.foldLeft(Set.empty[Path]) { case (set, elem) =>
-        if (!elem.path.isAbsolute) set + elem.path else set
+      val nonAbsolute = command.value.resources.value.map(_.path).collect {
+        case Some(value) if !value.isAbsolute => value
       }
       if (nonAbsolute.nonEmpty) IO.raiseError(PathIsNotAbsolute(nonAbsolute))
```
umbreak, line 216: This is not necessary since it is already present in JsonLdDecoder

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
```diff
@@ -74,4 +79,57 @@ object ArchiveReference {
       ref.original.toString + project.toString + path.toString + representation.toString
   }
 
+  implicit final val archiveReferenceJsonLdDecoder: JsonLdDecoder[ArchiveReference] = {
+    // custom decoder for resource ref to rev and tag references from separate fields
+    implicit val resourceRefDecoder: JsonLdDecoder[ResourceRef] =
+      (cursor: ExpandedJsonLdCursor) =>
+        for {
+          iri    <- cursor.downField(nxv + "resourceId").get[Iri]
```
umbreak, line 87: This should probably go in the companion object of `ResourceRef `

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
```diff
@@ -74,4 +79,57 @@ object ArchiveReference {
...
+          revOpt <- cursor.downField(nxv + "rev").get[Option[Long]]
+          tagOpt <- cursor.downField(nxv + "tag").get[Option[TagLabel]]
+          value  <- (revOpt, tagOpt) match {
+                      case (None, None)      => Right(ResourceRef.Latest(iri))
+                      case (Some(rev), None) => Right(ResourceRef.Revision(iri, rev))
+                      case (None, Some(tag)) => Right(ResourceRef.Tag(iri, iri, tag))
+                      case _                 =>
+                        Left(ParsingFailure("An archive resource reference cannot use both 'rev' and 'tag' fields."))
```
umbreak, line 95: A `ResourceRef` cannot use both 'rev' and 'tag' fields

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
```diff
@@ -74,4 +79,57 @@ object ArchiveReference {
...
+    implicit val absolutePathDecoder: JsonLdDecoder[Path] = {
+      val pathDecoder: JsonLdDecoder[Path] = (cursor: ExpandedJsonLdCursor) =>
+        cursor.getValueTry[Path](str => Paths.get(str))
+      pathDecoder.flatMap { path =>
+        if (path.isAbsolute) Right(path)
+        else Left(ParsingFailure(s"The archive reference path '$path' is not absolute."))
+      }
+    }
+
+    val resourceReferenceDecoder: JsonLdDecoder[ResourceReference] =
```
umbreak, line 109: This can be done automatically using `deriveConfigJsonLdDecoder` as for storages or elasticsearch views. it is a trait with different implementations by a discriminator

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveState.scala
```diff
@@ -1,38 +1,83 @@
...
+        ContextValue.empty,
+        base = Some(base.iri),
+        prefixMappings = allMappings.prefixMappings,
+        aliases = allMappings.aliases
+      )
+
+      val relative          = Uri("archives") / project.organization.value / project.project.value
+      val relativeAccess    = relative / id.toString
+      val relativeShortForm = relative / ctx.compact(id, useVocab = false)
+      val resourceUris      = RootResourceUris(relativeAccess, relativeShortForm)
```
umbreak, line 64: Why don't you put this on its own `ArchiveResourceUris`, so it can be tested separately

umbreak (changes requested):

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
bogdanromanx: the evaluate command in sourcing already has it, so I'll remove it

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
bogdanromanx: I prefer to have the check also here if it's already implemented and tested; if the archive will be used in a different way in the future it will not crash unexpectedly.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ReferenceExchange.scala
```diff
@@ -0,0 +1,53 @@
...
+import io.circe.Json
+import monix.bio.{IO, UIO}
+
+/**
+  * Contract definition for registering and consuming the ability to retrieve resources in a common JSON-LD format.
+  * Plugins can provide implementations for custom resource types such that those resources can be handled in an
+  * uniform way. Examples of use-cases would be: the ability to register resources defined by plugins (e.g. views) to
+  * be included in archives or being able to implement a generic `GET /resources/{org}/{proj}/_/{id}` endpoint.
+  */
+trait ReferenceExchange {
```
umbreak, line 17: We will also need some other class like `ReferenceExchangeCollection`

umbreak (changes requested):

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
bogdanromanx: it can't because it's specific to archives, it needs to deserialize variants like:
```
{
  "resourceId": "http://localhost/id"
}
```
```
{
  "resourceId": "http://localhost/id",
  "tag": "mytag"
}
```
```
{
  "resourceId": "http://localhost/id",
  "rev": 1
}
```
into a `ResourceRef`.

`ResourceRef` already has a decoder that follows the following formats:
```
"http://localhost/id"
"http://localhost/id?rev=1"
"http://localhost/id?tag=mytag"
```

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
bogdanromanx: Again, the message is specific to archives because of the custom encoding. A user of archive would have no knowledge of `ResourceRef`.

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
bogdanromanx: because of the custom decoding of `resourceRef` derivation cannot be used here; that's what I tried first.

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveState.scala
bogdanromanx: yep, makes sense

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
umbreak: The resourceRef decoding is already provided, so what's the problem?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ReferenceExchange.scala
bogdanromanx: why? my expectation is that we'll have many instances of this class bound with many, like:
```
make[ResourceReferenceExchange]
make[SchemaReferenceExchange]
make[ResolverReferenceExchange]
...
many[ReferenceExchange]
  .ref[ResourceReferenceExchange]
  .ref[SchemaReferenceExchange]
  .ref[ResolverReferenceExchange]
  ...
```

umbreak: ?? when you use the `_`, you don't know what type of resource it is. So yes, you need to try them all, that's why you need a collection.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ReferenceExchange.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+    * The resource value type.
+    */
+  type A
+
```
umbreak, line 23: probably we also want 

```scala
def apply(project: ProjectRef, schema: ResourceRef, reference: ResourceRef): UIO[Option[ReferenceExchangeValue[A]]]
```

Since we have to support: `v1/resources/{org}/{proj}/{schema}/{id}`

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ReferenceExchange.scala
bogdanromanx: it's an implementation concern; the download functionality would be realised with something like this:
```
class ArchiveDownload(exchanges: Set[ReferenceExchange]) {
  def apply(value: ArchiveValue): IO[ArchiveRejection, Source[ByteString, Any]] = ???
}
```
... where the instance would be wired automatically like the following (the set of values will be passed automatically during construction):
```
make[ArchiveDownload]
```

bogdanromanx: Adding another class would not be useful.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ReferenceExchange.scala
bogdanromanx: Interesting... we can add that, but I'd prefer to do it in a separate PR when working on that specific item.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ReferenceExchange.scala
umbreak: `ReferenceExchangeCollection` can be implemented later if we want. I'm just saying is gonna be needed in several places so it is going to be necessary to implement it.

`ArchiveDownload` is not the same as fetching the resource...

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
umbreak: The right way to do it would be to provide a proper constructor for `ResourceReference` and `FileReference` and keep to default ones private. Those constructors should check for the passed `Path` to be absolute and fail otherwise.

It makes sense to fail early during `ArchiveReference` construction instead of to have to fail on the state machine.

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
bogdanromanx: if you have the following json:
```
{
  "@type": "Resource"
  "resourceId": "http://localhost/id",
  "rev": 1,
  "path": "/id.json"
}
```
you have to decode the same json twice:
1. to get the `ResourceRef` from the root
2. to get the `ResourceReference` from the root

The generic derivation would expect something like the following to work:
```
{
  "@type": "Resource"
  "ref": {
    "resourceId": "http://localhost/id",
    "rev": 1
  },
  "path": "/id.json"
}
```

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
bogdanromanx: That's probably better and achieves the same goal...

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
umbreak: that's why we usually have a different case class to deserialize Json-LD than to actually use it for command/events(like in elasticsearch), it would probably be more readable.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ReferenceExchange.scala
imsdu: We can't have a single `ResourceExchange` for both `EventExchange` an `ReferenceExchange` ?
The main difference with `EventExchange` is that contains an extra step where we extract the reference from the event no ?

umbreak: Once the `ResourceF[A]` is fetched, in both cases you have an `StateExchange`. That part should definitely be reused.

Both have to implement:
```scala
      fetchLatest: F => IO[Rejection, ResourceF[S]],
      fetchByTag: (F, TagLabel) => IO[Rejection, ResourceF[S]],
```
 where F is `ResourceRef` in the case of `ReferenceExchange` and `<: Event` in the case of `EventExchange`. So it seems kind of possible to me too

wwajerowicz (dismissed):

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
umbreak: Actually the best way would be to have this in the sdk:
```scala
final case class AbsolutePath private (value: Path)
object AbsolutePath {
  def apply(path: Path): Either[FormatError, AbsolutePath] = ???
  implicit val absolutePathEncoder: Encoder[AbsolutePath] = ???
  implicit val absolutePathJsonLdEncoder: JsonLdEncoder[AbsolutePath] = ???
}
```

Since then you can use it in your `ResourceReference` and I can use it in `DiskStorageFields`.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ReferenceExchange.scala
bogdanromanx: I think conflating these two things is a bit premature at this stage and I don't see any real benefit as it only adds complexity. I will reconsider this after the archive plugin.

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
```diff
@@ -56,21 +62,65 @@ object ArchiveReference {
...
+
+    /**
+      * Safely constructs a resource reference that checks that if a path is provided it will be absolute.
+      *
+      * @param ref            the referenced resource id
+      * @param project        the parent project of the referenced resource
+      * @param path           the target location in the archive
+      * @param representation the format in which the resource should be represented
+      */
+    def apply(
```
umbreak, line 74: we don't need this if you use an `AbsolutePath` instead

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReference.scala
```diff
@@ -56,21 +62,65 @@ object ArchiveReference {
...
+  object FileReference {
+
+    /**
+      * Safely constructs a file reference that checks that if a path is provided it will be absolute.
+      *
+      * @param ref            the referenced resource id
+      * @param project        the parent project of the referenced resource
+      * @param path           the target location in the archive
+      */
+    def apply(
```
umbreak, line 111: same

umbreak (changes requested):
I'd recommend creating the `AbsolutePath`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ReferenceExchange.scala
```diff
@@ -0,0 +1,68 @@
...
+import io.circe.Json
+import monix.bio.{IO, UIO}
+
+/**
+  * Contract definition for registering and consuming the ability to retrieve resources in a common JSON-LD format.
+  * Plugins can provide implementations for custom resource types such that those resources can be handled in an
+  * uniform way. Examples of use-cases would be: the ability to register resources defined by plugins (e.g. views) to
+  * be included in archives or being able to implement a generic `GET /resources/{org}/{proj}/_/{id}` endpoint.
+  */
+trait ReferenceExchange {
```
umbreak, line 18: This is still wrong. You haven't merge this code with `EventExchange` as @imsdu and me pointed out

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ReferenceExchange.scala
bogdanromanx: This will be handled later.

umbreak (approved):

PR #2088, opened 2021-02-25
Blazegraph routes part 3: query routes

wwajerowicz:


delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/BlazegraphViewsRoutes.scala
```diff
@@ -98,6 +102,24 @@ class BlazegraphViewsRoutes(
                       }
                     )
                   },
+                  // Query a blazegraph view
+                  (pathPrefix("sparql") & pathEndOrSingleSlash) {
+                    implicit val um: FromEntityUnmarshaller[String] =
```
umbreak, line 107: It isn't very readable to put it inside of this block. Why not having the `FromEntityUnmarshaller` on top of the class or on the companion object?

if you don't want it to be for every `String`, it is probably better to:

```scala
final case class SparqlResponse(value: String) extends AnyVal
```

And then have the Unmarshaller on the companion object

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJson.scala
```diff
@@ -50,12 +58,22 @@ sealed trait JsonValueInstances {
 
   implicit def uioJson(
       uio: UIO[Json]
-  )(implicit s: Scheduler, cr: RemoteContextResolution, jo: JsonKeyOrdering): ResponseToJson =
+  )(implicit
```
umbreak, line 61: I wouldn't reuse this one. This is already very complicated, let's try to avoid unneccessary extra noise.

Please provide yet another implementation (like I did with `ResponseToJson`). In your case you want `ResponseToSparqlResults` or whatever. In this case it will be more readable and won't affect this.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/marshalling/RdfMarshalling.scala
```diff
@@ -37,12 +37,15 @@ trait RdfMarshalling {
     */
   implicit def jsonMarshaller(implicit
```
umbreak, line 38: revert

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJsonLd.scala
```diff
@@ -65,7 +65,7 @@ object ResponseToJsonLd extends FileBytesInstances {
           onSuccess(ioRoute.runToFuture)(identity)
         }
 
-        requestMediaType {
+        requestMediaType.apply {
```
umbreak, line 68: revert

umbreak (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/BlazegraphViewsRoutes.scala
wwajerowicz: This is `FromEntityUnmarshaller` for the request query, it's needed there to enforce correct content type on request. I agree that it can probably be moved somewhere else.

umbreak, line 107: why don't you have this at the top? If you don't want to do if for any `String` type, create your own wrapper, like:

```scala
final case class SparqlQuery(value: String) extends AnyVal
object SparqlQuery {
    implicit val um: FromEntityUnmarshaller[String] = ???
}
```

umbreak (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/BlazegraphViewsRoutes.scala
umbreak: in `SparqlClient` change the query signatures to take `SparqlQuery` instead of plain `String`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJson.scala
wwajerowicz: I think we should keep it this way, since we need some generic way to send JSON with different `Content-Type` header. I agree that there are too many implicits but adding another `ResponseTo*` class with its own implicits conversions and magic is not really solving the problem in my opinion.

umbreak: I don't agree. `ResponseToJsonLd` deals with jsonld responses, where you can have compacted/expanded/ntriples/dot responses.

It has ways to deal with IO, UIO, plain values, etc.

In your case you don't care about all those things. You want something much more simple, that only deals with `application/sparql-query`. So basically, something very similar to `ResponseToJson`.

Please provide a solution that is scoped to what you are trying to solve, instead of broadening the sof the current `ResponseToJsonLd` even more.

umbreak: plus an implicit with default value is an anti pattern most of the times.

wwajerowicz: There is no change in `ResponseToJsonLd`, only in `ResponseToJson`.

umbreak: Ok. In this case why don't you do the following:
1. Create an object with the following:
```scala
object ResponseToJsonType {
   def apply[E: JsonLdEncoder](mediaType: MediaType,
      uio: UIO[Either[Response[E], Complete[Json]]]
  )(implicit s: Scheduler, cr: RemoteContextResolution, jo: JsonKeyOrdering): Route =
  extractRequest { request =>
          if (HeadersUtils.matches(request.headers, mediaType)) {
            val ioRoute = uio.flatMap {
              case Left(r: Reject[E])       => UIO.pure(reject(r))
              case Left(e: Complete[E])     => e.value.toCompactedJsonLd.map(r => complete(e.status, e.headers, r.json))
              case Right(v: Complete[Json]) => UIO.pure(complete(v.status, v.headers, v.value))
            }
            onSuccess(ioRoute.runToFuture)(identity)
          } else reject(unacceptedMediaTypeRejection(mediaTypes))
        }
}
```
2. keep `ResponseToJson` as is but on line 35 just make a call to `ResponseToJsonType(`application/json`, uio)`
3. Create a `ResponseToSparql`:
```scala

sealed trait ResponseToSparql {
  def apply(): Route
}

object ResponseToSparql extends SparqlValueInstances {

  private[directives] type UseLeft[A] = Either[Response[A], Complete[SparqlResults]]
  private[directives] type UseRight   = Either[Response[Unit], Complete[SparqlResults]]

  private[directives] def apply[E: JsonLdEncoder](
      uio: UIO[Either[Response[E], Complete[SparqlResults]]]
  )(implicit s: Scheduler, cr: RemoteContextResolution, jo: JsonKeyOrdering): ResponseToSparql =
    new ResponseToSparql {

      override def apply(): Route = ResponseToJsonType(`application/sparql-query`, uio)
        
    }
}

sealed trait SparqlValueInstances {

  implicit def uioSpaqrl(
      uio: UIO[SparqlResults]
  )(implicit s: Scheduler, cr: RemoteContextResolution, jo: JsonKeyOrdering): ResponseToSparql = ???

  implicit def ioSparql[E: JsonLdEncoder: HttpResponseFields](
      io: IO[E, SparqlResults]
  )(implicit s: Scheduler, cr: RemoteContextResolution, jo: JsonKeyOrdering): ResponseToJson = ???

}

```

umbreak: In this way we don't need the extra implicit and we have the 2 cases separated, while the main block is handled in the same way inside `ResponseToJsonType`

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/ResponseToSparql.scala
```diff
@@ -0,0 +1,43 @@
...
+import ch.epfl.bluebrain.nexus.delta.rdf.utils.JsonKeyOrdering
+import ch.epfl.bluebrain.nexus.delta.sdk.directives.Response.Complete
+import ch.epfl.bluebrain.nexus.delta.sdk.directives.{Response, ResponseToJsonType}
+import ch.epfl.bluebrain.nexus.delta.sdk.marshalling.{HttpResponseFields, RdfMarshalling}
+import io.circe.Json
+import io.circe.syntax._
+import monix.bio.{IO, UIO}
+import monix.execution.Scheduler
+
+sealed trait ResponseToSparql {
```
umbreak, line 18: what is this sealed trait for? you are not using it

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/BlazegraphViewsRoutes.scala
```diff
@@ -1,17 +1,17 @@
 package ch.epfl.bluebrain.nexus.delta.plugins.blazegraph.routes
 
+import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.StatusCodes.Created
-import akka.http.scaladsl.model.{MediaType, MediaTypes, StatusCodes}
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.Route
 import cats.implicits._
+import ResponseToSparql._
```
umbreak, line 8: This import shouldn't be needed

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJsonType.scala
```diff
@@ -0,0 +1,40 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.utils.HeadersUtils
+import io.circe.Json
+import monix.bio.UIO
+import monix.execution.Scheduler
+
+sealed trait ResponseToJsonType {
+  def apply(): Route
+}
+
+object ResponseToJsonType {
```
umbreak, line 21: This way is fine, but then you're going to have to add the instances here:
```scala
object ResponseToJsonType extends JsonValueInstances with SparqlValueInstances
```

when you have that in place, you can and should remove the explicit import

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJson.scala
```diff
@@ -2,56 +2,28 @@ package ch.epfl.bluebrain.nexus.delta.sdk.directives
...
-import ch.epfl.bluebrain.nexus.delta.sdk.syntax._
 import io.circe.Json
 import monix.bio.{IO, UIO}
 import monix.execution.Scheduler
 
-sealed trait ResponseToJson {
-  def apply(): Route
-}
-
 object ResponseToJson extends JsonValueInstances {
```
umbreak, line 16: this `extends JsonValueInstances` is not doing anything here anymore, since what you have in the emit is an instance of `ResponseToJsonType`

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/ResponseToSparql.scala
```diff
@@ -0,0 +1,43 @@
...
+  def apply(): Route
+}
+
+object ResponseToSparql extends SparqlValueInstances {
```
umbreak, line 22: this extends SparqlValueInstance is not doing anything here, since in emit you have an instance of `ResponseToJsonType`

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJsonType.scala
wwajerowicz: I can't because `SparqlValueInstances` instances are in blazegraph plugin and this is in `sdk`

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/BlazegraphViewsRoutes.scala
wwajerowicz: It's needed because I can't make `ResponseToJsonType` extend `SparqlValueInstances`.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJsonType.scala
umbreak: Ah, then you yo have do do it differently. Simplify ResponseToJsonType to be just an object that returns the route:

```scala
object ResponseToJsonType {
	apply(): Route = ???
}
```

And call it from ResponseToJson and ResponseToSparql.

And then have something like `BlazegraphDirectives` where you have your own emit:
```
object BlazegraphDirectives {

  /**
    * Completes the current Route with the provided conversion to SparqlResults
    */
  def emit(response: ResponseToSparql): Route =
    response()
}
```

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/BlazegraphViewsRoutes.scala
umbreak: Then defeats the purpose of the Magnet Pattern

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJsonType.scala
wwajerowicz: Not sure I understand what's the `None` here.

umbreak: Sorry, you don't need None

umbreak (approved):

bogdanromanx (approved):

Issue #2086, opened 2021-02-25
Add a way to add ApiMappings from plugins to populate the projects apiMappings


umbreak:
We should have an interface like

```
trait DefaultApiMappings {
def apply(): ApiMappings
}
object DefaultApiMappings {
   def apply(pathOverrides: (String, Iri)*): DefaultApiMappings =
      new DefaultApiMappings { 
         private val mappings = ApiMappings(pathOverrides.toMap)
         def apply(): ApiMappings = mappings
      } 
}
```
That some plugins can implement and provide into the wiring. For example, for storage we would need:
```scala
many[DefaultApiMappings].add(DefaultApiMappings("defaultStorage" -> nxv + "diskStorageDefault" ))
```

Then on the core we need to collect a `Set[DefaultApiMappings]` and pass it along to the `Projects` interface, so that when a project is fetched, those ApiMappings are returned.

To remember: v1.4.x default API mappings are as follows:
```scala
  val defaultPrefixMapping: Map[String, AbsoluteIri] = Map(
    "resource"        -> Schemas.unconstrainedSchemaUri,
    "schema"          -> Schemas.shaclSchemaUri,
    "view"            -> Schemas.viewSchemaUri,
    "resolver"        -> Schemas.resolverSchemaUri,
    "file"            -> Schemas.fileSchemaUri,
    "storage"         -> Schemas.storageSchemaUri,
    "nxv"             -> nxv.base,
    "documents"       -> nxv.defaultElasticSearchIndex.value,
    "graph"           -> nxv.defaultSparqlIndex.value,
    "defaultResolver" -> nxv.defaultResolver.value,
    "defaultStorage"  -> nxv.defaultStorage.value
  )
```

PR #2115, opened 2021-03-02
Added support for collecting ApiMappings from all desired modules

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2086

### DIFF
--- a/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutesSpec.scala
+++ b/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutesSpec.scala
@@ -7,7 +7,7 @@ import akka.http.scaladsl.model.headers.{`Last-Event-ID`, Accept, OAuth2BearerTo
 import akka.http.scaladsl.server.Route
 import ch.epfl.bluebrain.nexus.delta.kernel.utils.{UUIDF, UrlUtils}
 import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
-import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.{contexts, nxv, schema}
+import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.{contexts, nxv, schema, schemas}
 import ch.epfl.bluebrain.nexus.delta.sdk.generators.{ProjectGen, ResourceGen, SchemaGen}
 import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceRef.Latest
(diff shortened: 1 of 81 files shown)


delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutesSpec.scala
```diff
@@ -52,20 +52,22 @@ class ResolversRoutesSpec
   private val asAlice = addCredentials(OAuth2BearerToken(alice.subject))
   private val asBob   = addCredentials(OAuth2BearerToken(bob.subject))
 
-  private val org      = Label.unsafe("org")
-  private val am       = ApiMappings(Map("nxv" -> nxv.base, "Person" -> schema.Person))
-  private val projBase = nxv.base
-  private val project  =
+  private val org                = Label.unsafe("org")
+  private val defaultApiMappings = ApiMappings("_" -> schemas.resources, "resource" -> schemas.resources)
```
imsdu, line 56: We could use `Resources.mappings here no` ?

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResolversRoutesSpec.scala
```diff
@@ -52,20 +52,22 @@ class ResolversRoutesSpec
...
     ProjectGen.project("org", "project2", uuid = uuid, orgUuid = uuid, base = projBase, mappings = am)
 
   private val (_, projects) = {
     implicit val subject: Subject = Identity.Anonymous
     ProjectSetup
       .init(
         orgsToCreate = List(org),
-        projectsToCreate = List(project, project2)
+        projectsToCreate = List(project, project2),
+        defaultApiMappings = defaultApiMappings
```
imsdu, line 70: And put Resources.mappings as a default value here as it seems that is what we use during the tests when it does not really matter

imsdu (changes requested):

bogdanromanx (changes requested):
Looks good, but I think we should use a plural alias everywhere:
`file` -> `files`
`archive` -> `archives`
etc.

umbreak:
No. Those are the schema aliases as in 1.4.x and get used in the following context (for example):
v1/resources/{org}/{proj}/`{schema}`/{id}

bogdanromanx:
if these are the aliases we're using in v1.4 we need to keep them, but since the segment is used as an identifier for a collection of resources it should be plural (that's why we have "documents" and not "document"); should we also add their plurals then?

umbreak:
We could. However I don't want to have a nightmarish API again where you have 100 ways of accessing the same endpoints.

bogdanromanx (approved):

Issue #2085, opened 2021-02-25
Add a way to exchange metadata fields between plugins


umbreak:
`PluginDef` should have another field (as we have with RemoteContextResolution) as follows:
```scala
/**
  * Metadata context provided by the plugin.
  */
def metadata: Json
```

This will then by used in `WiringInitializer` to compose the `contexts.metadata` from the core deepMerge with all the plugins.

Once this is done we should remove from the current metadata.json the knowledge of the modules (storages, elasicsearch, blazegraph, files, etc)

bogdanromanx:
An alternative would be something like:
```
def metadataContext: ContextValue
```
for direct context value merge.

PR #2147, opened 2021-03-11
Added a way to aggregate metadata fields

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2085

Also makes sure `@type` is displayed correctly (compacted) on listings

### DIFF
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
@@ -20,7 +20,7 @@ import ch.epfl.bluebrain.nexus.delta.sdk.marshalling.{RdfExceptionHandler, RdfRe
 import ch.epfl.bluebrain.nexus.delta.sdk.model.ComponentDescription.PluginDescription
 import ch.epfl.bluebrain.nexus.delta.sdk.model.identities.ServiceAccount
 import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectCountsCollection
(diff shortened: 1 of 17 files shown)


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
```diff
@@ -51,26 +51,37 @@ class DeltaModule(appCfg: AppConfig, config: Config)(implicit classLoader: Class
...
-        contexts.search     -> searchCtx,
-        contexts.tags       -> tagsCtx,
-        contexts.version    -> versionCtx,
-        contexts.offset     -> offsetCtx,
-        contexts.statistics -> statisticsCtx
-      )
-      .merge(otherCtxResolutions.toSeq: _*)
+  many[MetadataContextValue].addEffect(MetadataContextValue.fromFile("contexts/metadata.json"))
+
+  make[MetadataContextValue]
```
bogdanromanx, line 56: this is not needed as you can just have the `Set[MetadataContextValue]` as a dependency below...

bogdanromanx (approved):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
umbreak: I know is not needed, but it is useful since it might be used in the future by some other plugin. So instead of each plugin having to combine the `Set[MetadataContextValue]` they can use `metadataContext: MetadataContextValue @Id("aggregated-metadata")`

bogdanromanx (approved):

Issue #2143, opened 2021-03-10
EvaluationTimeout or EvaluationError will blow up on the routes


umbreak:
Reproduce:
1. Set the elasticsearch view `base` in the config to a non existing Uri
2. Create a org
3. Create a project. You will see the following errors in the logs
```
app 2021-03-10 14:59:47 WARN  c.e.b.n.d.s.p.ShardedAggregate - A Task is hiding an error while 'processing EvaluationResult for the id 'myorg/myproject_https://bluebrain.github.io/nexus/vocabulary/defaultElasticSearchIndex''
app ch.epfl.bluebrain.nexus.delta.sourcing.processor.AggregateResponse$EvaluationTimeout: null
app 	at ch.epfl.bluebrain.nexus.delta.sourcing.processor.EventSourceProcessor.toAggregateResponse$1(EventSourceProcessor.scala:207)
app 	at ch.epfl.bluebrain.nexus.delta.sourcing.processor.EventSourceProcessor.$anonfun$behavior$14(EventSourceProcessor.scala:355)
app 	at ch.epfl.bluebrain.nexus.delta.sourcing.processor.EventSourceProcessor.$anonfun$behavior$3(EventSourceProcessor.scala:191)
```

The issue is related to the fact that when an `EvaluationError` occurs inside the `ShardedAggregate`we use `IO.terminate`, which doesn't allow us to catch errors. I think we shouldn't use IO.terminate, but have a specific `Rejection` for when `EvaluationError` happens, so that we can issue an `IO.raiseError(Rejection(mapper.to(evaluationError)))`

The culprit is around [here](https://github.com/BlueBrain/nexus/blob/master/delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/processor/ShardedAggregate.scala#L123).

umbreak:
It is due to [this line](https://github.com/BlueBrain/nexus/blob/master/delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/processor/ShardedAggregate.scala#L123)

PR #2149, opened 2021-03-11
Fixed IO.terminate on EvaluationError

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2143
Fixes the rejection error when a host is not accessible.
Fixes issue on wiring (@Id missing)


delta/app/src/main/resources/app.conf
```diff
@@ -244,7 +234,7 @@ app {
       snapshot-strategy = ${app.defaults.snapshot-default}
       stop-strategy = ${app.defaults.passivate-never}
       ask-timeout = 15 seconds
-      evaluation-max-duration = 3 seconds
+      evaluation-max-duration = 11 seconds
```
imsdu, line 237: Why not 15s ?

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveRejection.scala
```diff
@@ -131,6 +135,22 @@ object ArchiveRejection {
     */
   final case class WrappedFileRejection(rejection: FileRejection) extends ArchiveRejection(rejection.reason)
 
+  /**
+    * Rejection returned when attempting to evaluate a command but the evaluation took more than the configured maximum value
+    */
+  final case class ArchiveEvaluationTimeout(command: CreateArchive, timeoutAfter: FiniteDuration)
```
imsdu, line 141: We could have only one rejection covering both which would take the EvaluationError.
From what I see, we don't process them differently and if we want to the EvaluationError can help us for that.
We may also want EvaluationError not to extend Exception so as not to be able to hide them

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveRejection.scala
```diff
@@ -131,6 +135,22 @@ object ArchiveRejection {
...
+      extends ArchiveRejection(
+        s"Timeout while evaluating the command '${ClassUtils.simpleName(command)}' for archive '${command.id}' after '$timeoutAfter'"
+      )
+
+  /**
+    * Rejection returned when attempting to evaluate a command but the evaluation took more than the configured maximum value
+    */
+  final case class ArchiveEvaluationFailure(command: CreateArchive)
```
imsdu, line 149: We lose the error message from the underlying failure

imsdu (changes requested):

delta/app/src/main/resources/app.conf
umbreak: it must be smaller than the `ask-timeout`. But yes, It can be 14s for example

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveRejection.scala
umbreak: I made it this way because It gives a better message to the client. If I have to implement the message in the common rejection is going to be as verbose:
```scala
final case class ArchiveEvaluationTimeout(err: EvaluationError) extends ArchiveRejection(err match {
case ...
})
```

So I don't see any benefit from it.

About the `Exception`, yes I will remove it

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveRejection.scala
umbreak: We don't loose it. Remember that any of those errors get logged already inside `EventSourceProcessor`. Here as a reason I just display what I want to expose to the client, which is not everything.

wwajerowicz (approved):

imsdu (approved):

Issue #2095, opened 2021-02-26
Make SSEs that target organizations and projects for specific resource types (e.g. resolvers) efficient


bogdanromanx:
A possible solution would be to add a tag on events that contains both the resource type and its scope, like for example:
```
resolver_{org_label}
resolver_{org_label}/{project_label}
```
Most of the custom resource types will not add a lot of entries in the tag_views table, with possibly the exception of files in which case it may be worth it.

Issue #2094, opened 2021-02-26
Implement Resolver SSEs for specific organization and project


bogdanromanx:
We need the SSEs endpoints to fetch SSEs from every resolvers type. This can be done with a variation of the EventExchange and the GlobalEventLog:

SSEs for all resolvers (`v1/resolvers/events` - already implemented)
SSEs for resolvers from an org (v1/resolvers/{org}/events)
SSEs for all resolvers on an project (v1/resolvers/{org}/{proj}/events)

PR #2153, opened 2021-03-12
Added tags for efficiency and consolidate SSEs

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2094
Fixes https://github.com/BlueBrain/nexus/issues/2095

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/Event.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/Event.scala
@@ -8,7 +8,7 @@ import java.time.Instant
 /**
   * Super type of all events.
   */
-trait Event extends Product with Serializable {
+sealed trait Event extends Product with Serializable {
 
   /**
     * @return the revision this events generates
@@ -34,7 +34,19 @@ object Event {
       * @return the project where the event belongs
       */
     def project: ProjectRef
+
   }
+
+  trait OrganizationScopedEvent extends Event {
+
+    /**
+      * @return the organization where the event belongs
+      */
+    def label: Label
+
+  }
+
+  trait UnScopedEvent extends Event
 
   /**
     * The global event tag.
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/Event.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/Event.scala
@@ -28,25 +28,29 @@ sealed trait Event extends Product with Serializable {
 
 object Event {
 
-  trait ProjectScopedEvent extends Event {
-
-    /**
-      * @return the project where the event belongs
-      */
-    def project: ProjectRef
-
-  }
+  trait UnScopedEvent extends Event
 
   trait OrganizationScopedEvent extends Event {
 
     /**
       * @return the organization where the event belongs
       */
-    def label: Label
-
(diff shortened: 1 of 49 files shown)


wwajerowicz (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/Event.scala
```diff
@@ -34,8 +34,20 @@ object Event {
       * @return the project where the event belongs
       */
     def project: ProjectRef
+
+  }
+
+  trait OrganizationScopedEvent extends Event {
```
imsdu, line 40: We don't really need this for `OrganizationScopedEvent` as only OrganizationEvent fits in it no ?

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resources/ResourcesImpl.scala
```diff
@@ -216,13 +216,7 @@ object ResourcesImpl {
...
       next = Resources.next,
       evaluate = Resources.evaluate(resourceResolution),
-      tagger = (ev: ResourceEvent) =>
-        Set(
-          Event.eventTag,
-          moduleType,
-          Projects.projectTag(ev.project),
-          Organizations.orgTag(ev.project.organization)
-        ),
+      tagger = EventTags.forProjectScopedEvent(moduleType),
```
imsdu, line 219: I understand that for light and rare events, we might want to add more tags but for resources, the cost may be a little high to have 6 tags for resources no ?

imsdu (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resources/ResourcesImpl.scala
umbreak: We agreed when we discussed that issue that we are going to do it for all events.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/Event.scala
umbreak: Does it hurt? I wanted to have it for consistency. Now we have 3 subclasses of `Event` and each of them makes sense.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resources/ResourcesImpl.scala
umbreak: It is precisely for resources and files that this is going to become useful.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/Event.scala
imsdu: It felt also weird when I read it as `ProjectScopedEvent` could be seen as `OrganizationScopedEvent`
Wouldn't it make more sense if you want to keep `OrganizationScopedEvent` to have that approach ?
Transform `label` to `organizationLabel` implement it in `ProjectScopedEvent` and in EventTags, `forProjectScopedEvent` would be `forOrganizationScopedEvent` merged with project related tags ?

imsdu (approved):

wwajerowicz (approved):

PR #2155, opened 2021-03-15
Allow plugins to depend on other plugins

bogdanromanx:
During service bootstrapping the plugins are discovered through scanning `PluginDef` instances in their respective jar files and loaded in separate classloaders. If the `PluginDef` makes use of any classes from a different plugin, the plugin initialization fails due to lack of access to other plugin's classes.

The change allows the initialization of plugins with a mutable ClassLoader configured as a parent to all plugin classloaders that is modified while plugins are loaded, adding each successful plugin classloader to the scope that is looked up.

### DIFF
--- a/delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/plugin/PluginsClassLoader.scala
+++ b/delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/plugin/PluginsClassLoader.scala
@@ -2,12 +2,18 @@ package ch.epfl.bluebrain.nexus.delta.service.plugin
 
 import java.net.URL
 import scala.annotation.tailrec
+import scala.collection.mutable.ListBuffer
 import scala.util.{Failure, Success, Try}
 
 /**
-  * A classloader that aggregates the plugins class loaders
+  * A mutable classloader that aggregates the plugins class loaders.
   */
 class PluginsClassLoader(pluginsClassLoader: List[PluginClassLoader], parent: ClassLoader) extends ClassLoader(parent) {
+
+  private val cls = ListBuffer.from(pluginsClassLoader)
+
+  private[plugin] def addPluginClassLoader(classLoader: PluginClassLoader): Unit =
+    cls.addOne(classLoader)
 
   /**
     * Loads the class with the specified class name.
@@ -20,7 +26,7 @@ class PluginsClassLoader(pluginsClassLoader: List[PluginClassLoader], parent: Cl
     */
   override def loadClass(className: String): Class[_] =
     loadClassFromParent(className)
-      .orElse(loadClassFromPlugins(className))
+      .orElse(loadClassFromPlugins(className, cls.toList))
       .getOrElse(throw new ClassNotFoundException(className))
 
   /**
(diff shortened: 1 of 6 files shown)


delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/plugin/PluginsClassLoader.scala
```diff
@@ -20,7 +26,7 @@ class PluginsClassLoader(pluginsClassLoader: List[PluginClassLoader], parent: Cl
     */
   override def loadClass(className: String): Class[_] =
     loadClassFromParent(className)
-      .orElse(loadClassFromPlugins(className))
+      .orElse(loadClassFromPlugins(className, cls.toList))
```
umbreak, line 29: why don't you use directly the ListBuffer ?
It doesn't seem efficient to me to convert the buffer to a list every time a class is attempted to be loaded from the plugins.

umbreak (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/plugin/PluginsLoader.scala
```diff
@@ -25,22 +27,73 @@ class PluginsLoader(loaderConfig: PluginLoaderConfig) {
...
     */
-  def load: IO[PluginError, (PluginsClassLoader, List[PluginDef])] =
-    for {
-      jarFiles                   <- UIO.delay(loaderConfig.directories.flatMap(loadFiles))
-      (pluginsDef, classLoaders) <- IO.traverse(jarFiles)(loadPluginDef).map(_.flatten.sortBy(_._1).unzip)
-      accClassLoader              = new PluginsClassLoader(classLoaders, parentClassLoader)
-    } yield (accClassLoader, pluginsDef)
+  def load: IO[PluginError, (ClassLoader, List[PluginDef])] = {
+    UIO.delay(loaderConfig.directories.flatMap(loadFiles)).flatMap { jarFiles =>
+      // recursively load the jar files, retrying in case of errors if there's at least on plugin loaded per pass
```
umbreak, line 38: at least one

umbreak (changes requested):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/plugin/PluginsLoader.scala
```diff
@@ -25,22 +27,73 @@ class PluginsLoader(loaderConfig: PluginLoaderConfig) {
...
+
+            partitioned match {
+              // everything was loaded, adding each plugin class loader and return all plugin defs
+              case (Nil, loaded)          =>
+                UIO.delay {
+                  loaded.foreach { case (_, pcl) => cl.addPluginClassLoader(pcl) }
+                  Left((Nil, cl, plugins ++ loaded.map(_._1)))
+                }
+              // nothing resolved, pick the first error and return
+              case ((_, error) :: _, Nil) =>
```
imsdu, line 61: We could return all errors no ? 
Or at least log them all and group them by plugin or something ?

imsdu (changes requested):

wwajerowicz (approved):

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/plugin/PluginsClassLoader.scala
bogdanromanx: To avoid later possible concurrent modification errors. The number of entries is very low, so there should be no problems.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/plugin/PluginsLoader.scala
bogdanromanx: yep

umbreak (approved):

imsdu (approved):

Issue #1824, opened 2020-11-24
Expose GET operations on all resource types under /resources/{org}/{proj}/_/{resource_id}


bogdanromanx:
Pre-v1.5 the resources endpoint covered operations for all resource types, like for example:
```
/resources/{org}/{proj}/_/{resource_id}
/resources/{org}/{proj}/schema/{resource_id}
/resources/{org}/{proj}/view/{resource_id}
/resources/{org}/{proj}/resolver/{resource_id}
```
Evaluate if it's necessary to maintain this compatibility.

PR #2161, opened 2021-03-16
Added support to use '_' endpoints

umbreak:
Uses a rewrite of the `uncosumedPath` form the context when the current unconsumed path is `/resources/{org}/{proj}/_/{id}`

Fixes https://github.com/BlueBrain/nexus/issues/1824


wwajerowicz (approved):

imsdu (commented):
It way be worth it to add a test in MainBehaviours that would create core resources and check that this works well at least for those ?

umbreak:
Let's cover the rest with integration tests

PR #2165, opened 2021-03-17
Added archive routes

### LINKED ISSUE #2039: Archive Plugin: Define the routes

bogdanromanx:
Fixes #2039


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageRejection.scala
```diff
@@ -266,4 +269,21 @@ object StorageRejection {
 
   implicit final val evaluationErrorMapper: Mapper[EvaluationError, StorageRejection] = StorageEvaluationError.apply
 
+  implicit final val storageRejectionHttpResponseFields: HttpResponseFields[StorageRejection] =
```
umbreak, line 272: If you are putting the `HttpResponseFields` in the *Rejection instead of in the *Routes, can you do it as well for `ElasticSearchViews` and `BlazegraphViews` to be consistent?

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceUris.scala
```diff
@@ -55,7 +55,17 @@ object ResourceUris {
     def incomingShortForm(implicit base: BaseUri): Uri = accessUriShortForm / "incoming"
     def outgoingShortForm(implicit base: BaseUri): Uri = accessUriShortForm / "outgoing"
     def project(implicit base: BaseUri): Uri           = ResourceUris.project(projectRef).accessUri
+  }
 
+  /**
+    * A resource that is rooted in a project but not persisted or indexed.
+    */
+  final case class EphemeralResourceInProjectUris(
```
umbreak, line 63: I understand that's just to get rid of outgoing and incoming links, but I don't know if its worth it. The incoming and outgoing endpoints will just work and return an empty result, which is fine.

Besides notice that incoming links might be present on the platform, from some people that created a resource and linked it to the id of the archive.

Basically, you can keep it or remove it, as you like :)

umbreak (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageRejection.scala
bogdanromanx: right, done

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/ResourceUris.scala
bogdanromanx: I would keep it as is, since the example is an edge-case that most likely will never happen.

umbreak (approved):

wwajerowicz (approved):

PR #2170, opened 2021-03-18
Adapted archives integration tests

bogdanromanx:


tests/src/test/resources/kg/archives/archive-response.json
```diff
@@ -1,37 +1,32 @@
...
+      "project": "{{project2}}",
+      "resourceId": "https://dev.nexus.test.com/simplified-resource/2"
     },
     {
-      "@type" : "File",
-      "path" : "some/other/nexus-logo.png",
-      "project" : "{{project1}}",
-      "resourceId" : "https://dev.nexus.test.com/simplified-resource/logo"
+      "@type": "File",
+      "path": "/some/other/nexus-logo.png",
```
umbreak, line 16: so the project is not art of the response? It should be

umbreak (changes requested):

tests/src/test/resources/kg/archives/archive-response.json
umbreak, line 16: the `project` field should always be present in the response, whether it is another project or the current project.

umbreak (changes requested):

umbreak (approved):

Issue #2080, opened 2021-02-24
Views SSEs


umbreak:
We need the SSEs endpoints to fetch SSEs from every view type. This can be done with a variation of the `EventExchange` and the `GlobalEventLog`:
- SSEs for all views (`v1/views/events`)
- SSEs for views from an org (`v1/views/{org}/events`)
- SSEs for all views on an project (`v1/views/{org}/{proj}/events`)

Issue #1807, opened 2020-11-20
The resource event routes do not include all event types (e.g.: resolvers, schemas, views etc.)


bogdanromanx:
This is a change of behaviour since v1.4

umbreak:
There are 3 ways to deal with it:
- Create a tag `event` for each resource type and fetch them using `eventsByTag`.
- Request each tag from each module separately and merge all the streams ordered by timestamp
- Do not provide this global functionality and just offer SSEs per event type (resources, resolvers, files, etc)

umbreak:
We need the following SSEs, which can be implementing using the `EventExchange` and the `GlobalEventLog`:
- SSEs for all resources (`v1/resources/events`)
- SSEs for all resources on an org (`v1/resources/{org}/events`)
- SSEs for all resources on an project (`v1/resources/{org}/{proj}/events`)

PR #2171, opened 2021-03-19
Added support for aggregated SSEs

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2080
Fixes https://github.com/BlueBrain/nexus/issues/1807

I've refactored a bit things. I've reintroduced `EventExchange`, since what is doing is pretty different from `ReferenceExchange` (event thought there is a little code duplication I feel is worth to have good separation of concerns. For example:
- we don't need a `ReferenceExchange` for projects, but we need an `EventExchange`. 
- we don't need metadata transformation for a `ReferenceExchange` but we need it for `EventExchange`.

I've also removed `GlobalEventLog` since its signature wasn't useful for SSEs and it is only used in views. Now each view has its own.

I've introduced `SseEventLog` to deal with Sse events.


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/SseEventLog.scala
```diff
@@ -0,0 +1,95 @@
...
+  def apply(
+      eventLog: EventLog[Envelope[Event]],
+      orgs: Organizations,
+      projects: Projects,
+      exchanges: Set[EventExchange]
+  ): SseEventLog = new SseEventLog {
+
+    private lazy val exchangesList = exchanges.toList
+
+    def stream(offset: Offset): Stream[Task, Envelope[JsonLdValue]] =
```
imsdu, line 63: We could also serialize the enveloppe too no and return the complete serialized value no ?
`Envelope[JsonLdValue]]` seems a little weird, either we want the `Envelope[Event]` or the JsonLd no ?

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/SseEventLogDummy.scala
```diff
@@ -0,0 +1,50 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.Event.{OrganizationScopedEvent, ProjectScopedEvent}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.organizations.OrganizationRejection
+import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.{ProjectRef, ProjectRejection}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{Envelope, Event, Label}
+import ch.epfl.bluebrain.nexus.delta.sdk.{JsonLdValue, SseEventLog}
+import fs2.Stream
+import monix.bio.{IO, Task, UIO}
+// $COVERAGE-OFF$
+
+final class SseEventLogDummy(envelopes: Seq[Envelope[Event]], f: PartialFunction[Event, JsonLdValue])
```
imsdu, line 15: As we don't really care about the returned values which should be tested elsewhere (and if we remove the need of an envelope as I mentioned in another comment), we could just have:
```
final class SseEventLogDummy(events: (Label, Org, JsonLdValue))
```
And filter the list in the project / org method

imsdu (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
```diff
@@ -563,7 +569,7 @@ object BlazegraphViews {
       initialState = Initial,
       next = next,
       evaluate = evaluate(validateP, validateRef),
-      tagger = EventTags.forProjectScopedEvent(moduleType),
+      tagger = EventTags.forProjectScopedEvent(moduleTag),
```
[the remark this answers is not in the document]
umbreak: Why to keep the other one? 

There is no use case currently for that (note that if I do that I have to create a custom `EventTags `, which I don't want unless is necessary.

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphIndexingEventLog.scala
```diff
@@ -0,0 +1,77 @@
...
+          tag: Option[TagLabel]
+      ): Stream[Task, Chunk[Message[ResourceF[Graph]]]] =
+        eventLog
+          .eventsByTag(Projects.projectTag(project), offset)
+          .map(_.toMessage)
+          .groupWithin(batchMaxSize, batchMaxTimeout)
+          .discardDuplicates()
+          .evalMapFilterValue { event =>
+            Task
+              .tailRecM(exchangesList) { // try all event exchanges one at a time until there's a result
```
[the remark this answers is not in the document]
umbreak: This is something I already mentioned to @bogdanromanx in the original PR where he did this.

there is definitely a way to do it. However it can get a little complicated quickly. So let's think about the tradeoffs in another PR

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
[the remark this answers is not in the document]
imsdu: When we populate the view caches and start the indexing

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/SseEventLog.scala
```diff
@@ -0,0 +1,95 @@
...
+    )(implicit mapper: Mapper[ProjectRejection, R]): IO[R, Stream[Task, Envelope[JsonLdValue]]] =
+      projects
+        .fetch(project)
+        .as(exchange(eventLog.eventsByTag(Projects.projectTag(project), offset)))
+        .mapError(mapper.to)
+
+    private def exchange(stream: Stream[Task, Envelope[Event]]): Stream[Task, Envelope[JsonLdValue]] =
+      stream
+        .map { envelope =>
+          exchangesList.tailRecM[Option, Envelope[JsonLdValue]] {
```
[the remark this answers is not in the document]
umbreak: Deferred to later PR. This wasn't introduced by me but by @bogdanromanx

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/SseEventLog.scala
umbreak: No, we want both, asi you can see in `ResponseToJsonLd`.

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/SseEventLogDummy.scala
umbreak: We need the envelopes, as I pointed out

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/SseEventLog.scala
umbreak: And I don't want to change `ResponseToJsonLd`  even more to accommodate to this.
`Envelope[JsonLdValue]` is fine since the `JsonLdValue` still contains the `Event`. So you keep having all the information if you wanted

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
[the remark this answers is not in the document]
umbreak: Good point. Then I'm adding it

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/SseEventLog.scala
umbreak: We have to send a `ServerSentEvent`, which requires the json data plus some of the envelope fields.
I could make the `SseEventLog` return the `ServerSentEvent` directly. But I don't think that's any better because
1. Is akka specific.
2. We will still need `Envelope` to support all the other use-cases in `ResponseToJsonLd `

So in order to keep `ResponseToJsonLd` sane and to be aligned with the regular SSE events, I think it is a good tradeoff

imsdu: We can see it later but maybe the regular events could evolve a bit to have a common ground for the two types of SSEs

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/SseEventLogDummy.scala
umbreak: I'm open to something like that If it makes things more readable. So let's consider that for a separate PR if we arrive to a better solution.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/SseEventLog.scala
umbreak: I'm open to something like that If it makes things more readable. So let's consider that for a separate PR if we arrive to a better solution.

imsdu: I am ok with it, so let's open issues for the different points to resolve in other PRs so that we talk further about it during the next issue curation session.

imsdu (approved):

PR #2174, opened 2021-03-23
Distribute streams thanks to cluster sharding

imsdu:
Move away from Cluster Singleton to use regular cluster sharding with remember entities feature enabled

* Use only one sharded actor for every indexing stream, make sure it retrieves all information when it starts
* Split the `StreamSupervisor` in 2 coordinators implementations to handle regular streams and indexing streams which can be restarted/stopped
* Introducing `StreamSwitch` to start/stop streams and avoid actor reboot hysteria thanks to an uuid
* Enable the new rebalance algorithm from akka 2.6.10: https://doc.akka.io/docs/akka/current/typed/cluster-sharding.html#shard-allocation


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamBehaviour.scala
```diff
@@ -0,0 +1,284 @@
...
+        def init(): Behavior[IndexingViewCommand[V]] = {
+          logger.debug("Init indexing for view {} in project", id, project)
+          runStream(None)
+          balancing()
+        }
+
+        // State after init or after reset and when view gets updated
+        def balancing(): Behavior[IndexingViewCommand[V]] = Behaviors
+          .receiveMessage[IndexingViewCommand[V]] {
+            case Started(v, s)    =>
```
umbreak, line 140: don't you need any `ClassTag` on the pattern match here? How does it know the type after erasure?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamBehaviour.scala
```diff
@@ -0,0 +1,284 @@
...
+import com.typesafe.scalalogging.Logger
+import fs2.Stream
+import monix.bio.{Task, UIO}
+import monix.execution.Scheduler
+
+import java.util.UUID
+import scala.util.{Failure, Success}
+
+/**
+  * The view index behavior defined in order to manage the indexing stream
```
umbreak, line 23: It took me a while to follow this. An extra documentation with the different behaviour and under which conditions you jump from a behaviour to the next would be useful

umbreak (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphIndexingCoordinator.scala
```diff
@@ -100,40 +105,76 @@ private class IndexingStream(
...
+        (res, progress) => new IndexingStream(client, cache, res, config).build(indexingLog, projection, progress),
+        client.deleteNamespace(_).void,
+        projection,
+        indexingRetryStrategy
+      )
+    }
+    .tapEval { coordinator =>
+      def onEvent(event: BlazegraphViewEvent) = coordinator.run(event.id, event.project, event.rev)
+      IO.unless(MigrationState.isIndexingDisabled)(
+        BlazegraphViewsIndexing("BlazegraphIndexingCoordinatorScan", config.indexing, views, onEvent)
```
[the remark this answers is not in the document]
imsdu: The reason is that it introduces a circular dependency between the BlazegraphViews and its coordinator that distage was not able to handle (it could not proxy BlazegraphViews, asking if it was final).
So the reason was to break this dependency as we should avoid them when we can.
There would be another option to handle this, but this would require to init the cache in the module and not in BlazegraphViews.

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphIndexingCoordinator.scala
```diff
@@ -100,40 +105,76 @@ private class IndexingStream(
...
 
-  type StartCoordinator              = IndexingViewResource => Task[Unit]
-  type StopCoordinator               = IndexingViewResource => Task[Unit]
-  type BlazegraphIndexingCoordinator = IndexingStreamCoordinator[IndexingViewResource]
-
   private val logger: Logger = Logger[BlazegraphIndexingCoordinator.type]
 
+  type BlazegraphIndexingCoordinator = IndexingStreamCoordinator[IndexingBlazegraphView]
+
+  private[indexing] def illegalArgument[A](error: A) =
```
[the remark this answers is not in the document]
imsdu: I had merge conflicts, I will check that

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ProjectsCounts.scala
```diff
@@ -64,7 +66,7 @@ object ProjectsCounts {
 
     def buildStream(
         progress: ProjectionProgress[ProjectCountsCollection]
-    ): Stream[Task, ProjectCountsCollection] = {
+    ): Stream[Task, Unit] = {
```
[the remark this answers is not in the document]
imsdu: It is not used for now and it is meant to have it similar to other streams

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamBehaviour.scala
```diff
@@ -0,0 +1,284 @@
...
+                    case Some((currentIndex, switch)) if fetchView.rev > currentIndex.rev =>
+                      UIO.delay(
+                        logger.debug(
+                          "New revision of view {} in project {} is detected, the indexing will be restart from the beginning",
+                          fetchView.rev,
+                          id,
+                          project
+                        )
+                      ) >>
+                        switch.stop >> clearIndex(currentIndex.index) >> restartStream(fetchView)
```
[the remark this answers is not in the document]
imsdu: It just should be an UIO and not a Task no ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamBehaviour.scala
```diff
@@ -0,0 +1,284 @@
...
+                            project,
+                            progress.offset
+                          )
+                      ) >>
+                        StreamSwitch.run(
+                          streamName(fetchView.rev),
+                          buildStream(fetchView, progress),
+                          retryStrategy,
+                          onFinalize,
+                          onFinalize
```
umbreak, line 86: On cancel we would probably prefer to try to restart the stream, instead of just passivating it, no?

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamBehaviour.scala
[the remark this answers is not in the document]
umbreak: Yeah probably that. But on the caller side we should at least log the error. It is not critical if we don't delete the previous index but it can create a lot of garbage, so we would like to know that happened.

Alternatively we could have a healing process that deletes unused indices (in the future).

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamBehaviour.scala
imsdu: The Stop message does not passivate the actor, it gets restarted after a backoff

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphIndexingCoordinator.scala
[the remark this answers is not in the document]
umbreak: If that's the case I would rather have something more explicit in `BlazegraphViewsIndexing`, like:
```scala
def populateCache(name: String, config: ExternalIndexingConfig, views: BlazegraphViews) =
   apply(name, 
   config, 
   views, 
   event  =>
                               views
                                 .fetch(event.id, event.project)
                                 .redeemCauseWith(_ => IO.unit, res => index.put(ViewRef(res.value.project, res.value.id), res))
) 

def triggerCoordinator(name: String, config: ExternalIndexingConfig, coordinator: BlazegraphIndexingCoordinator) =
   apply(name, config, views, event => coordinator.run(event.id, event.project, event.rev))
```

same for elasticsearch

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamBehaviour.scala
umbreak: Where is that happening? I see `Behaviors.stopped` when receiving the message `StreamStopped`.
And if the actor gets restarted that means it also gets restarted on manual cancelation, which is not what we probably want, right?

imsdu: Remember entities is taking responsibility for restarting the actor as long as passivation is not explicitly asked (which only happens when it is deprecated).

umbreak: To me there is a difference between
1. The stream will start/restart when the cluster sharded entity is requested again by somewhere within our app
2. The stream will start/restart immediately after a failure

I understand that when using cluster sharding, and requesting an entity, this one will get spawn if not already present. But it is not that what I'm talking about.

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphPluginModule.scala
```diff
@@ -138,10 +145,9 @@ class BlazegraphPluginModule(priority: Int) extends ModuleDef {
         cr: RemoteContextResolution @Id("aggregate"),
         ordering: JsonKeyOrdering
     ) =>
-      new BlazegraphViewsRoutes(views, viewsQuery, identities, acls, projects, progresses, coordinator)(
+      new BlazegraphViewsRoutes(views, viewsQuery, identities, acls, projects, progresses, coordinator.restart)(
```
umbreak, line 148: Your proxy error could have been caused by this: `coordinator.restart`
Since you are accessing `coordinator` here when distage might not have yet been able to get its dependencies

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamCoordinator.scala
```diff
@@ -1,138 +1,86 @@
...
     }
 
-    val definition = TransientEventDefinition.cache(entityType, Initial, eval, TransientStopStrategy.never)
-    ShardedAggregate.transientSharded(definition, config, Some(retryStrategy)).map(new IndexingStreamCoordinatorImpl(_))
+  def send(id: Iri, project: ProjectRef, command: IndexingViewCommand[V]): UIO[Unit] = UIO.delay {
+    val clusterSharding = ClusterSharding(as)
+    clusterSharding.entityRefFor(key, entityId(project, id)) ! command
   }
 
+  def restart(id: Iri, project: ProjectRef): UIO[Unit] = send(id, project, Restart)
```
umbreak, line 70: docs?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamCoordinator.scala
```diff
@@ -1,138 +1,86 @@
...
-      case (Initial, RestartIndexing(V(view)))                           => startFromBeginning(view)
-      case (cur: Current, RestartIndexing(V(view)))                      => cur.supervisor.stop() >> startFromBeginning(view)
-      case (_, StartIndexing(o))                                         => Task.raiseError(new IllegalArgumentException(s"Wrong type '${simpleName(o)}'"))
-      case (_, RestartIndexing(o))                                       => Task.raiseError(new IllegalArgumentException(s"Wrong type '${simpleName(o)}'"))
+      clusterSharding.entityRefFor(key, entityId(project, id)) ! ViewRevision(rev)
     }
 
-    val definition = TransientEventDefinition.cache(entityType, Initial, eval, TransientStopStrategy.never)
-    ShardedAggregate.transientSharded(definition, config, Some(retryStrategy)).map(new IndexingStreamCoordinatorImpl(_))
+  def send(id: Iri, project: ProjectRef, command: IndexingViewCommand[V]): UIO[Unit] = UIO.delay {
```
umbreak, line 65: `private` or `private[indexing]`

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamBehaviour.scala
imsdu: There is a retry strategy to restart the stream and it does not restart the actor.
But when this one is not applicable, only then it escalates to the actor which gets restarted somewhere else after a backoff.
Which makes sense because, the akka node may not be healthy

umbreak: I just want to make sure, when a stop (stream interruption) is requested, actor does not try to run again the stream.

I don't anywhere where this is explicitly coded though (I might have missed it). In the cluster sharding documentation it says

> When rememberEntities is enabled, whenever a Shard is rebalanced onto another node or recovers after a crash, it will recreate all the entities which were previously running in that Shard.

> To permanently stop entities send a ClusterSharding.Passivate to the ActorRef[ShardCommand] that was passed in to the factory method when creating the entity. Otherwise, the entity will be automatically restarted after the entity restart backoff specified in the configuration.

So I guess you are sending a `Passivate` in that case? You mentioned you are sending a `Passivate` on deprecation

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphPluginModule.scala
imsdu: It was `BlazegraphViews` that could not be proxied with a cglib error asking if it was final (and it was final).
The routes come later in the dependency tree.
With circular dependencies, we expose ourselves to this proxy problem, it also makes code and tests more complicated so getting rid of it when we can is not a bad thing.
I will see how to make the change about populating cache and view easier to understand.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamBehaviour.scala
imsdu: Yes, when we don't want the entity to respawn automatically we have to use passivation, not the stop message.
But so far, there is no endpoint to passivate a stream, just to restart it.
So passivation only occurs when the fetched view is deprecated.

I will add the links in the akka documentation in the comments.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamBehaviour.scala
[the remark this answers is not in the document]
imsdu: I just logged it for now, I created an issue to discuss about setting up a healing process: https://github.com/BlueBrain/nexus/issues/2176

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/indexing/IndexingStreamBehaviour.scala
imsdu: It comes from the way, the command ADT is written

umbreak (approved):

bogdanromanx (approved):

wwajerowicz (approved):

PR #2178, opened 2021-03-23
Redirect by default when creating an archive

### LINKED ISSUE #2164: Ensure plugins can be loaded when the app starts

### LINKED ISSUE #2175: Archive creation should redirect to the archive download address by default
`Accept: */*` or `Accept: application/x-tar` when creating an archive should return a SeeOther response with the appropriate location for downloading the archive that was just created.

An `Accept` header with a JsonLD compatible media type should return the Archive creation metadata.

bogdanromanx:
Fixes #2175
Fixes #2164


delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/routes/ArchiveRoutes.scala
```diff
@@ -87,6 +93,11 @@ class ArchiveRoutes(
 
   private def sourceToFileResponse(source: AkkaSource): FileResponse =
     FileResponse("archive.tar", MediaTypes.`application/x-tar`, 0L, source)
+
+  private def metadataResponse: Directive1[Boolean] =
```
umbreak, line 97: you don't need this, you just need:

```scala
extractRequest.map(req => HeadersUtils.findFirst(req.headers, mediaTypes).isDefined)
```

umbreak (changes requested):

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/routes/ArchiveRoutes.scala
bogdanromanx: I don't see how that's better, but alright...

imsdu (approved):

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/routes/ArchiveRoutes.scala
umbreak: It is better because we keep using the same methods for the same things concepts.

If you don't find the current method in `HeadersUtils` a useful one or it can be improved, please go on and change it (also changing it where is used, DeltaDirectives and so on).

Something like that might be better than what we currently have:

```
def findFirst(accept: Accept, serviceMediaTypes: Seq[MediaType]): Option[MediaType]= ???
```

Then we benefit everywhere (not just on the ArchivesRoutes)

bogdanromanx: The accept header may not be present, in which case it's better to just pass along the header collection because it makes it simple at the call site.

But for this function to work with archives, we need to choose whether we want an exact match or not. I've modified it to do just that.

umbreak: you don't need the exact match. Just have a:

```scala
private def metadataResponse: Directive1[Boolean] =
extractRequest.map(req => !HeadersUtils.findFirst(req.headers, Set(`application/x-tar`)).isDefined)
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/utils/HeadersUtils.scala
```diff
@@ -8,15 +9,33 @@ object HeadersUtils {
...
-  def findFirst(headers: Seq[HttpHeader], serviceMediaTypes: Seq[MediaType]): Option[MediaType] = {
-    val ct       = new MediaTypeNegotiator(headers)
-    val accepted = if (ct.acceptedMediaRanges.isEmpty) List(MediaRanges.`*/*`) else ct.acceptedMediaRanges
-    accepted.foldLeft[Option[MediaType]](None) {
-      case (s @ Some(_), _) => s
-      case (None, mr)       => serviceMediaTypes.find(mt => mr.matches(mt))
+  def findFirst(
+      headers: Seq[HttpHeader],
+      serviceMediaTypes: Seq[MediaType],
+      exactMatch: Boolean = false
```
umbreak: this is not needed

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/utils/HeadersUtils.scala
```diff
@@ -8,15 +9,33 @@ object HeadersUtils {
...
+  ): Option[MediaType] =
+    if (exactMatch)
```
umbreak: this neither

umbreak (approved):

PR #2179, opened 2021-03-24
Added SDK views module to avoid leaking login into the SDK

umbreak:


build.sbt
```diff
@@ -502,6 +524,7 @@ lazy val elasticsearchPlugin = project
   .dependsOn(
     migration  % Provided,
     sdk        % "provided;test->test",
+    sdkViews,
```
bogdanromanx, line 527: I think the service should compile depend on sdkViews and the plugins should use this dependency in scope provided.

bogdanromanx (changes requested):

build.sbt
umbreak: Why the service needs to know about `sdkViews`?

The current setup works and it seems to me is the one which makes more sense:
- sdkViews has as provided the SDK
- sdkViews is used only by plugins who need it (elasticsearch, blazegraph and compositeviews)

bogdanromanx: While the current configuration also works it will result in larger plugin bundles... are we ok with that?

umbreak: There is a minuscule increase in size:
before:
```
396K	$HOME/github/nexus/delta/app/target/universal/stage/plugins/archive.jar
984K	$HOME/github/nexus/delta/app/target/universal/stage/plugins/blazegraph.jar
840K	$HOME/github/nexus/delta/app/target/universal/stage/plugins/elasticsearch.jar
5.4M	$HOME/github/nexus/delta/app/target/universal/stage/plugins/storage.jar
```
after:
```
396K	$HOME/github/nexus/delta/app/target/universal/stage/plugins/archive.jar
1.0M	$HOME/github/nexus/delta/app/target/universal/stage/plugins/blazegraph.jar
888K	$HOME/github/nexus/delta/app/target/universal/stage/plugins/elasticsearch.jar
5.4M	$HOME/github/nexus/delta/app/target/universal/stage/plugins/storage.jar
```

bogdanromanx: agreed

bogdanromanx (changes requested):
The CI files also need to be modified to build the dependent plugins and app when changes are detected on this path.

umbreak:
Done. I've also added the composite views to the CI...for some reason it wasn't added yet

bogdanromanx (dismissed):

PR #2186, opened 2021-03-25
Preparation for CompositeView indexing logic

umbreak:
Keeping the `ProgressStrategy` not sealed will allow us to extend it on the composite views module with some more restart strategies like for example:
```scala
final case class PartialRestart(projectionIds: Set[CompositeViewProjectionId]) extends ProgressStrategy
```

Also moving the logic to deal with the progress to each of the implementations will make it easier in the case of composite views instead of having an abstraction that has to deal with that many things.


delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphStreamBuilder.scala
```diff
@@ -0,0 +1,117 @@
...
+      progressStrategy: StreamBuilder.ProgressStrategy
+  ): Stream[Task, Unit] = {
+
+    implicit val projectionId: ProjectionId = view.projectionId
+    Stream
+      .eval {
+        for {
+          _        <- prev.fold(Task.unit)(prev => client.deleteNamespace(prev.index).void >> cache.remove(prev.projectionId))
+          props    <- ClasspathResourceUtils.ioPropertiesOf("blazegraph/index.properties")
+          _        <- client.createNamespace(view.index, props)
```
imsdu, line 50: We don't have to attempt to create the namespace everytime, just if it is a new view or if it has been updated

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchStreamBuilder.scala
```diff
@@ -0,0 +1,129 @@
...
+      prev: Option[ViewIndex[IndexingElasticSearchView]],
+      view: ViewIndex[IndexingElasticSearchView],
+      progressStrategy: StreamBuilder.ProgressStrategy
+  ): Stream[Task, Unit] = {
+    implicit val projectionId: ProjectionId = view.projectionId
+    Stream
+      .eval {
+        for {
+          _        <- prev.fold(Task.unit)(prev => client.deleteIndex(idx(prev)).void >> cache.remove(prev.projectionId))
+          _        <- client.createIndex(idx(view), Some(view.value.mapping), view.value.settings)
```
imsdu, line 52: Same

delta/sdk-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/views/indexing/IndexingStreamBehaviour.scala
```diff
@@ -188,49 +142,43 @@ object IndexingStreamBehaviour {
...
-              case ViewRevision(rev) if rev > viewIndex.rev   =>
-                logger.debug(
-                  "'ViewRevision' event has been received and a newer revision exists for view {} in project {}",
-                  id,
-                  project
-                )
-                runStream(Some((viewIndex, switch)))
+              case ViewRevision(rev) if rev > viewIndex.rev =>
+                logger.debug("'ViewRevision' event has been received and a newer revision exists for view {} in project {}", id, project)
+                runStream(Some(Started(viewIndex, switch)), ProgressStrategy.Continue)
```
imsdu, line 155: A dedicated `ProgressStrategy` when a view is updated may make things easier ?

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/projections/stream/DaemonStreamCoordinator.scala
```diff
@@ -21,30 +21,29 @@ import scala.collection.concurrent
...
+    Task.delay(tasks.put(name, DaemonStream(name, stream, retryStrategy))) >>
       Task.delay {
         val settings    = ClusterShardingSettings(as).withRememberEntities(true)
         val shardingRef = ClusterSharding(as).init(
           Entity(EntityTypeKey[SupervisorCommand]("daemonStream")) { entityContext =>
             val task = tasks(entityContext.entityId)
             DaemonStreamBehaviour(
               entityContext.entityId,
-              task.streamTask,
+              task.stream,
```
imsdu, line 46: daemon instead of task

imsdu (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphStreamBuilder.scala
umbreak: Even though this is true, this operation is not expensive. To add the logic to make this work is probably worse than just try every time we build a stream.

delta/sdk-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/views/indexing/IndexingStreamBehaviour.scala
umbreak: I don't want to make things overly complicated. They are already a bit complicated. And I don't see a need for that distinction

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphStreamBuilder.scala
imsdu: In the current implementation, it is a:
```
IO.when(prev.isDefined) { client.createNamespace(view.index, props)  }
```

umbreak: that's wrong. The first time you don't have a previous

delta/sdk-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/views/indexing/IndexingStreamBehaviour.scala
imsdu: The new progress strategy could carry the information of the old view to stop and we could get rid of this option.
But it is just a thought, maybe it is not worth it.

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphStreamBuilder.scala
imsdu: My bad, I need another coffee I guess :).
But the point was it should not be that complicated

delta/sdk-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/views/indexing/IndexingStreamBehaviour.scala
umbreak: On the current state of things it does not seem very useful. We will see how it evolves with composite views, but I don't expect the need for anything more complicated.

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphStreamBuilder.scala
umbreak: Yes it shouldn't. But my point is that it is not worth it, since those methods already handle well the cases when the index/namespace already exists.

In this case I prefer a consistent and simple logic, than to pass along more information just to fulfill this.

wwajerowicz (approved):

imsdu (approved):

PR #2202, opened 2021-03-29
Encrypt remote source access token

wwajerowicz:


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/config/CompositeViewsConfig.scala
```diff
@@ -24,7 +25,8 @@ final case class CompositeViewsConfig(
     aggregate: AggregateConfig,
     keyValueStore: KeyValueStoreConfig,
     pagination: PaginationConfig,
-    indexing: ExternalIndexingConfig
+    indexing: ExternalIndexingConfig,
+    encryption: EncryptionConfig
```
imsdu, line 29: The encryption should be common to all the application even if it is only used in two plugins no ?
Allowing people to use different passwords/salt for the two plugins does not bring much.

imsdu (changes requested):

imsdu (dismissed):

bogdanromanx (approved):

PR #2218, opened 2021-04-01
CompositeView: Implemented indexing logic for non-remote sources

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2103

- Indexing stream logic now is compacted into `IndexingStream` (and its implementations `ElasticSearchIndexingStream`, `BlazegraphIndexingStream` and `CompositeIndexingStream`).
- Made `rebuildStrategy` optional field.
- Made context a `ContextObject` instead of a `Json`.

TODO:
- Deletion of previous indices will be done as an [asynchronous process](https://github.com/BlueBrain/nexus/issues/2176) in order to split the complexity (the current `IndexingStream` implementations are already dealing with too many things).
- Verify correct ProjectionProgress of CompositeViews
- Verify correct restart of CompositeViews
- Add validation on projection `query`


.github/workflows/ci-sdk-views.yml
```diff
@@ -0,0 +1,71 @@
...
+      - 'delta/sdk-views/**'
+      - 'delta/sourcing/**'
+      - 'build.sbt'
+      - 'project/**'
+jobs:
+  review:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
```
bogdanromanx, line 30: remove strategy, it only applies to matrix buils

.github/workflows/ci-sdk-views.yml
```diff
@@ -0,0 +1,71 @@
...
+        uses: coursier/cache-action@v5
+      - name: Setup Scala
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.11
+      - name: Static Analysis
+        run: sbt -Dsbt.color=always -Dsbt.supershell=false "project sdkViews" clean scalafmtCheck test:scalafmtCheck scalafmtSbtCheck scapegoat
+      - name: Tests
+        run: sbt -Dsbt.color=always -Dsbt.supershell=false "project sdkViews" clean coverage test coverageReport coverageAggregate
+  publish:
```
bogdanromanx, line 45: remove this, it's no longer present in the other CI jobs; it has been replaced by snapshot

.github/workflows/ci-sdk-views.yml
```diff
@@ -0,0 +1,71 @@
...
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: true
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Coursier Cache
+        uses: coursier/cache-action@v5
```
bogdanromanx, line 36: coursier/cache-action@v6

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
```diff
@@ -531,7 +544,8 @@ object BlazegraphViews {
                              )
       views                = new BlazegraphViews(agg, eventLog, index, projects, orgs, sourceDecoder)
       _                   <- validateRefDeferred.complete(validateRef(views))
-      _                   <- BlazegraphViewsIndexing.populateCache(config.indexing, views, index).void
+      _                   <- BlazegraphViewsIndexing.deleteNotUsedNamespaces()
```
bogdanromanx, line 547: if this is supposed to run asynchronously... it needs to be forked, right?
additionally, if this replays the event log, it should probably be executed along with cache indexing to avoid multiple concurrent replays.

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeIndexingCoordinator.scala
```diff
@@ -0,0 +1,79 @@
...
+  private def fetchView(views: CompositeViews, config: CompositeViewsConfig) = (id: Iri, project: ProjectRef) =>
+    views
+      .fetch(id, project)
+      .map { res =>
+        Some(
+          ViewIndex(
+            res.value.project,
+            res.id,
+            res.value.uuid,
+            ViewProjectionId("none"),
```
bogdanromanx, line 35: is this right?

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeView.scala
```diff
@@ -26,7 +26,7 @@ final case class CompositeView(
     project: ProjectRef,
     sources: NonEmptySet[CompositeViewSource],
     projections: NonEmptySet[CompositeViewProjection],
-    rebuildStrategy: RebuildStrategy,
+    rebuildStrategy: Option[RebuildStrategy],
```
bogdanromanx, line 29: would this work better if we were to have `RebuildStrategy.None`?

bogdanromanx (changes requested):
Looks good overall.

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeIndexingCoordinator.scala
umbreak: Yes. There is no projection for the whole view, but for each compositeView.projection

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeView.scala
umbreak: Why?

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
umbreak: As you can see this is not yet implemented. I'll come on the details when I actually implement it.

But no, it doesn't need to be forked. Currently `populateCache` is not being forked and it works, right? Note that in the end it calls `DaemonStreamCoordinator.run`

bogdanromanx (approved):

wwajerowicz (approved):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeIndexingStream.scala
```diff
@@ -0,0 +1,267 @@
...
+                case prev                                    =>
+                  prev.as(SkipIndexingUntil(prev.value.offset.max(skipIndexingUntil.offset)))
+              }
+              .getOrElse(sProgress.as(skipIndexingUntil))
+            acc + (pProgress.value -> pProgress.as(noSkipIndexing)) + (sId -> sourceProgress)
+        })
+    }
+  }
+
+  private def idx(projection: ElasticSearchProjection, rev: Long): IndexLabel =
```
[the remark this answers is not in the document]
umbreak: Isn't it clear enough since the parameter of the method is a `projection`?

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeIndexingStream.scala
```diff
@@ -0,0 +1,267 @@
...
+    IndexLabel.unsafe(ElasticSearchViews.index(projection.uuid, rev, esConfig))
+
+  private def ns(projection: SparqlProjection, rev: Long): String =
```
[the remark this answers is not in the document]
umbreak: Isn't it clear enough since the parameter of the method is a `projection`?

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeIndexingStream.scala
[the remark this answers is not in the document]
imsdu: I was thinking about the generated ES index name, it may be useful have a different pattern than the ones created by the elasticsearch_plugin

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphIndexingStream.scala
```diff
@@ -0,0 +1,93 @@
...
+  ): Stream[Task, Unit] =
+    Stream
+      .eval {
+        // Evaluates strategy and set/get the appropriate progress
+        createNamespace(view.index) >> handleProgress(strategy, view.projectionId)
+      }
+      .flatMap { progress =>
+        indexingSource(view.projectRef, progress.offset, view.resourceTag)
+          .evalMapValue(BlazegraphIndexingStreamEntry.fromEventExchange(_))
+          .evalMapFilterValue {
```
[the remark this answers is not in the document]
umbreak: Having it combined in one stream evaluation it does not make it easier to follow, but actually more difficult (since you have to map to a Some, even though that part of the evaluation never discards).

I agree that it is more difficult to follow because there is part of the logic in `fromEventExchange` and part in the `IndexingStream` but that's due to the fact that this `fromEventExchange` can be then reused in `CompositeIndexingStream` and `BlazegraphIndexingStream`

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeIndexingStream.scala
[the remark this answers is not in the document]
umbreak: I don't see why this can be very useful...but do you want me to prefix it with `composite_${value}` ?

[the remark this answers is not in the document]
imsdu: The idea is that if there is a problem at the ES/BG level from the name of the index/namespace, it should be easy to trace it back to the plugin and the view

PR #2227, opened 2021-04-06
Restore resource.json for retro-compatibility purposes

### LINKED ISSUE #2135: Restore resource.json for retro-compatibility purposes

imsdu:
Fixes #2135


umbreak (changes requested):
I think we should not do that for resource.json
we should just do a merge with the actual `all-metadata.json` adding the nxv

something like
```scala
aggMetadataCtx
   .map(_.merge(ContextObject(JsonObject("nxv" -> nxv.base))))

```

umbreak:
We don't have to do anything about it. I close it.

PR #2242, opened 2021-04-09
Introduce convention for plugin configuration

### LINKED ISSUE #2074: Introduce a configuration structure convention for plugins to avoid collisions

imsdu:
Fixes #2074


umbreak (changes requested):
I'm not sure if that's exactly what @bogdanromanx had in mind. To me, just to prefix the configuration block with `plugins` is not bringing anything.
We probably can wait until @bogdanromanx to know what was all this about.

imsdu:
The namespace allows to isolate them from other entries so it is valuable as it allows to better identify them and perform checks.

The issue is discussing about a configuration root not so much about the checks.

About the checks, we can imagine that the composite views plugin could change a setting on the ES / BG plugin such as changing a timeout / a retry strategy.

umbreak:
Ok. I don't see much of a benefit. But if you guys thing that's "nicer", yeah sure.

bogdanromanx (approved):

PR #2244, opened 2021-04-12
Add SSE client

### LINKED ISSUE #2187: CompositeView: Add SSE client for views

imsdu:
Fixes #2187


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/client/DeltaClient.scala
```diff
@@ -1,44 +1,99 @@
...
+    }
+
+    implicit val cred: Option[AuthToken] = token(source)
+
+    def send(request: HttpRequest): Future[HttpResponse] = {
+      client[HttpResponse](request.withCredentials)(IO.pure(_)).runToFuture
+    }
+
+    val uri =
+      source.endpoint / "resources" / source.project.organization.value / source.project.project.value / "events"
```
bogdanromanx, line 68: will this return all the events for a project (incl: resolvers, schemas etc.)?

bogdanromanx (approved):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/client/DeltaClient.scala
```diff
@@ -1,44 +1,99 @@
...
       Get(
         source.endpoint / "projects" / source.project.organization.value / source.project.project.value / "statistics"
       ).addHeader(accept).withCredentials
     client.fromJsonTo[ProjectCount](statisticsEndpoint)
   }
 
+  private def toOffset(id: String): Offset =
+    Try(TimeBasedUUID(UUID.fromString(id))).orElse(Try(Sequence(id.toLong))).getOrElse(NoOffset)
+
+  def events[A: Decoder](source: RemoteProjectSource, offset: Offset): Stream[Task, (Offset, A)] = {
```
umbreak, line 54: What is this `A` in our case. It is `A <: Event` ? I don't think `Decoder` is going to be enough here, since we cannot produce a `Decoder` for subtypes of Event. That's why we have `EventExchange`.
What we need the `events` method to return is a `Envelope[Event]`.
 
We will also need something like:
```scala
def fetchResource(id: Iri, project: ProjectRef, tag: Option[TagLabel]): Option[EventExchangeValue[_, _]] = ???
```

Notice that in the very end, what we need for the composite views is to be able to produce an `IndexingSource` from a `RemoteProjectSource` and an offset.

umbreak (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/client/DeltaClient.scala
umbreak: yes, this endpoint currently returns aggregated events

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/client/DeltaClient.scala
umbreak: Basically we need this same logic [IndexingStream](https://github.com/BlueBrain/nexus/blob/master/delta/sdk-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/views/indexing/IndexingSource.scala#L57) but with the following differences:
- The stream is not coming from a DB query, but from an HTTP query.
- The `toResource` method has to make an HTTP query to fetch the resource, and after that use the `EventExchange` for the conversion

imsdu: It can't really return an `Envelope[Event]` as SSE's don't provide akka-persistence fields.
To decode in this context, we don't need to deserialize the whole event, we just need to extract the id right ? We don't care about other fields, we just want to be able to make a second call to the remote delta instance to get the whole resource.

umbreak: We need the following:
- offset
- instant
- persistenceId
- sequenceNr
- id
- projectRef

If  we know what we want to return, we don't need an `A`, but a specific case class.

imsdu: We really have to go though this Envelope/EventExchange thing for a remote project source ?
When fetching the resource, we can fetch the expanded form and from this, get the graph and whatever we need to put it in blazegraph ?

umbreak: It is not as simple as that, since we need metadata and not metadata, we need some fields to build the correct `Message` and not to say that all logic is well tight together with the `EventExchange` currently.

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeIndexingStream.scala
```diff
@@ -372,15 +372,15 @@ object CompositeIndexingStream {
       indexingController.restart(id, project, Restart(PartialRestart(projections)))
     val remoteProjectCounts: RemoteProjectsCounts = source =>
       deltaClient
-        .projectCount(source)
+        .statistics(source)
```
umbreak, line 375: back to statistics? I thought I changed it in my previous PR to `projectCount`

umbreak (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeIndexingStream.scala
wwajerowicz: I thought that was before Simon made his change, I'll change it back

umbreak (dismissed):

bogdanromanx (dismissed):

umbreak (approved):

bogdanromanx (approved):

Issue #2259, opened 2021-04-14
Converting an empty Graph to compacted jsonld fails


umbreak:
The following fails:
```scala
Graph.empty.copy(rootNode = iri).toCompactedJsonLd(ContextEmpty)
```
with 
```
Caused by: ch.epfl.bluebrain.nexus.delta.rdf.RdfError$ConversionError: Error on the conversion stage 'framing'
Details: null
```

PR #2260, opened 2021-04-14
Added interval restart logic

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2245
Fixes https://github.com/BlueBrain/nexus/issues/2259

### DIFF
--- a/delta/plugins/composite-views/src/main/resources/composite-views.conf
+++ b/delta/plugins/composite-views/src/main/resources/composite-views.conf
@@ -6,8 +6,8 @@ akka.actor {
   }
 
   serialization-bindings {
-    "ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.model.CompositeViewEvent"              = "circeComposite"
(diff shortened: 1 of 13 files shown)


imsdu:
Something mixing things like:
* IO.timeout so that it stops after a while
* scheduler.scheduleAtFixedRate to start it again
would not work ?

umbreak:
Why do we need to do it that way?
I'm working with `F[_]` so `scheduleAtFixedRate` is not going to fly out.

I don't find this implementation more difficult either.

imsdu:
We could just let the timeout and let the remember entities restart it ? Or use in some way the coordinator.restart ?

It is not this addition that is complicated, it is more that that there is an accumulation with what already exists.

umbreak:
I don't see how any of your suggestions is simpler than what I did. Unless I see some concrete example with some benefits, I'm not confident of the change. And I don't want to touch `IndexingStreamCoordinator`, since it is complex enough already. Management of lifecycle and so on happen on the `CompositeIndexingStream` previously and now. So it is consistent.

 Another option is to put everything inside an actor and keep the nextInterval (Option[Instant]) as the actor state. However I would argue that option is more verbose.

PR #2267, opened 2021-04-16
Fixed IT ResourcesSpec failing tests

### LINKED ISSUE #2266: Requests for listings where both after and from parameters are present should be rejected

bogdanromanx:
The change also disables pub-sub notifications for events by tag
and lowers the consistency delay.

Fixed #2266

### DIFF
--- a/delta/app/src/main/resources/application-cassandra.conf
+++ b/delta/app/src/main/resources/application-cassandra.conf
@@ -10,7 +10,7 @@ akka {
         keyspace-autocreate = ${app.database.cassandra.keyspace-autocreate}
         tables-autocreate = ${app.database.cassandra.tables-autocreate}
         keyspace = ${app.database.cassandra.keyspace}
-        pubsub-notification = on
+        pubsub-notification = off
       }
       snapshot {
         keyspace-autocreate = ${app.database.cassandra.keyspace-autocreate}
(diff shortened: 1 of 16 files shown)


delta/app/src/main/resources/application-cassandra.conf
```diff
@@ -23,7 +23,7 @@ akka {
       events-by-tag {
         first-time-bucket = "20181213T00:00"
         # This delay helps to order events. Setting this to anything lower than 2s is highly discouraged.
-        eventual-consistency-delay = 10s
+        eventual-consistency-delay = 5s
```
imsdu, line 26: This timeout seem low compared to the `tag-write-timeout` and the evaluation timeout, we may miss some events

umbreak (approved):

delta/app/src/main/resources/application-cassandra.conf
```diff
@@ -10,7 +10,7 @@ akka {
         keyspace-autocreate = ${app.database.cassandra.keyspace-autocreate}
         tables-autocreate = ${app.database.cassandra.tables-autocreate}
         keyspace = ${app.database.cassandra.keyspace}
-        pubsub-notification = on
+        pubsub-notification = off
```
imsdu, line 13: According to the doc:

```
# Setting pubsub-notification to "off" will disable the journal sending these announcements.
    # When enabled with `on` it will throttle the number of notfication messages per tag to at most 10 per second.
    # Alternatively a duration can be defined, such as pubsub-notification=300ms, which will be throttling interval
    # for sending the notifications.
```
It may better to set a duration there as 10 times per second is a little aggressive ?

delta/app/src/main/resources/application-cassandra.conf
bogdanromanx: I will introduce it in a subsequent PR.

PR #2272, opened 2021-04-19
Use project base as fallback

### LINKED ISSUE #2269: Use project base as a fallback when creating resources
When no `@base` is provided for a resource, the one defined in the project should be picked up as a fallback to properly expand it.

imsdu:
Fixes #2269


delta/rdf/src/test/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/ExpandedJsonLdSpec.scala
```diff
@@ -24,6 +27,12 @@ class ExpandedJsonLdSpec extends AnyWordSpecLike with Matchers with Fixtures {
       ExpandedJsonLd(compacted).accepted shouldEqual ExpandedJsonLd.expanded(expectedExpanded).rightValue
     }
 
+    "be constructed successfully with a base defined in JsonLdOptions" in {
```
umbreak, line 30: I'm not sure this needs to be tested at all, since it is part of the json-ld spec and the underlying library. But well I guess it doesn't hurt

umbreak (approved):

wwajerowicz (approved):

bogdanromanx (approved):

PR #2274, opened 2021-04-19
Apply default settings to ES views and ignore scope init error when org/proj are deprecated

### LINKED ISSUE #2261: Elasticsearch views should include default settings by default
If no settings are provided, the default settings should be applied to an elasticsearch view

### LINKED ISSUE #2270: Scope initialization should not raise errors when the org/project is deprecated

imsdu:
Fixes #2261 
Fixes #2270


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewValue.scala
```diff
@@ -49,7 +49,7 @@ object ElasticSearchViewValue {
       includeMetadata: Boolean = false,
       includeDeprecated: Boolean = false,
       mapping: JsonObject,
-      settings: Option[JsonObject] = None,
+      settings: JsonObject,
```
umbreak, line 52: Without a default value? I think we use the `ElasticSearchViewValue` to generate the `deriveConfigJsonLdDecoder`. If there is no default value it means the client to always provide the `settings` field. Or am I missing something?

umbreak: I see you added a `applyDefaultSettings` method to hack around this. However you shouldn't do it this way. Notice that `ElasticSearchViewValue` is just an intermediate representation used to easily transform a client json-ld to the target case class `ElasticSearchView`.

Here, what you should do instead is keep `settings: Option[JsonObject] = None` and then just have a method:
```scala
def settingsOrDefault: JsonObject = ???
```

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViewJsonLdSourceDecoder.scala
```diff
@@ -23,15 +23,31 @@ class ElasticSearchViewJsonLdSourceDecoder private (
...
+        project,
+        iri,
+        mapJsonToString(s)
+      )
+    }
+
+  /**
+    * Apply default settings if nothing has been provided
+    */
+  private def applyDefaultSettings(source: Json): UIO[Json] =
```
umbreak, line 43: Don't do it here

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewValue.scala
umbreak: That's how we did it for [StorageFields](https://github.com/BlueBrain/nexus/blob/master/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageFields.scala#L86) which is sort of the same use case

umbreak (changes requested):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewValue.scala
imsdu: `ElasticSearchViewValue` is not an intermediate representation like `StorageFields`, that is why I proceeded this way as `ElasticSearchViewValue` is part of ES events.

umbreak: Still, it is not what one interact with from the SDK. You interact with `ElasticSearchView`. But I see why would be useful to have it on the event...

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViewJsonLdSourceDecoder.scala
```diff
@@ -23,15 +23,31 @@ class ElasticSearchViewJsonLdSourceDecoder private (
...
+    source.asObject match {
+      case Some(o) if !o.contains("settings") =>
```
umbreak, line 45: Working at the compacted JSON level is wrong, since this key can very well be `https://bluebrain.github.io/nexus/vocabulary/settings` and it should still work. Or it can be a curie or an alias and still work.

But I know that in few other places we have this assumption too. That's why this approach is not desirable and I suggested the other

umbreak (changes requested):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewValue.scala
imsdu: I also agree that we should have a consistent approach between resolvers/storages/views.

But it goes beyond this PR no ?

umbreak: I think for now it is better the other approach, where you keep settings as Option in `ElasticSearchViewValue` since it is the safest.

bogdanromanx: I'm happy with both options. I do see your point @umbreak, but I don't see any large implications going one way or another.

bogdanromanx (dismissed):

bogdanromanx (approved):
Looks more complicated than the previous impl, but ok.

umbreak (approved):

PR #2281, opened 2021-04-20
Source endpoint should only display plain Json

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2280


wwajerowicz (approved):

bogdanromanx (changes requested):
Why do we want to restrict other formats for the source?

umbreak:
Because when you want to retrieve the source, you don't want any json-ld magic going on. This endpoint was added for the people to fetch exactly what they posted and nothing else

bogdanromanx:
I think it's quite useful to be able to return the source in n-triples format, for example. I don't see any benefits in restricting this use. What benefits do you see?

umbreak:
And btw, that's how it is currently in v.1.4.x. You can never get an expanded/n-triples/etc version of the `/source` endpoint.

bogdanromanx:
I'm well aware of that, but this PR only removes functionality for no clear reason. Is there a reason we don't want to allow the serialization of the resource in all supported formats?

umbreak:
> I think it's quite useful to be able to return the source in n-triples format, for example. I don't see any benefits in restricting this use. What benefits do you see?

That is not the goal of the `/source` endpoint.  It is just to get back what you posted. That's why is called source. If you posted an expanded payload, you get back an expanded payload. If you posted a compacted, you get back a compacted.

umbreak:
> I'm well aware of that, but this PR only removes functionality for no clear reason. Is there a reason we don't want to allow the serialization of the resource in all supported formats?
I thought we are going to add functionality driven by requirements, not because we "think" it is nice.

umbreak:
> /source is a sub-resource of all resources in the system that represent the information provided by the client. The serialization of the resource should not play any role. I don't see a reason to restrict getting the source in a different format, when the functionality is well supported right now.

I said why in the previous comment already. Another reason is because the documentation says so: [Fetch a resource original payload](https://bluebrainnexus.io/docs/delta/api/current/kg-resources-api.html#fetch-a-resource-original-payload)

bogdanromanx:
> I thought we are going to add functionality driven by requirements, not because we "think" it is nice.
Indeed, we want to add functionality based on requirements, but this functionality is already there, so why remove it?

> Yes. One being the fact that if we allow multiple response formats, the JSON-LD algorithm will always kick in and it will potentially be different response than what the client posted (the source)
If the source is requested in a different format, there's a client assumption that the system will perform a transformation. The default should still be application/json or application/ld+json.

bogdanromanx:
> I said why in the previous comment already. Another reason is because the documentation says so: Fetch a resource original payload. If serialization kicks in, the payload is not going to be anymore the original one.

I think there's a slight difference in understanding what a resource is and its representation. The resource represents the information while the representation can vary.

bogdanromanx (approved):

Issue #2108, opened 2021-03-01
CompositeView: expose indexing statistics


bogdanromanx:
In case of sources, the statistics and the offset sub-resources will be computed by collecting all the projection statistics for that source and apply the max for each of the fields.

PR #2286, opened 2021-04-21
Added composite views statistics

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2108


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeStatisticsCollection.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+    * Constructor helper from a collection of [[CompositeStatisticsCollection]]
+    */
+  // TODO: It does not make any sense to just take maximums/minimums and the resulting summary it is just misleading.
+  // Summary should be removed from the API response and statistics endpoint should return a SearchResults[CompositeStatistics]
+  def apply(values: List[CompositeStatistics]): CompositeStatisticsCollection = {
+    val (head, tail) =
+      values.headOption.fold(ProgressStatistics.empty -> List.empty[CompositeStatistics])(_.value -> values.tail)
+    val summary      = tail.foldLeft(head) { case (acc, CompositeStatistics(_, _, stats)) =>
```
imsdu, line 31: We could get rid of the summary no ? It does not bring useful information as the max does not mean much for most of the fields

imsdu (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeStatisticsCollection.scala
umbreak: I agree. It is there just for backwards compatibility. But it is something we can discuss tomorrow morning. If it were for me I would remove it, as I explain in the TODO

bogdanromanx (approved):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeStatisticsCollection.scala
umbreak: Fixed

bogdanromanx (approved):

wwajerowicz (approved):

PR #2308, opened 2021-04-26
Fixed a series of IT failures

bogdanromanx:


tests/src/test/resources/kg/views/composite/composite-view.json
```diff
@@ -1,5 +1,5 @@
 {
-  "@type": ["Beta","CompositeView"],
+  "@type": ["CompositeView"],
```
umbreak, line 2: it can just be `"@type": "CompositeView"`

umbreak (changes requested):

delta/app/src/main/resources/app.conf
```diff
@@ -254,7 +254,7 @@ app {
       # the default number of results if none is specified
       default-size = 30
       # the maximum number of results if none is specified
-      size-limit = 100
+      size-limit = 1000
```
imsdu, line 257: Why putting a value that large by default ? v1.4 only set 100

bogdanromanx: It was discussed in a call... Fusion uses `1000`. I just checked, v1.4 default was `100` so I think the service silently ignored larger inputs.
@umbreak, @dhaneshnm thoughts on what we should about this?

umbreak: In the deployment configurations we have it much larger. 1000 as limit I find alright.

umbreak: I don't see any problem with a max of 1000

imsdu: Can it be a problem for blazegraph ? For elasticsearch, it should be ok but it could put more (too much ?) pressure on BG ?
If we change it, we should not forget to document it in the changes

bogdanromanx: @imsdu, do you have any concerns with a max value of 1000?

bogdanromanx: > Can it be a problem for blazegraph ?

Blazegraph queries do not use this pagination... the start and limits are set in the query itself.

umbreak: We can always override this value for blazegraph if we want to.

bogdanromanx: so, ok to merge?

umbreak (approved):

imsdu (approved):

PR #2315, opened 2021-04-27
Use only Circe encoding for events

umbreak:
This is a work on direction to support uuids on SSEs events. Before that, I've simplified the dealing with SSEs.

SSEs response cannot be turned to expanded, n-triples or anything else. Since that's the case, we did not need all those jsonLdEncoding.

### DIFF
--- a/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsRoutes.scala
+++ b/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsRoutes.scala
@@ -24,7 +24,6 @@ import ch.epfl.bluebrain.nexus.delta.sdk.directives.AuthDirectives
 import ch.epfl.bluebrain.nexus.delta.sdk.directives.DeltaDirectives._
 import ch.epfl.bluebrain.nexus.delta.sdk.instances.OffsetInstances._
 import ch.epfl.bluebrain.nexus.delta.sdk.marshalling.RdfMarshalling
-import ch.epfl.bluebrain.nexus.delta.sdk.model.IdSegment.StringSegment
 import ch.epfl.bluebrain.nexus.delta.sdk.model.acls.AclAddress
 import ch.epfl.bluebrain.nexus.delta.sdk.model.identities.Caller
 import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectRef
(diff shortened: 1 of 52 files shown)


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsRoutes.scala
```diff
@@ -275,36 +274,30 @@ final class ElasticSearchViewsRoutes(
 
   private def genericResourcesRoutes(implicit caller: Caller): Route =
     pathPrefix("resources") {
-      projectRef(projects).apply {
-        case ProjectRef(_, project) if project.value == "events" => reject()
```
bogdanromanx, line 279: we don't need this guard anymore?

delta/plugins/storage/src/test/resources/file/events/file-created.json
```diff
@@ -13,7 +13,7 @@
   "_rev" : 1,
   "_storage": {
     "@id": "https://bluebrain.github.io/nexus/vocabulary/disk-storage",
-    "@type": "DiskStorage",
+    "@type": "https://bluebrain.github.io/nexus/vocabulary/DiskStorage",
```
bogdanromanx, line 16: any reason why this is no longer compacted?

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsRoutes.scala
umbreak: No. Look at the PR on `UriDirectives.label` and `UriDirectives.idSegment`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToSse.scala
```diff
@@ -0,0 +1,81 @@
...
+      io: IO[Response[E], Stream[Task, Envelope[JsonValue]]]
+  )(implicit s: Scheduler, jo: JsonKeyOrdering, cr: RemoteContextResolution): ResponseToSse =
+    new ResponseToSse {
+
+      private def toSse(envelope: Envelope[JsonValue]) = {
+        val json       = envelope.event.encoder(envelope.event.value)
+        val id: String = envelope.offset match {
+          case TimeBasedUUID(value) => value.toString
+          case Sequence(value)      => value.toString
+          case NoOffset             => "-1"
```
bogdanromanx, line 43: I think it would be better to have this:

```
val idOpt: Option[String] = envelope.offset match {
  case TimeBasedUUID(value) => Some(value.toString)
  case Sequence(value)      => Some(value.toString)
  case NoOffset             => None
}
ServerSentEvent(defaultPrinter.print(json.sort), Some(envelope.eventType), idOpt)
```

bogdanromanx (dismissed):

delta/plugins/storage/src/test/resources/file/events/file-created.json
umbreak: Yep. It is just like that.

bogdanromanx: we're not constrained by compatibility with v1.4 because this field didn't exist... but yeah, we can do the api review and see.

bogdanromanx (approved):

imsdu (approved):

PR #2316, opened 2021-04-27
Remove all ids with space in resource sources

imsdu:


delta/migration/src/main/scala/ch/epfl/bluebrain/nexus/migration/v1_4/SourceSanitizer.scala
```diff
@@ -84,13 +84,11 @@ object SourceSanitizer {
     modified.deepDropNullValues
   }
 
-  private val nonAvailable = "not available".asJson
-
   val dropNotAvailableIds: Json => Json = Plated.transform[Json] { j =>
     j.asObject match {
-      case Some(o) if o("@id").contains(nonAvailable) =>
+      case Some(o) if o("@id").exists(_.asString.exists(_.contains(" "))) =>
```
bogdanromanx, line 89: would it be better that instead of dropping these ids to replace them with `%20`?

bogdanromanx (approved):

delta/migration/src/main/scala/ch/epfl/bluebrain/nexus/migration/v1_4/SourceSanitizer.scala
imsdu: Some of the values are "not available", "no files", "Schaffer Collateral Cell", "interneuron with soma in SR" and from the resources I looked at in details, the problem seems to be fixed  in later revisions of the resources by being removed.

bogdanromanx (approved):

Issue #2329, opened 2021-04-29
Added back nxv to api mappings


umbreak:
This is done in order to support using predicates on listings filtering. For example:
`v1/resources/{org}/{project}?type=nxv:View`

Issue #2327, opened 2021-04-29
Metadata and data graphs not well split when indexing on ES and Blazegraph


umbreak:
for example:
-  @type should be part of both metadata and data.
- the metadata fields like _projectUuids and _organizationUuid are now part of the resource graph and not the metadata graph

Issue #2321, opened 2021-04-28
@type field during elasticsearch indexing should be expanded


umbreak:
When views/storages/resolvers gets indexed into elasticsearch, their types are being indexed compacted, while they should be indexed as expanded.

PR #2330, opened 2021-04-29
Fixed data/metadata graph and added nxv to prefix mappings

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2327
Fixes https://github.com/BlueBrain/nexus/issues/2321
FIxes https://github.com/BlueBrain/nexus/issues/2329


delta/plugins/elasticsearch/src/test/resources/routes/elasticsearch-view-list-response.json
```diff
@@ -2,7 +2,7 @@
   "@context" : [
     "https://bluebrain.github.io/nexus/contexts/metadata.json",
     "https://bluebrain.github.io/nexus/contexts/search.json",
-    "https://bluebrain.github.io/nexus/contexts/all-metadata.json"
+    "https://bluebrain.github.io/nexus/contexts/listings-metadata.json"
```
bogdanromanx, line 5: can we call this `search-metadata.json` instead?

bogdanromanx (changes requested):
The context rename needs to be propagated also to the integration tests.

wwajerowicz (dismissed):

delta/plugins/elasticsearch/src/test/resources/routes/elasticsearch-view-list-response.json
umbreak: Is not a search, is for listings. Search is when using the endpoint `/v1/views/{org}/{proj}/{id}/_search`

bogdanromanx: we already have the search context that is used for listings. I think it makes sense.

bogdanromanx (changes requested):
The modifications to the Integration tests are still missing.

umbreak:
I'm not working on the integration tests right now, neither I have the powerful computer enough to orchestrate all of that. I'd be happy to make the changes tot he integration tests when there is a way for me to run them locally without plenty of timeouts.

bogdanromanx:
> I'm not working on the integration tests right now, neither I have the powerful computer enough to orchestrate all of that. I'd be happy to make the changes tot he integration tests when there is a way for me to run them locally without plenty of timeouts.

It's only a matter of doing a string replace for `all-metadata.json` -> `search-metadata.json`

bogdanromanx (approved):

PR #2331, opened 2021-04-29
Retry on timeout while indexing 

### LINKED ISSUE #2284: When indexing, a item should be retried when a timeout occurs
When there is a timeout happening on the Cassandra/Akka/elasticsearch/blazegraph part, the event should be retried and not marked as an error

wwajerowicz:
Fixes #2284


bogdanromanx (approved):

umbreak:
Since yesterday we commented on the meeting me and @bogdanromanx it would be also nice to have a retry mechanism passed along to the ShardedAggregate. A different one than a `never`...

delta/app/src/main/resources/app.conf
```diff
@@ -242,6 +241,7 @@ app {
       ask-timeout = 25 seconds
       evaluation-max-duration = 14 seconds
       stash-size = 10
+      retry = ${app.defaults.never-retry-strategy}
```
umbreak, line 244: Could we have the default as `constant-retry-strategy` ?

umbreak (changes requested):

delta/app/src/main/resources/app.conf
wwajerowicz: But then we'll have double retries: once in `ShardedAggregate` and once in `IndexingSource`?

umbreak: Yes. I don't think that's a problem since IndexingSource should retry for a long time anyway.
And if we cannot do that, what's the point of having the possibility to retry here at all?

umbreak (approved):

PR #2332, opened 2021-04-29
Update resolvers sub-resource endpoint to handle plugin-related resource types

### LINKED ISSUE #1895: Update resolvers sub-resource endpoint to handle plugin-related resource types

imsdu:
Fixes #1895 

* Based on ReferenceExchange
* We loop only once though all resolvers when resolving a resource
* I removed the resource type from the resolver reporting as it was useless and only brought noise
* I removed some dead code


umbreak (dismissed):
We are loosing the capability to get the resource by schema, but apparently we are not using it so it should be fine.

bogdanromanx (approved):
Needs merge conflicts to be resolved, though...

wwajerowicz (dismissed):

wwajerowicz (approved):

Issue #2348, opened 2021-04-30
SSEs _project and _schemaProjects field are the labels while it should be the project id


umbreak:
Adapt all resources SSEs to contain the `_project` and `schemaProject` as the id. This should be obtained by
```scala
ResourceUris.project(projectRef).accessUriShortForm
```

Issue #2347, opened 2021-04-30
_constrainedBy field sometimes has the revision, sometimes it doesn't


umbreak:
on SSEs the `_constraiedBy` field contain the revision (for resources) while for api responses it doesn't

Issue #2345, opened 2021-04-30
Listings with ES do not validate unexpected query parameters returns an elasticsearch error


umbreak:
For elasticsearch queries one can provide any query parameter and they'll be forwarded to elasticsearch API. However we should not support that on listings because then the es error is forwarded back tot he listings response

Issue #2344, opened 2021-04-30
Check the lowercase in ES listings


umbreak:
For example:
- /v1/resolvers/{org}/{proj}?type=cross does not match anything
- /v1/resolvers/{org}/{proj}?type=Cross matches one (if present)

PR #2362, opened 2021-05-04
Fixed bugs discovered during API review

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2348
Fixes https://github.com/BlueBrain/nexus/issues/2347
Fixes https://github.com/BlueBrain/nexus/issues/2345
Fixes https://github.com/BlueBrain/nexus/issues/2340
Fixes https://github.com/BlueBrain/nexus/issues/2325
Fixes https://github.com/BlueBrain/nexus/issues/2344
Fixes https://github.com/BlueBrain/nexus/issues/2341
Fixes https://github.com/BlueBrain/nexus/issues/2364


delta/plugins/elasticsearch/src/main/resources/defaults/default-settings.json
```diff
@@ -5,26 +5,13 @@
...
     "filter": {
-      "my_multiplexer": {
-        "type": "multiplexer",
-        "filters": [
-          "lowercase",
-          "uuid_type_4",
-          "word_delimiter_nexus"
-        ]
-      },
-      "uuid_type_4": {
```
bogdanromanx, line 21: we don't need this anymore?

bogdanromanx (changes requested):

delta/plugins/elasticsearch/src/main/resources/defaults/default-settings.json
bogdanromanx: we added that to be able to search resources by uuids in fusion admin

umbreak: No, `word_delimiter_nexus` is splitting as expected for UUIDs already.

bogdanromanx (approved):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
```diff
@@ -187,8 +187,10 @@ class AclsRoutes(identities: Identities, acls: Acls)(implicit
                     acls
                       .list(addressFilter)
                       .map { aclCol =>
-                        val filtered = aclCol.filterByPermission(caller.identities, aclsPermissions.read)
-                        SearchResults(filtered.value.size.toLong, filtered.value.values.toSeq)
+                        val accessibleAcls = aclCol.filterByPermission(caller.identities, aclsPermissions.read)
```
imsdu, line 190: We could add a test for this one so that it does not bite again

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/AclsRoutes.scala
umbreak: All those methods are already tested. The `++`, the `filterByPermission`, the `filter`.
We could add a regression test in the integration tests, but I won't do it in this PR

PR #2365, opened 2021-05-04
Create blazegraph namespace on view creation

### LINKED ISSUE #2320: Create blazegraph namespace on blazegraph view creation
Elasticsearch index gets created when elasticsearch view gets created (in order to check mappings payload) but blazegraph namespace does not.

It would be useful to create the namespace synchronously on view creation, to avoid errors when the client immediately after the project gets created attempts to query the default blazegraph namespace

imsdu:
Fixes #2320


umbreak (changes requested):
Wy this logic is not on the state machine, as it is for ElasticSearch? it seems more consistent.
See `ElasticSearchViews.evaluate`

imsdu:
Because it is not part of the validation of the a blazegraph view.

umbreak:
It doesn't seem as a reason enough. Evaluation doesn't need to be related only to evaluations

imsdu:
This way, we also make sure that the view has been persisted before creating it in Blazegraph.
And it is similar to what we do to create a project.

umbreak:
> This way, we also make sure that the view has been persisted before creating it in Blazegraph.
Why? Here the order doesn't matter at all. Even more, if for some reason we cannot create the index, we would rather have not created the view yet.


I keep finding it not consistent with ElasticSearch. Having a consistent way to doing things (even more amngs views where the logic is pretty similar) helps us to find where a logic exists.

bogdanromanx (approved):

bogdanromanx:
I think this is fine... the namespace creation could time out resulting in a namespace created but the event not persisted otherwise.

umbreak:
It is better a namespace created (which is not a problem at all) than an event created when the blazegraph service is down altogether.

Besides, as I said, it is how elasticsearch logic works. Consistency on the logic is important

imsdu:
We have this approach in ES because we want to valid the mapping and we have no mean to do this without actually creating an index so it is more a workaround than a design choice.
Having the same approach as creating default views/storages after creating a project is better as we are sure that the event is persisted before doing any other action on the system.

umbreak (commented):
As you like

umbreak (approved):

PR #2376, opened 2021-05-04
Multiple CompositeViews fixes

### LINKED ISSUE #2335: CompositeView statistics are incorrect for remote sources and cross project sources
The statistics are collected from the view project instead of the correct location.

### LINKED ISSUE #2336: CompositeView event decoding for remote sources fails
Errors like the following are presented in the service logs:

```
io.circe.DecodingFailure$$anon$2: Attempt to decode value on failed cursor: DownField(resourceId)
2021-04-29 08:55:12 ERROR c.e.b.n.d.p.c.client.DeltaClient - Failed to decode sse event 'ServerSentEvent({"@context":"https://bluebrain.github.io/nexus/contexts/metadata.json","@type":"ResourceCreated","_constrainedBy":"https://bluebrain.github.io/nexus/schemas/unconstrained.json?rev=1","_instant":"2021-04-29T08:53:28.762Z","_organizationUuid":"6c24ef71-a51a-45a9-a1d6-9559cf2954f3","_project":"6nuticg7gvwg4ar/songs","_projectUuid":"f15a8d55-39c4-4d46-8913-0f4b816e28e7","_resourceId":"https://music.example.com/absolution/1","_rev":1,"_schemaProject":"6nuticg7gvwg4ar/songs","_source":{"@context":"https://music.example.com/context","@id":"https://music.example.com/absolution/1","@type":"Song","length":252,"number":1,"on":"absolution","title":"Apocalypse Please"},"_subject":"http://delta:8080/v1/realms/compositevgvoaorhwtobacij/users/mnpfuvjfnxepvwki","_types":["https://music.example.com/Song"]},Some(ResourceCreated),Some(5dec60b0-a8c8-11eb-9e79-81778bf135de),None)'
```

### LINKED ISSUE #2372: Composite view common blazegraph sparql routes collides with blazegraph sparql routes
add `.rejectOn[ViewNotFound]` on both colliding routes

### LINKED ISSUE #2373: CompositeViews ES projection query endpoints override the payload sort parameters

### LINKED ISSUE #2374: CompositeView ES projections merge the source json to the result document

bogdanromanx:
Fixes #2374
Fixes #2373
Fixes #2372
Fixes #2336
Fixes #2335


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchIndexingStreamEntry.scala
```diff
@@ -78,14 +79,21 @@ final case class ElasticSearchIndexingStreamEntry(
   def containsTypes[A](resourceTypes: Set[Iri]): Boolean =
     resourceTypes.isEmpty || resourceTypes.intersect(resource.types).nonEmpty
 
-  private def toDocument(includeMetadata: Boolean, sourceAsText: Boolean, context: ContextValue): Task[Json] = {
+  private def toDocument(
+      includeMetadata: Boolean,
+      sourceAsText: Boolean,
+      context: ContextValue,
+      mergeSource: Boolean
```
umbreak, line 86: Why is this here? When building the `ElasticSearchIndexingStreamEntry` you can just create the `IndexingData` with an empty source (check the apply methods for it) and then we don't need this, nor yet another if/else

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/ElasticSearchQuery.scala
```diff
@@ -27,15 +27,13 @@ trait ElasticSearchQuery {
     * @param project      the project where the view exists
     * @param query        the elasticsearch query to run
     * @param qp           the extra query parameters for the elasticsearch index
-    * @param sort         the sorting configuration
     */
   def query(
```
umbreak, line 31: It is alright that we are not supporting SortList on elasticsearch queries, but then we have to be consistent and do the same in `ElasticSearchViewsQuery` and `ElasticSerchRoutes`

umbreak (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/ElasticSearchQuery.scala
bogdanromanx: what do you think about the `qp` should I remove these as well?

umbreak: No. That's useful and part of the elasticsearch API

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchIndexingStreamEntry.scala
umbreak: I don't understand. You added this for composite views only. sourceAsText does not exist for composite views, so there is nothing to do about that in this case, no?

bogdanromanx:
```
/**
    * An ElasticSearch projection for [[CompositeView]].
    */
  final case class ElasticSearchProjection(
      id: Iri,
      uuid: UUID,
      query: SparqlConstructQuery,
      resourceSchemas: Set[Iri],
      resourceTypes: Set[Iri],
      resourceTag: Option[TagLabel],
      includeMetadata: Boolean,
      includeDeprecated: Boolean,
      permission: Permission,
      sourceAsText: Boolean = false,
      mapping: JsonObject,
      settings: Option[JsonObject] = None,
      context: ContextObject
  )
```

wwajerowicz (approved):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchIndexingStreamEntry.scala
umbreak: Thats wrong. Check the v1.4.x docs. And sourceAsText doesn't make any sense for compositeviews

umbreak: https://bluebrainnexus.io/docs/delta/api/current/views/composite-view-api.html#elasticsearchprojection

bogdanromanx: It's not wrong, we decided in the new design to make it consistent with the ElasticsearchView config. Just like the rest of the metadata, while it may not necessarily be useful, we want to offer the option and be consistent.

bogdanromanx: Either way, removing that field is not in the scope of this PR.

umbreak: Ok, then we'll do it right after. 
And yes it is wrong to have sourceAsText in composite views. What you want to index in an elasticsearach view is the source, that's all you have. That's why there is a sourceAsText true|false. 

In composite views you don't have a source, but you have a query result which is transformed to json. There is no room for consistency since an elastic search view and an elastic search projection are not the same thing.

PR #2380, opened 2021-05-05
Default es mapping/settings should not be serialized with the event to make evolution easier

### LINKED ISSUE #2379: Default es mapping/settings should not be serialized with the event to make their evolution easier

imsdu:
Fixes #2379


bogdanromanx (approved):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViewJsonLdSourceDecoder.scala
```diff
@@ -95,23 +94,21 @@ object ElasticSearchViewJsonLdSourceDecoder {
...
-        )
-      }
+      IndexingElasticSearchViewValue(
+        resourceSchemas = i.resourceSchemas,
+        resourceTypes = i.resourceTypes,
+        resourceTag = i.resourceTag,
+        sourceAsText = i.sourceAsText,
+        includeMetadata = i.includeMetadata,
+        includeDeprecated = i.includeDeprecated,
+        mapping = Some(i.mapping),
```
umbreak, line 106: Tis means that the `mapping` field for the API client it is still mandatory right? I guess it is ok.

imsdu: Yep, it is still mandatory

PR #2389, opened 2021-05-05
Remove sourceAsText from CompositeViews

### LINKED ISSUE #2381: ElasticSearchProjections in CompositeViews should not include "sourceAsText" option

wwajerowicz:
Fixes #2381


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeIndexingStream.scala
```diff
@@ -215,7 +215,7 @@ final class CompositeIndexingStream(
       val index = idx(projection, view)
       if (deleteCandidate) esRes.delete(index).map(Some.apply)
       else
-        esRes.index(index, projection.includeMetadata, projection.sourceAsText, projection.context, mergeSource = false)
```
bogdanromanx, line 218: the IndexingData also needs to be modified to have the source as `Json.obj()`, otherwise in composite views the source will be automatically merged to the resulting document.

bogdanromanx (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeIndexingStream.scala
```diff
@@ -215,7 +215,7 @@ final class CompositeIndexingStream(
...
+        esRes.index(index, projection.includeMetadata, false, projection.context)
```
umbreak, line 218: can you add `sourceAsText = false` so it is clearer?

umbreak (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeIndexingStream.scala
umbreak: I think it is easier if we just change the following [here](https://github.com/BlueBrain/nexus/blob/master/delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphIndexingStreamEntry.scala#L87)
```scala
    IndexingData(exchangedValue)
      .map(indexingData => BlazegraphIndexingStreamEntry(indexingData.copy(source = Json.obj())))
```

bogdanromanx: The behaviour is very specific to CompositeViews... I would not change it in the blazegraph plugin.

umbreak: Or `IndexingData` can have a method 
```scala
def discardSource: IndexingData = copy(source = Json.obj())
```

And then you can call that one on `BlazegraphIndexingStreamEntry`.

umbreak: > The behaviour is very specific to CompositeViews... I would not change it in the blazegraph plugin.
Yeah, alright. The @wwajerowicz the change is on line CompositeIndexingStream:214
```scala
val esRes = ElasticSearchIndexingStreamEntry(resource.discardSource)
```

umbreak (approved):

bogdanromanx (approved):

PR #2395, opened 2021-05-05
Updated integration tests to match latest service changes

bogdanromanx:


wwajerowicz (dismissed):

imsdu (approved):

tests/src/test/resources/kg/files/linking-notfound.json
```diff
@@ -1,6 +1,6 @@
 {
   "@context" : "https://bluebrain.github.io/nexus/contexts/error.json",
-  "@type" : "ResourceNotFound",
+  "@type" : "FileNotFound",
```
umbreak, line 3: How comes that this is still a `FileNotFound` and it works. I thought I've changed all those for `ResourceNotFound`?

umbreak (changes requested):

tests/src/test/resources/kg/files/linking-notfound.json
umbreak: Yes please. Just the JSON-LD representation has to be changed. I probably forgot about it.

umbreak (approved):

PR #2400, opened 2021-05-06
Keep original source in ES document when present

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2396

### DIFF
--- a/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchIndexingStreamEntry.scala
+++ b/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchIndexingStreamEntry.scala
@@ -2,7 +2,7 @@ package ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.indexing
 
 import ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.client.{ElasticSearchBulk, IndexLabel}
 import ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.model.contexts
-import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.{BNode, Iri}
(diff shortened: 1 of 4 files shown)


imsdu (approved):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchIndexingStreamEntry.scala
```diff
@@ -87,9 +87,19 @@ final case class ElasticSearchIndexingStreamEntry(
...
+        .toCompactedJsonLd(context)
+        .map(_.obj.asJson)
+    else if (resource.source.isEmpty())
+      graph
+        .toCompactedJsonLd(context)
+        .map(_.obj.asJson)
     else
-      graph.toCompactedJsonLd(context).map(ld => mergeJsonLd(resource.source, ld.json).removeAllKeys(keywords.context))
+      (graph -- graph.rootTypesGraph)
+        .replaceRootNode(BNode.random) // This is done to get rid of the @id in order to avoid overriding the source @id
```
bogdanromanx, line 100: what if the source did not contain an `@id`?

wwajerowicz (approved):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchIndexingStreamEntry.scala
bogdanromanx, line 100: what if the source does not contain an @id?

bogdanromanx (changes requested):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchIndexingStreamEntry.scala
umbreak: Then there is no @id. It is up to the client to decide that

bogdanromanx (approved):

PR #2401, opened 2021-05-06
Expose the effectiveApiMappings as a separate metadata field

### LINKED ISSUE #2399: Project API mappings show both user defined and default ones without means to discriminate between the two

bogdanromanx:
Fixes #2399

### DIFF
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/ProjectsModule.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/ProjectsModule.scala
@@ -66,12 +66,14 @@ object ProjectsModule extends ModuleDef {
         projects: Projects,
         projectsCounts: ProjectsCounts,
         baseUri: BaseUri,
+        mappings: ApiMappingsCollection,
         s: Scheduler,
         cr: RemoteContextResolution @Id("aggregate"),
         ordering: JsonKeyOrdering
     ) =>
       new ProjectsRoutes(identities, acls, projects, projectsCounts)(
         baseUri,
+        mappings.merge,
         config.projects.pagination,
         s,
         cr,
@@ -93,6 +95,8 @@ object ProjectsModule extends ModuleDef {
 
   many[PriorityRoute].add { (route: ProjectsRoutes) => PriorityRoute(pluginsMaxPriority + 7, route.routes) }
 
-  make[ProjectEventExchange]
+  make[ProjectEventExchange].from { (projects: Projects, base: BaseUri, mappings: ApiMappingsCollection) =>
+    new ProjectEventExchange(projects)(base, mappings.merge)
+  }
   many[EventExchange].ref[ProjectEventExchange]
 }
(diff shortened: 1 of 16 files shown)


delta/sdk/src/main/resources/contexts/projects-metadata.json
```diff
@@ -4,6 +4,18 @@
     "_uuid": "https://bluebrain.github.io/nexus/vocabulary/uuid",
     "_organizationUuid": "https://bluebrain.github.io/nexus/vocabulary/organizationUuid",
     "_organizationLabel": "https://bluebrain.github.io/nexus/vocabulary/organizationLabel",
+    "_effectiveApiMappings": {
+      "@id": "https://bluebrain.github.io/nexus/vocabulary/effectiveApiMappings",
+      "@container": "@set"
+    },
+    "namespace": {
```
umbreak, line 11: it is metadata, it must start by `_`

delta/sdk/src/main/resources/contexts/projects-metadata.json
```diff
@@ -4,6 +4,18 @@
...
+      "@id": "https://bluebrain.github.io/nexus/vocabulary/namespace",
+      "@type": "@id"
+    },
+    "prefix": {
+      "@id": "https://bluebrain.github.io/nexus/vocabulary/prefix",
```
umbreak, line 16: it is metadata, it must start by `_`
The value of prefix is not an id, so `@type: @id` is wrong

umbreak (changes requested):

umbreak (changes requested):
It is missing the elasticsearch mappings for the `_effectiveApiMappings `

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/ProjectsModule.scala
```diff
@@ -66,12 +66,14 @@ object ProjectsModule extends ModuleDef {
         projects: Projects,
         projectsCounts: ProjectsCounts,
         baseUri: BaseUri,
+        mappings: ApiMappingsCollection,
```
umbreak, line 69: Instead of having this `ApiMappingsCollection` and doing `mappings.merge` in multiple places, we could just have this in `ProjectsModule`

```scala
make[ApiMappings].named("aggregated-mappings").from{
   (value: Set[ApiMappings]) => value.foldLeft(ApiMappings.empty)(_ + _)
}
```

We can then get rid of `ApiMappingsCollection` and just use 
```scala
mappings: ApiMappings @Id("aggregated-mappings")
```
wherever we need to use it

umbreak (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/ProjectsModule.scala
bogdanromanx: I'm not fond of the use of qualifiers when we have a separate data type. The merge will only be executed once for each binding so it's not a high penalty cost.

umbreak (commented):
You'll have to update `src/main/resources/contexts/search-metadata.json` and `src/main/resources/contexts/indexing-metadata.json`, for now manually I guess

bogdanromanx:
> You'll have to update `src/main/resources/contexts/search-metadata.json` and `src/main/resources/contexts/indexing-metadata.json`, for now manually I guess

good catch!

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/ProjectsModule.scala
umbreak: We do it in may other places and it works pretty well. But alright if you want to keep the `ApiMappingsCollection` and do the merge multiple times

umbreak (approved):

PR #2406, opened 2021-05-07
Fusion doc updates

Stafeeva:
- SubApps
- Workflows


adulbrich (commented):
put the canvas section above the step section 👍

adulbrich (dismissed):

docs/src/main/paradox/docs/fusion/architecture.md
```diff
@@ -28,7 +28,17 @@ Expect the SubApp feature and its functionality will expand and change in the ne
...
+
+## Customization
+
+You can customize the Header of Nexus Fusion by setting up the following environment variables:
+
+- `LOGO_IMG`: Url for an image to be used as application logo in the Header, for example, `https://drive.boogle.com/jnsjdnsjs`
+- `LOGO_LINK`: Url for the logo, for example, `https://www.epfl.ch`
+- If you use Nexus Forge, it is possible to include a `Forge templates` button by providing the url as `FORGE_LINK`, for example, `https://some-url.hi`
+
+The full list of environment variables can be found @link:[here](https://github.com/BlueBrain/nexus-web/blob/main/README.md#env-variables-list).
```
umbreak, line 44: I think the link should be any of the following:
- https://github.com/BlueBrain/nexus-web#env-variables-list
- https://github.com/BlueBrain/nexus-web/blob/main/README.md#env-variables-list

umbreak (changes requested):

docs/src/main/paradox/docs/fusion/architecture.md
Stafeeva: Looks like we are using the second option here already?

umbreak: That doesn't seem to work for some weird reason. Please remove the ```@link``` just for that one

PR #2422, opened 2021-05-10
Update schemas documentation

imsdu:
Screenshot_2021-05-10 Schemas · Blue Brain Nexus(1)


docs/src/main/paradox/docs/delta/api/current/kg-schemas-api.md
```diff
@@ -27,6 +27,13 @@ ancestor paths.
 
 @@@
 
+@@@ note { .warning }
+
+From Delta v1.5, resolution contexts and schema imports are only resolved during creates and updates.
```
umbreak, line 32: Change:
> From Delta v1.5, `resolution contexts`

to

> From Delta v1.5, remote contexts and `owl:imports`

docs/src/main/paradox/docs/delta/api/current/kg-schemas-api.md
```diff
@@ -228,7 +236,39 @@ where...
 **Example**
```
umbreak, line 236: The listings query parameters should be put in a multiline, so that it is more readable, as done in organizations docs

umbreak (changes requested):

docs/src/main/paradox/docs/delta/api/current/kg-schemas-api.md
umbreak: Also the `{sort}` field should say that by default the sorting is done by `_createdBy` and `@id`

bogdanromanx (approved):

PR #2459, opened 2021-05-19
File paths now respect tar spec + add n-quads

### LINKED ISSUE #2363: Archives `format` should support `n-quads`
Missing support for n-quads

### LINKED ISSUE #2412: IllegalArgumentException: requirement failed: File path name must be between 0 and 99 characters long, was X
Stacktrace:
```
2021-05-07 13:52:23 ERROR c.e.b.n.d.s.m.RdfExceptionHandler - An exception was thrown while evaluating a Route'
java.lang.IllegalArgumentException: requirement failed: File path name must be between 0 and 99 characters long, was 125
	at flatMap @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.api.JsonLdJavaApi$.$anonfun$compact$6(JsonLdJavaApi.scala:38)
	at parTraverseUnordered @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.api.JsonLdJavaApi$.documentLoader(JsonLdJavaApi.scala:97)
	at map @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution.$anonfun$apply$4(RemoteContextResolution.scala:36)
	at parTraverseUnordered @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.api.JsonLdJavaApi$.documentLoader(JsonLdJavaApi.scala:97)
	at flatMap @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution.inner$1(RemoteContextResolution.scala:33)
	at parTraverseUnordered @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.api.JsonLdJavaApi$.documentLoader(JsonLdJavaApi.scala:97)
	at map @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution.$anonfun$apply$4(RemoteContextResolution.scala:36)
	at map @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution$$anon$1.$anonfun$resolve$2(RemoteContextResolution.scala:60)
	at tailRecM @ ch.epfl.bluebrain.nexus.delta.service.plugin.PluginsLoader.$anonfun$load$3(PluginsLoader.scala:41)
	at map @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution.$anonfun$apply$2(RemoteContextResolution.scala:33)
	at parTraverseUnordered @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.api.JsonLdJavaApi$.documentLoader(JsonLdJavaApi.scala:97)
	at map @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution.$anonfun$apply$4(RemoteContextResolution.scala:36)
	at parTraverseUnordered @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.api.JsonLdJavaApi$.documentLoader(JsonLdJavaApi.scala:97)
	at flatMap @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution.inner$1(RemoteContextResolution.scala:33)
	at parTraverseUnordered @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.api.JsonLdJavaApi$.documentLoader(JsonLdJavaApi.scala:97)
	at flatMap @ ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution.inner$1(RemoteContextResolution.scala:33)
```

Cause:
path filename longer than [what alpakka likes when creating a tar](https://github.com/akka/alpakka/blob/f2971ca8a4a71b541cddbd5bf35af3a2a56efe71/file/src/main/scala/akka/stream/alpakka/file/model.scala#L120).

We have to limit the filename to 100 characters (maybe stripping middle segments)

imsdu:
Fixes #2412 
Fixes #2363 

Also:
* Use the proper extension for n-triples, n-quads and dot
* For generated paths,  distribute the files according to the resource representation or if it is a file to be able to fetch both the source and the compacted form for example and to identify them better.

### DIFF
--- a/delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveValue.scala
+++ b/delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveValue.scala
@@ -1,7 +1,9 @@
 package ch.epfl.bluebrain.nexus.delta.plugins.archive.model
 
-import cats.implicits.toBifunctorOps
-import ch.epfl.bluebrain.nexus.delta.plugins.archive.model.ArchiveRejection.DuplicateResourcePath
+import cats.implicits._
+import ch.epfl.bluebrain.nexus.delta.plugins.archive.model.ArchiveRejection.InvalidResourceCollection
+import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.AbsolutePath
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
 import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.decoder.JsonLdDecoderError.ParsingFailure
 import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.decoder.configuration.semiauto.deriveConfigJsonLdDecoder
 import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.decoder.{Configuration, JsonLdDecoder}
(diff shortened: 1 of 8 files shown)


delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveValue.scala
```diff
@@ -19,18 +21,35 @@ final case class ArchiveValue private (resources: NonEmptySet[ArchiveReference])
...
+    def validatePath(path: AbsolutePath) =
+      path.value.getFileName.toString.length < 100 && path.value.getParent.toString.length < 155
+
+    val (_, duplicates, invalids, longIds) = resources.value.foldLeft(
+      (Set.empty[AbsolutePath], Set.empty[AbsolutePath], Set.empty[AbsolutePath], Set.empty[Iri])
+    ) { case ((visitedPaths, duplicates, invalids, longIds), reference) =>
+      (
+        visitedPaths ++ reference.path,
+        duplicates ++ reference.path.filter(visitedPaths.contains),
+        invalids ++ reference.path.filterNot(validatePath),
```
umbreak, line 46: Why do we differentiate between `invalids` and `longIds`? Aren't both invalid paths? I don't think we need this detail here.

A path is wrong if the filename is greater than 100 or if the whole path is greater than 255

umbreak (changes requested):

umbreak:
the fact that we reject immediately is fine now, since we only offer tar. However, if we were to support zip for example, we wouldn't want to reject at this stage, but when fetching

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveValue.scala
imsdu: The long ids is only a concern if you don't provide a valid path.
The invalid paths are about the ones provided by the users that are rejected.

umbreak: In both cases the client would have to adapt the path, that's why I didn't see the need to treat that separately.
But if you think is clearer for the client with the custom rejections, that's fine too.

imsdu: The broken rule is not the same and the error message is not the same.
If you look at the rejection, the error message is generated according to the broken rules in order to try to be not too verbose.

umbreak (approved):

Issue #2359, opened 2021-05-03
Collapse fetchBy and fetchAt and fetch on all surface APIs and simplify all routes


umbreak:
Because of the fact that we have in each surface API at least 3 ways to fetch a resource, each of the surface APIs is quite verbose.

It also makes the routes verbose because it forces to explicitly call each method in different sitautions.

All that duplication can be simplified by introducing some DSL. For example:
```scala
sealed trait ResourceVersion extends Product with Serializable
object ResourceVersion {
   implicit def fromLong(value: Long): ResourceVersion = Rev(value)
   implicit def fromTag(value: TagLabel): ResourceVersion = Tag(value)
   final case object Latest extends ResourceVersion
   final case class Rev(value: Long) extends ResourceVersion
   final case class Tag(value: TagLabel) extends ResourceVersion
}
```

The we would have a single fetch on each surface API
```scala
def fetch(id: IdSegment, project: ProjectRef, version: ResourceVersion): IO[E, A]
```

And a directive on the lines of:
```scala
def version: Directive1[ResourceVersion] = 
    (parameter("rev".as[Long].?) & parameter("tag".as[TagLabel].?)) {
      case (Some(_), Some(_)) => emit(simultaneousTagAndRevRejection)
      case (Some(rev), _)     => ResourceVersion.Rev(rev)
      case (_, Some(tag))     => ResourceVersion.Tag(tag)
      case _                  => ResourceVersion.Latest
    }
```

bogdanromanx:
Since we have `ResourceRef`, we could have `IdSegmentRef` instead of `ResourceVersion`...

PR #2492, opened 2021-05-27
Added IdSegmentRef to simplify routes and SDKs

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2359

### DIFF
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/SchemasRoutes.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/SchemasRoutes.scala
@@ -139,8 +139,8 @@ final class SchemasRoutes(
                     },
                     // Fetch a schema original source
                     (pathPrefix("source") & get & pathEndOrSingleSlash & idSegmentRef(id)) { id =>
-                      authorizeFor(ref, Read).apply {
-                        operationName(s"$prefixSegment/schemas/{org}/{project}/{id}/source") {
+                      operationName(s"$prefixSegment/schemas/{org}/{project}/{id}/source") {
+                        authorizeFor(ref, Read).apply {
                           val sourceIO = schemas.fetch(id, ref).map(_.value.source)
                           emit(sourceIO.leftWiden[SchemaRejection].rejectOn[SchemaNotFound])
                         }
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/SchemasRoutes.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/SchemasRoutes.scala
@@ -11,19 +11,18 @@ import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.encoder.JsonLdEncoder
 import ch.epfl.bluebrain.nexus.delta.rdf.utils.JsonKeyOrdering
 import ch.epfl.bluebrain.nexus.delta.sdk.directives.DeltaDirectives._
 import ch.epfl.bluebrain.nexus.delta.sdk.model.routes.Tags
-import ch.epfl.bluebrain.nexus.delta.sdk.Permissions.{events, schemas => schemaPermissions}
+import ch.epfl.bluebrain.nexus.delta.sdk.Permissions.events
+import ch.epfl.bluebrain.nexus.delta.sdk.Permissions.schemas.{read => Read, write => Write}
 import ch.epfl.bluebrain.nexus.delta.sdk.Projects.FetchUuids
 import ch.epfl.bluebrain.nexus.delta.sdk._
 import ch.epfl.bluebrain.nexus.delta.sdk.circe.CirceUnmarshalling
 import ch.epfl.bluebrain.nexus.delta.sdk.directives.AuthDirectives
 import ch.epfl.bluebrain.nexus.delta.sdk.marshalling.RdfMarshalling
 import ch.epfl.bluebrain.nexus.delta.sdk.model.acls.AclAddress
-import ch.epfl.bluebrain.nexus.delta.sdk.model.identities.Caller
-import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectRef
 import ch.epfl.bluebrain.nexus.delta.sdk.model.routes.Tag
 import ch.epfl.bluebrain.nexus.delta.sdk.model.schemas.SchemaRejection
 import ch.epfl.bluebrain.nexus.delta.sdk.model.schemas.SchemaRejection.SchemaNotFound
-import ch.epfl.bluebrain.nexus.delta.sdk.model.{BaseUri, IdSegment, ResourceF}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{BaseUri, ResourceF}
 import ch.epfl.bluebrain.nexus.delta.sdk.syntax._
 import io.circe.Json
 import kamon.instrumentation.akka.http.TracingDirectives.operationName
@@ -102,7 +101,7 @@ final class SchemasRoutes(
                 // Create a schema without id segment
                 (post & pathEndOrSingleSlash & noParameter("rev") & entity(as[Json])) { source =>
                   operationName(s"$prefixSegment/schemas/{org}/{project}") {
-                    authorizeFor(ref, schemaPermissions.write).apply {
+                    authorizeFor(ref, Write).apply {
                       emit(Created, schemas.create(ref, source).map(_.void))
                     }
                   }
@@ -114,7 +113,7 @@ final class SchemasRoutes(
                         concat(
                           // Create or update a schema
                           put {
-                            authorizeFor(ref, schemaPermissions.write).apply {
+                            authorizeFor(ref, Write).apply {
                               (parameter("rev".as[Long].?) & pathEndOrSingleSlash & entity(as[Json])) {
                                 case (None, source)      =>
                                   // Create a schema with id segment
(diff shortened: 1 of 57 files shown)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/routes/StoragesRoutes.scala
```diff
@@ -164,33 +165,39 @@ final class StoragesRoutes(
...
                         )
                       }
                     },
                     // Fetch a storage original source
-                    (pathPrefix("source") & get & pathEndOrSingleSlash) {
-                      operationName(s"$prefixSegment/storages/{org}/{project}/{id}/source") {
-                        fetchSource(id, ref, res => Storage.encryptSource(res.value.source, crypto).toOption.get)
+                    (pathPrefix("source") & get & pathEndOrSingleSlash & idSegmentRef(id)) { id =>
+                      authorizeFor(ref, Read).apply {
+                        operationName(s"$prefixSegment/storages/{org}/{project}/{id}/source") {
```
bogdanromanx, line 182: name before authorization

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/routes/StoragesRoutes.scala
```diff
@@ -164,33 +165,39 @@ final class StoragesRoutes(
...
+                          val sourceIO = storages
+                            .fetch(id, ref)
+                            .map(res => Storage.encryptSource(res.value.source, crypto).toOption.get)
```
bogdanromanx, line 185: while this was there before, `toOption.get` needs to be handled differently because it can throw without a descriptive message

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/IdSegmentRef.scala
```diff
@@ -0,0 +1,61 @@
...
+    * @return the inner [[IdSegment]]
+    */
+  def value: IdSegment
+
+  /**
+    * Converts the current [[IdSegmentRef]] to an [[IdSegmentRef.Revision]]
+    */
+  def toRev(rev: Long): IdSegmentRef.Revision = IdSegmentRef.Revision(self.value, rev)
+
+  def toLatest: IdSegmentRef.Latest = IdSegmentRef.Latest(self.value)
```
bogdanromanx, line 20: docs for contracts

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/IdSegmentRef.scala
```diff
@@ -0,0 +1,61 @@
...
+object IdSegmentRef {
+
+  implicit def idSegmentToIdSegmentRef(id: IdSegment): IdSegmentRef = apply(id)
+  implicit def iriToIdSegmentRef(iri: Iri): IdSegmentRef            = Latest(iri)
+  implicit def stringToIdSegmentRef(string: String): IdSegmentRef   = Latest(string)
+
+  def apply(id: IdSegment): IdSegmentRef                             = Latest(id)
+  def apply(id: IdSegment, rev: Long): IdSegmentRef                  = Revision(id, rev)
+  def apply(id: IdSegment, tag: TagLabel): IdSegmentRef              = Tag(id, tag)
+  def fromRev(id: IdSegment, revOpt: Option[Long]): IdSegmentRef     = revOpt.fold(apply(id))(Revision(id, _))
```
bogdanromanx, line 43: `fromRevOpt` and `fromTagOpt` would read better

bogdanromanx (changes requested):
Looks a lot better than before...

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/IdSegmentRef.scala
```diff
@@ -0,0 +1,61 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.model
+
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+
+/**
+  * A segment from the positional API that should be an Id plus its revision/tag information
+  */
+sealed trait IdSegmentRef extends Product with Serializable { self =>
```
imsdu, line 8: We have now 2 ADTs describing the Latest/Tag/Rev where we will have a 4th one to get a value by timestamp.
It may be possible to have only one ADT to describe the 4 of them no ?

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
```diff
@@ -394,15 +370,11 @@ final class CompositeViews private (
...
-      project: ProjectRef,
-      rev: Option[Long]
-  ): IO[CompositeViewRejection, (Project, ViewResource)] =
+  private def fetchInner(id: IdSegmentRef, project: ProjectRef): IO[CompositeViewRejection, (Project, ViewResource)] =
     for {
       p     <- projects.fetchProject(project)
-      iri   <- expandIri(id, p)
-      state <- rev.fold(currentState(project, iri))(stateAt(project, iri, _))
+      iri   <- expandIri(id.value, p)
+      state <- id.asRev.fold(currentState(project, iri))(id => stateAt(project, iri, id.rev))
```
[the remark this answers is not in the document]
umbreak: That's not possible. Look at where `fetchInner is called`. 
```scala
id.asTag.fold(fetchInner)(...)(fetchBy(_, project))
```

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -320,93 +320,49 @@ final class Files(
...
-  )(implicit caller: Caller): IO[FileRejection, FileResponse] =
-    fetch(id, project, None)
-      .flatMap { resource =>
-        resource.value.tags.get(tag) match {
-          case Some(rev) => fetchContentAt(id, project, rev).mapError(_ => TagNotFound(tag))
-          case None      => IO.raiseError(TagNotFound(tag))
-        }
+  def fetchContent(id: IdSegmentRef, project: ProjectRef)(implicit caller: Caller): IO[FileRejection, FileResponse] =
+    id.asTag
+      .fold(
```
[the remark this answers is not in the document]
umbreak: Oh really? I've seen [this scala talk](https://www.youtube.com/watch?v=71yhnTGw0hY) where John De Goes argues for the opposite.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/routes/StoragesRoutes.scala
umbreak: Not at this scope.

Plus in theory that does not happen, since we do check for encryption and decryption on storage evaluate method. That means that if a storage made it to the by it should be possible to encrypt/decrypt it.

It is true that after some change we did some weeks ago, now is possible to change the app.conf crypto and mess it all up. In any case, it is not on the scope of this PR to fix it.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/IdSegmentRef.scala
umbreak: What 2 ADTs? Do you mean `ResourceRef`? And there is nothing YET about getting a value by timestamp. I don't see the point of thinking ahead here.

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/SchemasRoutes.scala
```diff
@@ -102,7 +101,7 @@ final class SchemasRoutes(
                 // Create a schema without id segment
                 (post & pathEndOrSingleSlash & noParameter("rev") & entity(as[Json])) { source =>
                   operationName(s"$prefixSegment/schemas/{org}/{project}") {
-                    authorizeFor(ref, schemaPermissions.write).apply {
+                    authorizeFor(ref, Write).apply {
```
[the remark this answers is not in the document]
umbreak: I'm keeping it this way. I'd sacrifice readability for fragility. The IDE should work (it works for me) and if it doesn't, the compilar will enlight you

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/routes/StoragesRoutes.scala
umbreak: Notice that this is not only happening on the routes, it also happens here: 

- `EventSerializer.jsonSecretEncryptEncoder`
- `StorageEvent.jsonSecretEncryptEncoder`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/IdSegmentRef.scala
imsdu: To avoid to refactor most of this again ?

umbreak: Things are to be refactored when the time is needed, not prematurely.
We can try to consolidate `ResourceRef` and `IdSegmentRef` in one single thing, but it is not going to be as trivial. This PR is already quite big, so I suggest we do that in another one if we see it useful.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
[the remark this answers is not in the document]
imsdu: I usually agree with folds are better than matches but in this particular case, it may be the opposite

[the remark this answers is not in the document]
umbreak: I'm not going to argue for a style issue because there is no right answer.

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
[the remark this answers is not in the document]
bogdanromanx: the risk is that later on one may not realise this behaviour and call this method with different expectations; maybe a simple rename to `fetchRevOrLatest` would be sufficient?

bogdanromanx (dismissed):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/IdSegmentRef.scala
umbreak: We might consider that later on.

PR #2523, opened 2021-06-07
Set blazegraph query timeout.

### LINKED ISSUE #2469: Blazegraph query timeout
Attempt to set a query timeout for very heavy queries that can disrupt the whole Blazegraph service.

It seems that it is already implemented [[1]](https://stackoverflow.com/questions/30787375/context-parameters-in-blazegraph-nano-sparql#answer-31772649) [[2]](https://jira.blazegraph.com/browse/BLZG-988), through an HTTP Header (to be tested):
```
curl -X POST http://localhost:8080/bigdata/sparql --data-urlencode \
'query=SELECT * { ?s ?p ?o } LIMIT 1' -H 'X-BIGDATA-MAX-QUERY-MILLIS:100'
```

wwajerowicz:
Fixes #2469

### DIFF
--- a/delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/client/BlazegraphClient.scala
+++ b/delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/client/BlazegraphClient.scala
@@ -3,12 +3,14 @@ package ch.epfl.bluebrain.nexus.delta.plugins.blazegraph.client
 import akka.actor.ActorSystem
 import akka.http.scaladsl.client.RequestBuilding.{Delete, Get, Post}
 import akka.http.scaladsl.model.StatusCodes._
-import akka.http.scaladsl.model.headers.{BasicHttpCredentials, HttpCredentials}
-import akka.http.scaladsl.model.{HttpEntity, Uri}
+import akka.http.scaladsl.model.headers.{BasicHttpCredentials, HttpCredentials, ModeledCustomHeader, ModeledCustomHeaderCompanion}
+import akka.http.scaladsl.model.{HttpEntity, HttpHeader, Uri}
 import akka.http.scaladsl.unmarshalling.FromEntityUnmarshaller
 import akka.http.scaladsl.unmarshalling.PredefinedFromEntityUnmarshallers.stringUnmarshaller
 import ch.epfl.bluebrain.nexus.delta.kernel.utils.ClasspathResourceUtils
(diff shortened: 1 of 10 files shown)


delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/client/BlazegraphClient.scala
```diff
@@ -110,16 +127,38 @@ class BlazegraphClient(
 
 object BlazegraphClient {
 
+  /**
+    * Header defining Blazegraph query timeout
+    *
+    * @param duration max query duration
+    */
+  final class `X-BIGDATA-MAX-QUERY-MILLIS`(duration: FiniteDuration)
```
umbreak, line 135: Do we need all of this in order to create a custom Http Header? It seems very verbose and unnecessary. I think you can create a raw header with a single line in akka http

umbreak (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/client/BlazegraphClient.scala
```diff
@@ -36,6 +39,19 @@ class BlazegraphClient(
...
+      indices: Iterable[String],
+      q: SparqlQuery,
+      responseType: Aux[R],
+      additionalHeaders: Seq[HttpHeader]
+  ): IO[SparqlClientError, R] = {
+    val headers = queryTimeout match {
+      case finite: FiniteDuration => additionalHeaders :+ RawHeader(timeoutHeader, finite.toMillis.toString)
+      case _                      => additionalHeaders
+    }
+    super.query(indices, q, responseType, headers)
```
imsdu, line 52: If we look here: https://github.com/BlueBrain/nexus/blob/master/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClientWorthRetry.scala#L24
And from the config, it looks like the canceled query will be retried 5 times before giving up.
It can be part of a second PR but shouldn't we have a dedicated client for user queries where we never retry.
The indexing streams would also not share the connection pool with the user related queries

wwajerowicz: I created a separate issue for that https://github.com/BlueBrain/nexus/issues/2530

umbreak (approved):

PR #2552, opened 2021-06-16
Automatic provisioning of projects.

### LINKED ISSUE #2520: Provision a default project for a user upon his first interaction with the API while authenticated
This is to support the EDX Course Integration (#2514), where for UX a user should be automatically provided with a specific project where he can interact with the system.

The general flow would be:
1. Login to Fusion
2. List organizations
3. Select the Course organization
4. List projects (the user project should already appear in the list)

Important aspects:
- the behaviour change may incur an additional computational cost; prefer the use of caches or devise a mean to perform this check only once (the existence of the pre-assigned project)
- when supporting multiple IdPs there's the possibility of subject (preferred_username) collisions; the default project must include a realm discriminator in its label
- the behaviour must have an enabled/disabled flag (e.g. `app.projects.automatic-provisioning.enabled = true`) in the configuration such that environments that are not meant to integrate with EDX do not have to incur the check's runtime penalty
- the collection of permissions that need to be assigned to the user for which the project is provisioned needs to be configurable (e.g. `app.projects.automatic-provisioning.permissions = ["resources/write", ...]`)
- the organization under which the projects will be provisioned needs to be configurable (e.g. `app.projects.automatic-provisioning.organization = "users"`)
- the project `description`, `base`, `vocab` and `api-mappings` need to be configurable (e.g. `app.projects.automatic-provisioning.config.base`)

Other notes:
- provisioning should only check for project existence
- provisioning does not need to reconcile permissions
- provisioning does not need to handle the existence of a deprecated project (if the user deprecated his project, this should not be handled; the guard that could be put in place is to avoid granting "projects/write" permission to the user, but this would impact changing the project's API mappings, base and vocab values)

wwajerowicz:
Fixes #2520


delta/app/src/main/resources/app.conf
```diff
@@ -199,6 +199,21 @@ app {
...
+    # configuration for the automatic provisioning of projects.
+    automatic-provisioning {
+      enabled = false
+      permissions = ["resources/read", "resources/write", "projects/read"]
+      description = "Auto provisioned project"
+      enabled-realms = {
+        realm = "users-realm"
+      }
+      api-mappings = {
+        documents = "https://bluebrain.github.io/nexus/vocabulary/defaultElasticSearchIndex"
```
umbreak, line 212: why is this here? This is part of the default api mappings, it does not need to be explicit.

delta/app/src/main/resources/app.conf
```diff
@@ -199,6 +199,21 @@ app {
 
     # configuration for the persistence of progress of projections
     persist-progress-config = ${app.defaults.persist-progress-config}
+
+    # configuration for the automatic provisioning of projects.
+    automatic-provisioning {
+      enabled = false
+      permissions = ["resources/read", "resources/write", "projects/read"]
+      description = "Auto provisioned project"
+      enabled-realms = {
```
umbreak, line 208: missing documentation.
also, you don't need this `=` here, right?

umbreak (changes requested):
the project gets created when doing listings on projects. What if the client doesn't do listings on projects?

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ProjectsDummy.scala
```diff
@@ -130,6 +134,18 @@ final class ProjectsDummy private (
...
   override def currentEvents(offset: Offset): fs2.Stream[Task, Envelope[ProjectEvent]] =
     journal.currentEvents(offset)
+
+  override def provisionProject(subject: Subject): UIO[Unit] = subject match {
+    case user @ User(subject, realm) if provisioningConfig.enabled =>
+      (for {
+        org       <- IO.fromOption(provisioningConfig.enabledReams.get(realm))
+        proj      <- IO.fromEither(Label.apply(subject)).mapError { _ => () }
+        projectRef = ProjectRef(org, proj)
+        exists    <- cache.fetch(projectRef).map(_.isDefined)
```
umbreak, line 144: should we use the cache, or the. regular `fetch` ?

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ProjectsDummy.scala
```diff
@@ -130,6 +134,18 @@ final class ProjectsDummy private (
 
   override def currentEvents(offset: Offset): fs2.Stream[Task, Envelope[ProjectEvent]] =
     journal.currentEvents(offset)
+
+  override def provisionProject(subject: Subject): UIO[Unit] = subject match {
+    case user @ User(subject, realm) if provisioningConfig.enabled =>
+      (for {
+        org       <- IO.fromOption(provisioningConfig.enabledReams.get(realm))
+        proj      <- IO.fromEither(Label.apply(subject)).mapError { _ => () }
```
umbreak, line 142: Is this error mappings going to give us any information about the failure reason?

delta/app/src/main/resources/app.conf
wwajerowicz: I just put it there as an example config.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsImpl.scala
```diff
@@ -114,11 +117,25 @@ final class ProjectsImpl private (
...
+  override def provisionProject(subject: Subject): UIO[Unit]              = subject match {
+    case user @ User(subject, realm) if provisioningConfig.enabled =>
+      (for {
+        org       <- IO.fromOption(provisioningConfig.enabledReams.get(realm))
+        proj      <- IO.fromEither(Label.apply(subject)).mapError { err =>
+                       logger.warn(s"Failed to create a project label for $user, due to error '${err.getMessage}'")
+                     }
+        projectRef = ProjectRef(org, proj)
+        exists    <- index.get(projectRef).map(_.isDefined)
+        _         <- provisionIfNotExists(exists, projectRef, user, acls, provisioningConfig)
```
umbreak, line 129: Do not pass this exists flag to the `provisionIfNotExists` method, but just do:
```scala
IO.when(!exists)(provisionIfNotExists(projectRef, user, acls, provisioningConfig))
```
and call it `provisionOnNotFound(prooject, user, acls, config)`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Projects.scala
```diff
@@ -159,6 +165,44 @@ trait Projects {
...
+  def provisionProject(subject: Subject): UIO[Unit]
+
+  private[delta] def provisionIfNotExists(
+      exists: Boolean,
+      projectRef: ProjectRef,
+      user: User,
+      acls: Acls,
+      provisioningConfig: AutomaticProvisioningConfig
+  ): IO[Unit, Unit] =
+    if (exists) UIO.unit
```
umbreak, line 183: this doesn't seem necessary nor useful here

wwajerowicz:
> the project gets created when doing listings on projects. What if the client doesn't do listings on projects?

The idea is that user will get to their projects through fusion, which will always involve listing the projects.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Projects.scala
```diff
@@ -159,6 +165,44 @@ trait Projects {
...
+    else {
+      val acl = Acl(AclAddress.Project(projectRef), user -> provisioningConfig.permissions)
+      (for {
+        _ <- acls
+               .append(acl, 0L)(user)
+               .onErrorRecover { case _: AclRejection.IncorrectRev => () }
```
umbreak, line 189: It can fail with `NothingToBeUpdated`, which should be recovered

umbreak, line 189: so if it returns IncorrectRev we are fine with it?

umbreak, line 189: so if it returns IncorrectRev we are fine with it?

umbreak (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectsConfig.scala
```diff
@@ -15,16 +25,106 @@ import pureconfig.generic.semiauto.deriveReader
...
 
 object ProjectsConfig {
-  implicit final val projectConfigReader: ConfigReader[ProjectsConfig] =
+
+  /**
+    * Configuration of automatic provisioning of projects.
+    *
+    * @param enabled      flag signalling whether automatic provisioning is enabled
+    * @param permissions  the permissions applied to the newly provisioned project
+    * @param enabledReams the realms for which the provisioning is enabled(map of realm label to organization in which the projects for the realm should be created)
```
imsdu, line 46: enabledReams -> enabledRealms

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Projects.scala
wwajerowicz: It means that the Acls have already been set before for that project and we don't want to override those.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Projects.scala
```diff
@@ -159,6 +165,44 @@ trait Projects {
...
   def currentEvents(offset: Offset = NoOffset): Stream[Task, Envelope[ProjectEvent]]
+
+  /**
+    * Create a project for a user according to automatic provisioning settings.
+    *
+    * @param subject  the user for which to provision the project.
+    */
+  def provisionProject(subject: Subject): UIO[Unit]
+
+  private[delta] def provisionIfNotExists(
```
umbreak, line 176: The return signature is weird: `IO[Unit, Unit]`...it does not give any information on the error . Plus, the only place where you are using provisionIfNotExists is doing `provisionIfNotExists.onErrorHandle(_ => ())`. So you aren't using the error channel at all in this case.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectsConfig.scala
```diff
@@ -15,16 +25,106 @@ import pureconfig.generic.semiauto.deriveReader
...
+    * @param apiMappings  the [[ApiMappings]] of the created project
+    * @param base         the base for the created project
+    * @param vocab        the vocab for the created project
+    */
+  final case class AutomaticProvisioningConfig(
+      enabled: Boolean,
+      permissions: Set[Permission],
+      enabledReams: Map[Label, Label],
+      description: String,
+      apiMappings: ApiMappings,
```
umbreak, line 57: I don't think that's of any use currently, since you just have apiMappings that will be added by default anyway. I would remove it if there is no current need for it, for simplicity

delta/sdk-testkit/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ProjectsBehaviors.scala
```diff
@@ -409,6 +424,48 @@ trait ProjectsBehaviors {
...
+      val projectRef       = ProjectRef(usersOrg, projectLabel)
+      val acl              = Acl(AclAddress.Project(projectRef), subject -> provisioningConfig.permissions)
+      projects.provisionProject(subject).accepted
+      projects.fetchProject(projectRef).accepted shouldEqual Project(
+        projectLabel,
+        uuid,
+        usersOrg,
+        orgUuid,
+        Some(provisioningConfig.description),
+        provisioningConfig.apiMappings,
```
umbreak, line 442: No use for apiMappings currently?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectsConfig.scala
```diff
@@ -15,16 +25,106 @@ import pureconfig.generic.semiauto.deriveReader
...
+    * @param description  the description of the created project
+    * @param apiMappings  the [[ApiMappings]] of the created project
+    * @param base         the base for the created project
+    * @param vocab        the vocab for the created project
+    */
+  final case class AutomaticProvisioningConfig(
+      enabled: Boolean,
+      permissions: Set[Permission],
+      enabledReams: Map[Label, Label],
+      description: String,
```
imsdu, line 56: Instead of repeating the 4 fields, you could have the `ProjectFields` here

umbreak (changes requested):

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ProjectsDummy.scala
wwajerowicz: I don't think we need that in the dummy implementation, but I can add it.

delta/app/src/main/resources/app.conf
umbreak: Are we going to use this in our use case (MOOC)? If not, we don't need it at all.

If it is just an example, you should put it with comments `#`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Projects.scala
umbreak: Aren't the ACLs set for a project anyway once the project is created through `ScopeInitialization (OwnerPermissionsScopeInitialization)`

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ProjectsDummy.scala
umbreak: Oh sorry, my bad. I didn't realize that's a dummy

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Projects.scala
```diff
@@ -159,6 +165,44 @@ trait Projects {
     * @param offset the last seen event offset; it will not be emitted by the stream
     */
   def currentEvents(offset: Offset = NoOffset): Stream[Task, Envelope[ProjectEvent]]
+
+  /**
+    * Create a project for a user according to automatic provisioning settings.
+    *
+    * @param subject  the user for which to provision the project.
+    */
+  def provisionProject(subject: Subject): UIO[Unit]
```
imsdu, line 174: `Projects` is a building block that is used everywhere and adding this in project complicates tests and other uses of `Projects` where we don't care about provisionning a project.
It would be better to create a higher level component (We could call it `ProjectProvisioning`) like we do right now for resolving/indexing resources.
So no impact on the tests and the dummy and it would isolate this portion of code.

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Projects.scala
wwajerowicz: But I think for this case we might want to have different permissions.

umbreak: I'm not sure it is worth the complexity and non deterministic behavior, since If there are already ACLs (IncorrectRev) we don't seem to mind _which_ permissions there are.

I'd rely on `OwnerPermissionsScopeInitialization` to do its job.

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ProjectsDummy.scala
wwajerowicz: I think cache should be faster here.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectsConfig.scala
wwajerowicz: There is the requirement for API mappings in the issue https://github.com/BlueBrain/nexus/issues/2520, please discuss there if you feel that they are not needed.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/Projects.scala
wwajerowicz: There is a requirement for custom permissions in the issue https://github.com/BlueBrain/nexus/issues/2520, please discuss there if you feel that it's not needed https://github.com/BlueBrain/nexus/issues/2520

umbreak:
I did comment directly on the issue https://github.com/BlueBrain/nexus/issues/2520#issuecomment-863350089 since it seems over engineered

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectProvisioning.scala
```diff
@@ -0,0 +1,81 @@
...
+      } yield ()).onErrorHandle { err =>
+        logger.warn(err)
+      }
+    }
+
+    override def apply(subject: Subject): UIO[Unit] = subject match {
+      case user @ User(subject, realm) if provisioningConfig.enabled =>
+        (for {
+          org       <- IO.fromOption(provisioningConfig.enabledRealms.get(realm))
+          proj      <- IO.fromEither(Label.apply(subject)).mapError { err =>
```
imsdu, line 71: This can really happen ? If yes, we should try to sanitize the subject to fit the constraint (truncate remove some special chars, ...)
Otherwise the user is blocked and won't have its project.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsImpl.scala
```diff
@@ -141,6 +141,8 @@ final class ProjectsImpl private (
       resource         <- IO.fromOption(evaluationResult.state.toResource, UnexpectedInitialState(cmd.ref))
       _                <- index.put(cmd.ref, resource)
     } yield resource
+
+  override def projectExists(project: ProjectRef): UIO[Boolean] = index.get(project).map(_.isDefined)
```
imsdu, line 145: Same than in the dummy, I would do a regular fetch like we do in other places

imsdu (changes requested):

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ProjectsDummy.scala
```diff
@@ -130,6 +130,8 @@ final class ProjectsDummy private (
 
   override def currentEvents(offset: Offset): fs2.Stream[Task, Envelope[ProjectEvent]] =
     journal.currentEvents(offset)
+
+  override def projectExists(project: ProjectRef): UIO[Boolean] = cache.fetch(project).map(_.isDefined)
```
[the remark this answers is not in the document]
wwajerowicz: But the project actor can possibly be on another node, while distributed data cache should be distributed to each node, so in the multi node setup this could mean another network call

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectProvisioning.scala
```diff
@@ -0,0 +1,81 @@
...
+      projects: Projects,
+      provisioningConfig: AutomaticProvisioningConfig
+  ): ProjectProvisioning = new ProjectProvisioning {
+
+    private def provisionOnNotFound(
+        projectRef: ProjectRef,
+        user: User,
+        acls: Acls,
+        provisioningConfig: AutomaticProvisioningConfig
+    ): UIO[Unit] = {
```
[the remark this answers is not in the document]
wwajerowicz: But on the other hand, we don't want to fail the listings when we can't create a project for the user.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectProvisioning.scala
wwajerowicz: It **shouldn't** happen, but  I didn't want to use `Label.unsafe`.  Removing chars could lead to username collisions.

delta/app/src/main/resources/app.conf
```diff
@@ -199,6 +199,25 @@ app {
...
+      # mapping between realm label and organization in which to provision the project
+      enabled-realms {
+        realm = "users-realm"
+      }
+      api-mappings {
+        # example API mapping
+        # documents = "https://bluebrain.github.io/nexus/vocabulary/defaultElasticSearchIndex"
+      }
+      vocab = ${app.http.base-uri}/
+      base = ${app.http.base-uri}/
```
bogdanromanx, line 219: both `base` and `vocab` should be optional as we may want to fallback on the project defaults; please comment these and add a description.

delta/sdk-testkit/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/testkit/ProjectsDummy.scala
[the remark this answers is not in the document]
bogdanromanx: I agree with @imsdu on this. The check needs to be consistent even if it would mean making a remote call.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectProvisioning.scala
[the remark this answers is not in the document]
bogdanromanx: I think it should be OK to fail the listing when the configuration is enabled. The user needs to know what happened and retry.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectProvisioning.scala
bogdanromanx: We have to sanitize the label to avoid unknown problems. The chance of a collision should be fairly small.

bogdanromanx (changes requested):

bogdanromanx (approved):

umbreak (approved):

PR #2555, opened 2021-06-17
Init website

imsdu:
* Both product page and the different versions of docs are built through `build.sh` into a target directory
* The product page files are just the ones copied from the master branch, I did not change anything
* The different versions of docs are handled through git [submodules](https://github.blog/2016-02-01-working-with-submodules/)
* A github action is responsible for pushing the target directory to github pages
* I added a cache action to cache npm modules as retrieving them is expensive

I checked the target directory locally and omitting the new v1.4.x/snapshot directories, they look the same.


imsdu:
Screenshot 2021-06-17 at 11-14-58 Versions · Blue Brain Nexus

bogdanromanx (approved):

Issue #2589, opened 2021-07-08
Add support for Accept-Encoding on delta


umbreak:
If `Accept-Encoding` is present on the HTTP Headers, delta service should be able to return encoded responses.

Issue #2588, opened 2021-07-07
Add support for compression on the HttpClient


umbreak:
We should add support for compression our implementation of [http client](https://doc.akka.io/docs/akka-http/current/common/encoding.html). This can improve performance on communication to the underlying ElasticSearch/ Blazegraph service (if they have compression support enabled).

PR #2590, opened 2021-07-08
Added support for encoded responses

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2588
Fixes https://github.com/BlueBrain/nexus/issues/2589

### DIFF
--- a/delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/BlazegraphViewsRoutes.scala
+++ b/delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/BlazegraphViewsRoutes.scala
@@ -75,7 +75,7 @@ class BlazegraphViewsRoutes(
     JsonLdEncoder.computeFromCirce(ContextValue(contexts.statistics))
 
   def routes: Route =
-    (baseUriPrefix(baseUri.prefix) & replaceUri("views", schema.iri, projects) & encodeResponse) {
+    (baseUriPrefix(baseUri.prefix) & replaceUri("views", schema.iri, projects)) {
       concat(
         pathPrefix("views") {
           extractCaller { implicit caller =>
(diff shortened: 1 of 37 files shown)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/routes/StoragesRoutes.scala
```diff
@@ -81,7 +81,7 @@ final class StoragesRoutes(
 
   @SuppressWarnings(Array("OptionGet"))
   def routes: Route =
-    (baseUriPrefix(baseUri.prefix) & replaceUri("storages", schemas.storage, projects)) {
+    (baseUriPrefix(baseUri.prefix) & replaceUri("storages", schemas.storage, projects) & encodeResponse) {
```
imsdu, line 84: We can't put it only in the Main where we handle cors/exceptions ?

imsdu (changes requested):

imsdu:
* Documentation is missing
* We could create an issue to also handle compression in the storage app ?

umbreak:
What documentation? This is part of the standard HTTP protocol, I don't feel we have to explicitly document it. Elasticsearch does not have any documentation for it either.

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/BlazegraphViewsRoutes.scala
```diff
@@ -75,7 +75,7 @@ class BlazegraphViewsRoutes(
     JsonLdEncoder.computeFromCirce(ContextValue(contexts.statistics))
 
   def routes: Route =
-    (baseUriPrefix(baseUri.prefix) & replaceUri("views", schema.iri, projects)) {
+    baseUriPrefix(baseUri.prefix) & replaceUri("views", schema.iri, projects) {
```
bogdanromanx, line 78: this kind of change yields compilation errors.

bogdanromanx (changes requested):
I think it's fine if we don't include this in the docs and the addition is good, but there are some issues with the PR checks (compilation, static analysis and tests).

Was this tested manually?

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/routes/BlazegraphViewsRoutes.scala
umbreak: I've fixed it

umbreak:
Yes, it was tested manually. I checked with wireshark that now the interaction with ES returns gzipped content. Same for the interactions with our API.

The few errors we had on the CI were due to the last modifications I did, where I removed the `()`

imsdu:
In documentation, we could just have a quick mention that we support it at the top of the general api reference page.
And not forget to add it later in the release notes

imsdu (approved):

bogdanromanx (approved):

PR #2633, opened 2021-07-21
Search Plugin skeleton

umbreak:
I took the liberty to also create the skeleton for the routes, service API and some models. It does not mean that they will nor they should stay as is, but more as a proof that the service wired and running properly.

### DIFF
--- a/delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchRoutes.scala
+++ b/delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchRoutes.scala
@@ -1,0 +1,55 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.search
+
+import akka.http.scaladsl.server.Directives.{as, concat, entity, get, pathEndOrSingleSlash, pathPrefix, post}
+import akka.http.scaladsl.server.Route
+import ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.routes.ElasticSearchViewsDirectives.extractQueryParams
+import ch.epfl.bluebrain.nexus.delta.plugins.search.models.SearchConfig
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution
+import ch.epfl.bluebrain.nexus.delta.rdf.utils.JsonKeyOrdering
+import ch.epfl.bluebrain.nexus.delta.sdk.circe.CirceUnmarshalling
+import ch.epfl.bluebrain.nexus.delta.sdk.directives.AuthDirectives
+import ch.epfl.bluebrain.nexus.delta.sdk.directives.DeltaDirectives.{baseUriPrefix, emit}
+import ch.epfl.bluebrain.nexus.delta.sdk.marshalling.RdfMarshalling
+import ch.epfl.bluebrain.nexus.delta.sdk.model.BaseUri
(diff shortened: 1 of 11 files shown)


delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/models/SearchRejection.scala
```diff
@@ -0,0 +1,48 @@
...
+  * @param reason a descriptive message as to why the rejection occurred
+  */
+sealed abstract class SearchRejection(val reason: String) extends Product with Serializable
+
+object SearchRejection {
+
+  /**
+    * Signals a rejection caused when interacting with the elasticserch client
+    */
+  final case class WrappedElasticSearchClientError(error: HttpClientError)
```
imsdu, line 26: If we are going to use an underlying aggregate elasticsearch view, we don't need to define another rejection no ?
I suppose we will use `ElasticSearchViewsQuery` to return the results and it already defines an error for that

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchRoutes.scala
```diff
@@ -0,0 +1,55 @@
...
+    with RdfMarshalling {
+
+  import baseUri.prefixSegment
+
+  def routes: Route =
+    baseUriPrefix(baseUri.prefix) {
+      pathPrefix("search") {
+        extractCaller { implicit caller =>
+          concat(
+            // Query the underlying aggregate elasticsearch view for global search
```
imsdu, line 36: Out of scope:
Will we store this special aggregate elasticsearch view in cache ? As it will not belong to any project ?
And update it as new projects are created ?

wwajerowicz (approved):

imsdu (changes requested):

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/models/SearchRejection.scala
umbreak: We could try to use the aggregate elasticsearch view. But do we need it? The only thing we need is to know the projects a caller has access to, which can be easily done with `projects.list` with the [appropriate filtering](https://github.com/BlueBrain/nexus/blob/master/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala#L60).

Afterwards just run your query:

```scala
client.search(query, projectsToIndices)(...)
```

There is no need of something more complicated right now. You don't need to construct an aggregate elasticsearch view (which anyway they do not currently support compositeview projections).

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchRoutes.scala
umbreak: I don't think we need any aggregate elasticsearch view, as I mentioned.

imsdu (approved):

bogdanromanx (changes requested):
Looks good, but let's wait for 1.6.0-M1 before merging.

bogdanromanx (approved):
We can release from a previous commit, nvm.

PR #2638, opened 2021-07-26
Added revised notebooks for Simulation Neuroscience MOOC v2

annakristinkaufmann:


annakristinkaufmann:
Will push to master instead.

PR #2652, opened 2021-07-29
Mooc tutorial update

adulbrich:
The new tutorial is in `try-nexus.md` (the old one was moved to `try-nexus-movielens.md`).

@annakristinkaufmann @MFSY I had multiple issues running the notebooks both in Binder and locally: I always get the error `event loop already running`. I managed to register two datasets but it does not work anymore since. Some files were uploaded though, not sure why. Maybe the size of the data is making it difficult? In addition, some directories are not created in the code and the `open()` function does not work out-of-the-box (until the directory is manually created), for example with MoueLight.

@dhaneshnm I don't see plugins in the Sandbox, we'll need to activate them so that I can show screenshots and so that students can visualize their data in the UI.


docs/src/main/paradox/docs/getting-started/try-nexus.md
```diff
@@ -1,190 +1,704 @@
 # Try Nexus
 
-In this tutorial, you will use the core features of the Nexus ecosystem through our sandbox. This requires minimal technical knowledge but the ability to install a Python library and run a jupyter notebook.
+## Step 0: Python Setup
 
-In the @ref:[first step](try-nexus.md#configuring-your-project-in-nexus-fusion), you'll learn:
+In this tutorial, you are encouraged to run Nexus Forge to interact with Nexus. Nexus Forge is a python library. We provide Python Notebooks with examples.
```
annakristinkaufmann, line 5: Maybe use "Jupyter notebooks" instead of  "Python notebooks" since they are not python specific

samuel-kerrien (approved):

PR #2653, opened 2021-07-30
Search endpoints

wwajerowicz:


delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/Search.scala
```diff
@@ -19,9 +28,63 @@ trait Search {
...
+        for {
+          allProjections    <- listProjections()
+          accessible        <-
+            acls.authorizeForAny(allProjections.map(v => ProjectAcl(v.view.project) -> v.projection.permission))
+          accessibleProjects = accessible.collect { case (p: ProjectAcl, true) => ProjectRef(p.org, p.project) }.toSet
+          accessibleIndices  = allProjections.collect {
+                                 case v if accessibleProjects.contains(v.view.project) =>
+                                   CompositeViews.index(v.projection, v.view, v.rev, indexingConfig.prefix).value
+                               }
+          results           <- client.search(payload, accessibleIndices.toSet, qp)().mapError(WrappedElasticSearchClientError)
```
imsdu, line 86: There is a problem if there are a lot of indices as the indices are part of the url (same constraint as AggregatedElasticSearchViews)
Also with this approach we will hit all indices accessible to the user even if we want to search on a subset of them (we could handle this problem later though).
And if someone wants to search on an index he does not have access to in his query, he will have an empty result and not a forbidden response (but it may be ok for now)

Maybe if we create a ES alias on the projections at creation with an alias looking like `{prefix}_search_{org}_{proj}` it would also be easier to filter them as we would have to loop on composite views but just fetch the user acls to build the accessible indices ?

Then one way to go (even it would hit all search indices) would be to make a query like:
```
GET "/{prefix}_search_{org}_{proj}/_search"
{
  "query": {
    "terms": {
      "_index": [
          "{prefix}_search_{org}_{proj}",
          "{prefix}_search_{org}_{proj}",
          ... //all indices] 
    }
    }
}
```
that would have to be merged with a ES bool query with the one passed by fusion

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchSpec.scala
```diff
@@ -0,0 +1,175 @@
...
+import org.scalatest.concurrent.Eventually
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+import org.scalatest.{CancelAfterFailure, Inspectors}
+
+import java.time.Instant
+import java.util.UUID
+import scala.concurrent.duration._
+
+class SearchSpec
```
imsdu, line 45: Search does not focus on the actual search but more on computing the accessible indices and we actually don't care about the query that we just pass on to the ES client.
So it would be great to be able to test only how we compute this list without having an ES client into this.
I think this is what we have done it somewhere for `AggregateElasticSearchViews`

imsdu (dismissed):

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchSpec.scala
wwajerowicz: Filtering the views and making sure that the correct indices are used when using ES query are two main things here and they both need to be tested

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/Search.scala
wwajerowicz: There's a lot of room to improve it here, but I'd say is better to address them in a subsequent PR

umbreak (changes requested):

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/Search.scala
wwajerowicz: I'll address it in the subsequent PR

umbreak (approved):

Issue #2525, opened 2021-06-08
Restrict volume of data through quotas (static configuration)


bogdanromanx:
For deployments that are meant to showcase the Nexus features, like for example one that supports an EDX course, there's a need to enforce quotas because of the lack of control on how many resources and files a user may create.

The system should allow the configuration and enforcement of quotas:

- maximum number of resources in the project (all persistent resource types with the exception of files)
~- maximum number of files in the project~
- maximum size of a single file in the project: Handled through storages
- maximum size of all files in the project: Handled through storages

**Expected behaviour:**

- project creation will apply a default quota defined in the configuration (e.g. `app.projects.quota.{quota-config-entry}`)
- an user can view the project quota if she/he has the `quotas/read` permission on the project
- the default project permissions need to include the new quota specific permissions
- an API call will return the project quota on read (e.g. `GET /v1/quotas/{org}/{proj}`)
- Fusion will display the quota configuration in the project details (mock-ups should be defined and presented to @adulbrich)
- nexus.js needs to be updated to support quota endpoints
- attempting to create new resources after the quota is reached should result in `400 Bad Request` or `429 Too Many Requests` (TBD) with a descriptive reason as to why the request has failed; Fusion and CLI should display the error in their standard notification mechanism
- deprecated resources should count for the quota

**Not included in the ticket:**
- quotas for number of revisions (per resource, per file, globally within a project)
- the CLI will show the quota configuration through a subcommand to projects (e.g. `projects quota {org}/{proj}`)
- the CLI will enable setting a new quota configuration through the same subcommand (e.g. `projects quota {org}/{proj} --max-resources=<int>`)

umbreak:
## Get project quotas

`GET /v1/quotas/{org}/{proj}`

```json
{
   "@type": "Quota",
   "resources": 1000
}
```

## Get project statistics

Modify the statistics endpoint to include `eventCount` and `resourceCount`. E.g.:

`GET /v1/projects/{org}/{proj}/statistics`

```json
{
    "@context": "https://bluebrain.github.io/nexus/contexts/statistics.json",
    "lastProcessedEventDateTime": "2021-08-04T13:33:01.329Z",
    "eventCount": 12,
    "resourceCount": 10
}
```

## Storage creation

Add the optional field capacilty. E.g.:
```
{
  "capacity": {max_accumulated_file_size}
}
```

umbreak:
Delta Split work:

[...]

PR #2663, opened 2021-08-05
Added resource counts to project statistics

umbreak:
Partially addresses https://github.com/BlueBrain/nexus/issues/2525


wwajerowicz (dismissed):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectCountsCollection.scala
```diff
@@ -45,16 +48,21 @@ object ProjectCountsCollection {
   /**
     * The counts for a single project
     *
-    * @param value  the number of events existing on the project
+    * @param eventsCount                the number of events existing on the project
+    * @param resourcesCount             the number of resources existing on the project
     * @param lastProcessedEventDateTime the time when the last count entry was created
     */
-  final case class ProjectCount(value: Long, lastProcessedEventDateTime: Instant)
+  final case class ProjectCount(eventsCount: Long, resourcesCount: Long, lastProcessedEventDateTime: Instant)
```
imsdu, line 55: We could drop the count on the values as we have it in the class name

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/projects/ProjectCountsCollection.scala
umbreak: I kept the count on the name because it is easier for circe serialization/deserialization. Otherwise I need a `Configuration` serializer and so on and I don't think it is worth it.

PR #2664, opened 2021-08-05
Search improvements

wwajerowicz:


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
```diff
@@ -213,16 +223,25 @@ class ElasticSearchClient(client: HttpClient, endpoint: Uri)(implicit as: ActorS
...
+    }
   }
 
   private def discardEntity(resp: HttpResponse) =
     UIO.delay(resp.discardEntityBytes()) >> IO.unit
 
-  private def indexPath(indices: Set[String]): String =
-    if (indices.isEmpty) allIndexPath else indices.mkString(",")
+  private def indexPathAndQuery(indices: Set[String], query: QueryBuilder): (String, QueryBuilder) = {
+    indices.toList match {
```
umbreak, line 240: I don't think we should always use the query to filter out the indices. We should only do it if the number of indices is higher than a threshold. It is way faster to only query a subset of indices than to query all and filter afterwards

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
```diff
@@ -213,16 +223,25 @@ class ElasticSearchClient(client: HttpClient, endpoint: Uri)(implicit as: ActorS
...
+      case single :: Nil => (single, query)
+      case more          => (allIndexPath, query.withIndices(more))
```
umbreak, line 242: what about the:
```scala
case Nil => ???
```

umbreak (changes requested):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
wwajerowicz: it shouldn't happen, since we check for empty before calling this function. If it's empty we'll call `_all` with empty `_index` `terms` filter, which will mean no indices will match

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
wwajerowicz: What should we use as the threshold? The whole fist line of the HTTP request has to be under 4kb [http.max_initial_line_length](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html).
It's not only path but also HTTP protocol parts`GET / HTTP/1.0` and query params. 
Should we make it configurable (default to 3072), and if `indices.mkString(",")` exceeds that number, than use the filtering?

umbreak: I'd say a threshold of 15-20 indices should be on the safe side for us. 

Notice that the aggregate views are limited by configuration (and the default value is around 10 If I remember well). That means that when using aggregated views we actually will always filter the indices on the URL instead of the payload, which will be better.

umbreak (dismissed):

umbreak (approved):

imsdu (approved):

PR #2671, opened 2021-08-06
Added logic to enforce quotas

umbreak:
Addresses https://github.com/BlueBrain/nexus/issues/2525


delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
```diff
@@ -448,16 +449,18 @@ object BlazegraphViews {
...
-          u <- uuidF()
-          _ <- idAvailability(c.project, c.id)
-        } yield BlazegraphViewCreated(c.id, c.project, u, c.value, c.source, 1L, t, c.subject)
-      case _       => IO.raiseError(ResourceAlreadyExists(c.id, c.project))
-    }
+    def create(c: CreateBlazegraphView) =
+      state match {
+        case Initial =>
+          for {
+            _ <- quotas.reachedForResources(c.project, c.subject)
```
imsdu, line 456: What about putting this check when we fetch the project ?
There is nothing wrong with the resource itself and this is in reality a rejection at the project level, just like `ProjectDeprecated`, `ProjectNotFound`, ...
If we have a method like:
`projects.fetch[Available]Project` (It is not the perfect name for this, it could be something else) that would check for deprecation and also for the quota.
This way, we also stay out of the different STMs and the changes on the different modules would be minimal.

imsdu (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
umbreak: From the logic point of view doesn't make much sense, right? In some places we might want to do `fetch[Available]Project` where we don't want to write afterwards.

Besides that, we currently have all the validation logic on the state machine (checking ids collision, checking some extra things on the payloads, etc). I don't see why this is not the same type of validation logic.

So, event though your suggestion is indeed more compacted, I don't think it is adequate. A method shouldn't do more than what is expected from it. Better explicit redundancy than implicit magic.

umbreak: And think about what's coming right after this PR:
- Ability to set quotas on the number of events per project
- Ability to set quotas on the number of revisions per resource

In order to implement those, your straightforward solution wouldn't work. And the current solution would work easily (you'd just need to add a check on update/tag/deprecate (e.g.: `reachedForEvents(project, subject) >> reachedFroRevision(project, subject, rev)`

imsdu: We currently don't have the one related to the project in the STM and the one related to the quota is related to the project so it makes some sense even if my solution found on a friday evening may not be the best fit but something similar should work.
The number of events per project would be related to the project so I don't see why this kind of approach wouldn't work.

The number of revisions per resource is another thing, we want to set it up for technical reasons and it should be something done in the routes when extracting the parameter.

umbreak: I agree for resources and for events, but not for revisions:

> The number of revisions per resource is another thing, we want to set it up for technical reasons and it should be something done in the routes when extracting the parameter.

If we start splitting the logic of where things need to be checked just because the code is more compacted (instead of more explicit) it is going to become a mess soon. We will start having business logic on the routes, on the STM and on the modules. I don't think that's a great solution either.

imsdu: For revisions, we agree it is a global value and not one that could be overriden by project no ?
Then it is more a "parsing" thing like we already do with project labels for example

umbreak: If it were a global value then yes. I wasn't sure that's what we wanted though

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
```diff
@@ -71,7 +72,7 @@ class Archives(
       value: ArchiveValue
   )(implicit subject: Subject): IO[ArchiveRejection, ArchiveResource] =
     (for {
-      p   <- projects.fetchActiveProject(project)
+      p   <- projects.fetchProject(project, Set(NotDeprecated))
```
imsdu, line 75: We could have constants for those sets (again naming may not be the best):
```
ProjectFetchOptions.notDeprecated
ProjectFetchOptions.newResource
ProjectFetchOptions.newRevision  // == ProjectFetchOptions.notDeprecated for now
```

imsdu (changes requested):

wwajerowicz (approved):

imsdu (approved):

PR #2676, opened 2021-08-10
Added support for event quotas

umbreak:
Addresses https://github.com/BlueBrain/nexus/issues/2525

This finalizes the quotas implementation based on static configuration


wwajerowicz (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/quotas/QuotasConfig.scala
```diff
@@ -14,27 +15,65 @@ import scala.annotation.nowarn
...
   * @param resources maximum number of resources per project
+  * @param events    maximum number of events per project
   * @param enabled   flag to enable or disable project quotas
   * @param custom    custom quotas for certain projects
   */
-final case class QuotasConfig(resources: Int, enabled: Boolean, custom: Map[ProjectRef, Int])
+final case class QuotasConfig(
+    resources: Option[Int],
+    events: Option[Int],
+    enabled: Boolean,
```
imsdu, line 25: I missed it before but in other config that we can disable (such as remote storage or s3. maybe others), we don't even parse the config if the enabled flag is set to false, it makes things a little safer as we would get a `Option[QuotasConfig]`.

imsdu (changes requested):

imsdu (dismissed):

PR #2679, opened 2021-08-11
Add capacity on disk storage and stats on storages

imsdu:
Addresses #2525 

* Add capacity to disk storage (enforcing it will come in a later PR)
* Add stats on storages
* Add new integration tests
* Fix existing ones
* Add docs


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesStatistics.scala
```diff
@@ -0,0 +1,166 @@
...
+
+    val cache =
+      CompositeKeyValueStore[ProjectRef, Iri, StorageStatEntry](
+        id,
+        (_, stats) => stats.lastProcessedEventDateTime.toEpochMilli
+      )
+
+    // Build the storage stat entry from the file event
+    def fileEventToStatEntry(f: FileEvent): Task[((ProjectRef, Iri), StorageStatEntry)] = f match {
+      case c: FileCreated if c.attributes.digest == Digest.NotComputedDigest =>
```
umbreak, line 106:
```scala
... if !c.attributes.digest.computed => ...
```

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesStatistics.scala
```diff
@@ -0,0 +1,166 @@
...
+    def buildStream: Stream[Task, Unit] =
+      Stream
+        .eval(projection.progress(projectionId))
+        .evalTap { progress =>
+          cache.putAll(progress.value.value)
+        }
+        .flatMap { progress =>
+          val initial = SuccessMessage(progress.offset, progress.timestamp, "", 1, progress.value, Vector.empty)
+          stream(progress.offset)
+            .evalMapAccumulate(initial) { case (acc, envelope) =>
```
umbreak, line 130: you seem to discard the second element on the tuple of `evalMapAccumulate`. Wouldn't `evalScan` work?

umbreak (changes requested):

delta/plugins/storage/src/main/resources/contexts/storages.json
```diff
@@ -5,6 +5,12 @@
     "_storageId": {
       "@id": "https://bluebrain.github.io/nexus/vocabulary/storageId",
       "@type": "@id"
+    },
+    "files": "https://bluebrain.github.io/nexus/vocabulary/files",
```
umbreak, line 9: `files` and `spaceUsed` are not needed here, since you have the `@vocab`.
As you can see, the ony things defined here are the ones that have a custom `@type`

umbreak (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesStatistics.scala
imsdu: I started with evalScan but it also includes the init element as part of the stream:
https://github.com/typelevel/fs2/blob/main/core/shared/src/main/scala/fs2/Stream.scala#L997

So in the processed count of the projection, there would be +1

wwajerowicz (approved):

umbreak (approved):

PR #2685, opened 2021-08-12
Enforce max capacity on disk storages

imsdu:
Addresses #2525


umbreak (approved):
There is nothing to control the capacity when linking files on `RemoteDiskStorages`. However, I don't see a simple way to tackle that, since in order to be able to do that we should know the file size when linking, which we don't currently

umbreak:
A remaining thing to be done on the storages side (partially related:
- Ability to define storages in project A and use it in project B (as we do for schemas or contexts, relying on CrossProjectResolver). This was done previously in a PR but was discarded in the time being.
- Once this exist we can actually get real value by using and reusing a single `RemoteDiskStorage` or `S3Storage` with the desired capacity.

umbreak (approved):

Issue #2689, opened 2021-08-13
Database initialization failures


umbreak:
- Cassandra tables are not populate even when `akka.persistence.cassandra.journal.tables-autocreate` is set to true.
- Tables created in `CassandraProjection` & `PostgresProjection` are created multiple times (as many times as instances of Projection[A] we have

PR #2690, opened 2021-08-13
Fixed DB initialization problems

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2689


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/AclsModule.scala
```diff
@@ -26,7 +27,10 @@ import monix.execution.Scheduler
 object AclsModule extends ModuleDef {
   implicit private val classLoader: ClassLoader = getClass.getClassLoader
 
-  make[EventLog[Envelope[AclEvent]]].fromEffect { databaseEventLog[AclEvent](_, _) }
+  make[EventLog[Envelope[AclEvent]]].fromEffect {
+    (flavour: DatabaseFlavour, as: ActorSystem[Nothing], _: DatabaseDefinitions) =>
```
imsdu, line 31: There should be a way to define a dependency to DatabaseDefinitions without repeating it.
I have looked in distage without finding something useful but what if when making the database flavour we would also run an effect that would create the tables if asked ?

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/AclsModule.scala
umbreak: the way I've seen to enforce order is to add it to the dependencies as I did (in distage).
However I can try what you're suggesting.

imsdu (approved):

Issue #2524, opened 2021-06-08
Removal of inactive projects


bogdanromanx:
For deployments that are meant to showcase the Nexus features, like for example one that supports an EDX course, a lot of projects are created for test purposes. Having the ability to archive and remove inactive projects will free resources without the need for human intervention.

The system should allow the control of an administrator to delete existing projects through:
- an API call to Delta
- a simple to use control interface in Fusion

Expected behavior:
- the function needs to be guarded by a custom permission `projects/delete`
- the Fusion control needs to be made available only to users that have the required permission (mock-ups to be defined and agreed with @adulbrich)
- the Fusion control needs to ask for user confirmation through typing, similar to the Github repository deletion, such that projects are not deleted by mistake
- after the deletion function is invoked, no further changes can be made to the project and Fusion will display a 'Deleting' status for the project
- upon completion of the delete function, the project reference is completely removed from the system:
  - all Delta managed resources (views, resolvers, storages, etc.) will be removed from caches
  - the project will no longer appear in listing operations
  - a new project with the same name can be created as if the project never existed
  - Fusion no longer displays the project in 'Deleting' status, the project is completely removed from listing operations

Other mentions:
- it is expected that the delete function is asynchronous; the API call would return a 202 Accepted status code upon successful execution
- keep track of project deletions in the logs (projects to be deleted will persist in the progress table)

Not in scope:
- Aggregated views
- CrossProjectResolvers
- projects configured with a RemoteDiskStorage or S3Storage cannot be archived unless the users explicitly consent that the files stored will be not included in the archive:
  - Fusion will inform the user that the project contains files on a RemoteDiskStorage or S3Storage that will not be included in the archive
  - the user will have to agree to execute the function even though files managed on these storage types will not be "handled" (will not be included in the archive and will not be removed from the respective storages)
  - Delta will require a query parameter on the API call `ignoreRemoteFiles=true` as means of "expressing consent"
  - attempting to archive a project without "expressing consent" will result in a 400 BadRequest with a detail as to why the action cannot be performed
 - deprecated projects can be archived
- the archive would be stored on disk on the Delta instance
- the archive needs to contain both resources and files

imsdu:
- In the first paragraph, it is written that it will not need human intervention but it appears that it does as it needs actions in fusion
- The archives will be of some use at some point ? If they are "test projects" done though the EDX course, they don't have much value no ?
[...]

PR #2720, opened 2021-08-23
Delete projects implementation

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2524

The important bits are:
1. The introduction of the event `ProjectMarkedForDeletion`
2. The introduction of the trait `ResourcesDeletion` (and its implementation on each module
3. The introduction of `DatabaseCleanup` to support cassandra and postgres DB cleanup

TODO:
- Add more tests on this functionality (so far I'm testing it manually, let's see what I discover)
- Add a way to clean the `ProjectsCounts`
- Figure out a way to fix the errors like (something related to recovery stuff)
- Check the existence of CrossProjectResolvers / AggregateViews that might exist for the deleted project and reject on deletion
```
[info] 2021-08-23 18:15:42 INFO  akka.actor.LocalActorRef - 
Message [akka.persistence.RecoveryPermitter$RecoveryPermitGranted$] from Actor[akka://delta-1-5-1-61-a542fb67-20210823-1053-SNAPSHOT/system/recoveryPermitter#1899543676] to 
Actor[akka://delta-1-5-1-61-a542fb67-20210823-1053-SNAPSHOT/system/sharding/blazegraph/31/blazegraph-myorg%252Fmyproject_https%253A%252F%252Fbluebrain.github.io%252Fnexus%252Fvocabulary%252FdefaultSparqlIndex/blazegraph-myorg%252Fmyproject_https%253A%252F%252Fbluebrain.github.io%252Fnexus%252Fvocabulary%252FdefaultSparqlIndex_state#-1551435466] 
was not delivered. [10] dead letters encountered. 
If this is not an expected behavior then Actor[akka://delta-1-5-1-61-a542fb67-20210823-1053-SNAPSHOT/system/sharding/blazegraph/31/blazegraph-myorg%252Fmyproject_https%253A%252F%252Fbluebrain.github.io%252Fnexus%252Fvocabulary%252FdefaultSparqlIndex/blazegraph-myorg%252Fmyproject_https%253A%252F%252Fbluebrain.github.io%252Fnexus%252Fvocabulary%252FdefaultSparqlIndex_state#-1551435466] may have terminated unexpectedly.

This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.
```


delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsImpl.scala
```diff
@@ -101,15 +130,17 @@ final class ProjectsImpl private (
       organizations.fetchActiveOrganization(ref.organization).void
     ) >>
       fetch(ref).flatMap {
-        case resource if options.contains(ProjectFetchOptions.NotDeprecated) && resource.deprecated =>
+        case resource if options.contains(ProjectFetchOptions.NotDeleted) && resource.value.markedForDeletion =>
```
imsdu, line 133: Shouldn't we do it the other way around ?
If you don't specify that you want to include deleted projects, it will always fail ?
If somebody fetches a resource which the project has been marked as deleted, it should fail no ?
As otherwise it could revive the persistent actor

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ResourcesDeletion.scala
```diff
@@ -0,0 +1,83 @@
...
+      extends ResourcesDeletion {
+    override def deleteRegistry(projectRef: ProjectRef): Task[ResourcesDeleted] =
+      Stream[Task, String](UrlUtils.encode(projectRef.toString))
+        .evalMap(encodedProject => currentEvents(projectRef, NoOffset).map(_ -> encodedProject))
+        .flatMap { case (stream, encodedProject) =>
+          stream
+            .groupWithin(50, 10.seconds)
+            .evalTap { events =>
+              val idsList = events.map(ev => UrlUtils.encode(lensId.get(ev.event).toString)).toList.distinct
+              Task.parTraverseUnordered(idsList)(id => stopActor(s"${moduleType}-${encodedProject}_${id}")) >>
```
imsdu, line 54: We already have the persistent id in the envelope no ?

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/ProjectsModule.scala
```diff
@@ -36,6 +41,25 @@ object ProjectsModule extends ModuleDef {
...
 
+  make[ProjectsCache].from { (config: AppConfig, as: ActorSystem[Nothing]) =>
+    ProjectsImpl.cache(config.projects)(as)
+  }
+
+  make[DeletionStatusCache].from { (config: AppConfig, as: ActorSystem[Nothing]) =>
+    ProjectsImpl.deletionCache(config.projects)(as)
+  }
+
+  make[ProjectsAggregate].fromEffect {
```
imsdu, line 52: We could make the different modules return the different tasks (delete files, caches, ...) to perform no ?

I am a little concerned about putting the aggregates at the distage level, it makes it easier to go around the module part and do stupid stuff if we don't pay attention

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FilesDeletion.scala
```diff
@@ -0,0 +1,73 @@
...
+          case Envelope(file: FileCreated, _, _, _, _, _) if file.storageType == StorageType.DiskStorage => file.storage
+          case Envelope(file: FileUpdated, _, _, _, _, _) if file.storageType == StorageType.DiskStorage => file.storage
+        }
+        .evalMap { ref =>
+          storages.fetch(ref, projectRef).map(_.value)
+        }
+        .collect { case Storage.DiskStorage(_, _, value, _, _) => value.volume }
+        .changes
+        .evalMap { volume =>
+          Task.delay(new Directory(new File(volume.value.toFile, projectRef.toString)).deleteRecursively())
```
imsdu, line 46: This should be done at the storage level

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala
```diff
@@ -100,50 +105,72 @@ final class ProjectsRoutes(
...
-                      parameter("rev".as[Long].?) {
-                        case Some(rev) => // Fetch project at specific revision
-                          emit(projects.fetchAt(ref, rev).leftWiden[ProjectRejection])
-                        case None      => // Fetch project
-                          emit(projects.fetch(ref).leftWiden[ProjectRejection])
+                      }
+                    },
+                    // Deprecate/delete project
+                    (delete & pathEndOrSingleSlash) {
+                      parameters("rev".as[Long], "wipe".?(false)) {
```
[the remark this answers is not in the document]
umbreak: from the rest API perspective, a DELETE verb should indicate a delete action. I think that's the best endpoint for the task.

But I agree, probably `prune` as a query parameter would be clearer

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/ProjectsModule.scala
```diff
@@ -44,20 +68,78 @@ object ProjectsModule extends ModuleDef {
...
+        mappings.merge,
+        cache,
+        deletionCache
+      )(baseUri, uuidF, as, scheduler)
+  }
+
+  many[ResourcesDeletionWithPriority].add {
+    (cache: ProjectsCache, agg: ProjectsAggregate, dbCleanup: DatabaseCleanup) =>
+      // This priority has to be bigger than any other module, since we want projects to be handled after other resources
+      100 -> ProjectDeletion(cache, agg, dbCleanup)
```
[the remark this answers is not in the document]
umbreak: How do you append the project one to the end, if you just collect a set of `ResourcesDeletion`? I think a priority here doesn't hurt, since we can control what comes first and what last

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsImpl.scala
```diff
@@ -33,6 +34,7 @@ final class ProjectsImpl private (
     agg: ProjectsAggregate,
     eventLog: EventLog[Envelope[ProjectEvent]],
     index: ProjectsCache,
+    deletionCache: DeletionStatusCache,
```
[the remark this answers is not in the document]
umbreak: on restart it will be built again by the `ProjectsDeletionStream`, no?

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
```diff
@@ -72,7 +72,7 @@ class Archives(
       value: ArchiveValue
   )(implicit subject: Subject): IO[ArchiveRejection, ArchiveResource] =
     (for {
-      p   <- projects.fetchProject(project, notDeprecated)
+      p   <- projects.fetchProject(project, notDeprecatedOrDeleted)
```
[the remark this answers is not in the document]
umbreak: It is? I'm not sure if that's strictly necessary the case. You can deprecate a project and then delete it (2 events) or you can directly delete a project (1 event). It doesn't seem the same to me

delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/processor/ShardedAggregate.scala
```diff
@@ -116,6 +116,16 @@ private[processor] class ShardedAggregate[State, Command, Event, Rejection](
       )
     )
 
+  /**
+    * Stops the entity with the given __id__
+    *
+    * @param id the entity identifier
+    */
+  override def stop(id: String): Task[Unit] =
```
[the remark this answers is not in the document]
umbreak: Yes. I don't think that's a big deal. However I agree this can be a future improvement

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsImpl.scala
umbreak: If someone fetches a project that has been marked for deletion but didn't specify in the options `ProjectFetchOptions.NotDeleted` it should go on (same logic as for deprecated). This sort of fetch doesn't happen very often in our system, but it can occur.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ResourcesDeletion.scala
umbreak: We do, but since the next call (`deleteAll`) needs to split the persistenceId in this 3 blocks, I though I would use reuse it. But I can use the persistenceId instead for this specific call, no problem

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/ProjectsModule.scala
umbreak: This could be done as a future refactoring.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FilesDeletion.scala
umbreak: It is done at the files level because as soon as we start supporting S3 and remote, it will have to be done here. So it centralizes the place where the actual deletion of files happens (instead of being spread across different modules)

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
```diff
@@ -219,9 +219,28 @@ final class CompositeViews private (
...
       project: ProjectRef,
       rev: Long
+  )(implicit subject: Subject): IO[CompositeViewRejection, ViewResource] =
+    deprecate(id, project, notDeprecatedOrDeletedWithEventQuotas, rev)
+
+  /**
+    * Deprecate a view without any extra checks on the projects API.
+    * @see [[deprecate(id, project, rev)]]
+    */
+  private[compositeviews] def deprecateWithoutProjectChecks(
```
wwajerowicz, line 229: The name is quite long, maybe something like `forceDeprecate` would be shorter?

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/ProjectsModule.scala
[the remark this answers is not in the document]
imsdu: The thing is Project will always come last.
And a way to do is to append the project part later and append it after you sort all the others

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsImpl.scala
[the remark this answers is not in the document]
imsdu: Not if the deletion has been completed.
But maybe I missed something

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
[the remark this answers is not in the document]
imsdu: I didn't mean at the event level but at the state level, projects that have been marked for deletion are not available for any resource writes

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsImpl.scala
imsdu: But if someone fetches a resource while deletion is happening, it can cause problems

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/DeleteProject.scala
```diff
@@ -0,0 +1,78 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectRef
+import ch.epfl.bluebrain.nexus.delta.sdk.model.resources.{ResourceCommand, ResourceEvent, ResourceRejection, ResourceState}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.search.Pagination.OnePage
+import ch.epfl.bluebrain.nexus.delta.sdk.syntax._
+import ch.epfl.bluebrain.nexus.delta.sourcing.{Aggregate, EventLog}
+import monix.bio.{IO, Task}
+
+import scala.concurrent.duration._
+
+class DeleteProject(
```
wwajerowicz, line 25: I don't think I understand why this is inside composite views? Isn't deleting elasticsearch and blazegraph handled already their respective `ViewsDeletion`s?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FilesDeletion.scala
imsdu: For the moment, we don't know what we will do with S3 or remotes or even disk storages in other contexts than the MOOC.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resolvers/ResolversDeletion.scala
```diff
@@ -0,0 +1,41 @@
...
+import monix.bio.Task
+
+final class ResolversDeletion private (
+    cache: ResolversCache,
+    stopActor: StopActor,
+    currentEvents: CurrentEvents[ResolverEvent],
+    dbCleanup: DatabaseCleanup
+) extends ProjectScopedResourcesDeletion(stopActor, currentEvents, dbCleanup, Resolvers.moduleType)(_.id) {
+
+  override def deleteData(projectRef: ProjectRef): Task[ResourcesDataDeleted] =
```
wwajerowicz, line 19: I think we should rename this method as it is not actually deleting anything. Something like `freeResources` would make more sense, as this is what it's actually doing for most resource types(elasticsearch and blazegraph views)

wwajerowicz (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
umbreak: Yes It would be shorter but less explicit. It won't forceDeprecate since it can still return a rejection for other reasons (related to the state machine)

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/DeleteProject.scala
umbreak: This was just a test I did previously. You are right, I forgot to delete it.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsImpl.scala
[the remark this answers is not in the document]
umbreak: I'll try to test that case manually for now.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FilesDeletion.scala
umbreak: We do know that we want this to be supported in dev/staging and production, since  otherwise we would have solve the current problem just with infrastructure tricks and no code. So I think it makes sense this way. We can always move it to storages if we are not going to implement deletion for other files. 

However, If you feel strong about it I can move this logic to storages.

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala
```diff
@@ -100,50 +105,72 @@ final class ProjectsRoutes(
...
+                        case (rev, true) if config.allowResourcesDeletion =>
```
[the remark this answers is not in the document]
umbreak: I think in this case is better this way. Notice that in case of deletion we have to do extra things (creating the location, and returning a SeeOther).

If we had that on the `Projects` trait it would have to return some weird response type like `IO[ProjectRejection, (Option[UUID], ProjectResource)]`.
So the current solution is more concise than the proposed solution. I don't see a reason why to change it.

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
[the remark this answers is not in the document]
umbreak: Something to be tested later on

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ResourcesDeletion.scala
```diff
@@ -50,8 +50,9 @@ object ResourcesDeletion {
           stream
             .groupWithin(50, 10.seconds)
             .evalTap { events =>
-              val idsList = events.map(ev => UrlUtils.encode(lensId.get(ev.event).toString)).toList.distinct
-              Task.parTraverseUnordered(idsList)(id => stopActor(s"${moduleType}-${encodedProject}_${id}")) >>
+              println("------->>>>" + events.map(_.persistenceId).toList.mkString(", "))
```
wwajerowicz, line 53: this should probably be done using logging or deleted.

wwajerowicz (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViewsDeletion.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  implicit private val subject: Subject = serviceAccount.subject
+
+  override def freeResources(projectRef: ProjectRef): Task[ResourcesDataDeleted] =
+    cache
+      .values(projectRef)
+      .flatMap { viewsList =>
+        Task.traverse(viewsList) { view =>
+          views
+            .deprecateWithoutProjectChecks(view.id, projectRef, view.rev)
```
imsdu, line 31: Did your tests go well with views ?
Because the deprecate event will be taken into account by the coordinator stream in (event-consistency-delay + refresh-interval) seconds.
What if the event is created and then deleted before the coordinator stream can read it ?
Maybe it would be safer to directly call the stream clean up from here ?

wwajerowicz (dismissed):

delta/plugins/elasticsearch/src/main/resources/defaults/default-mapping.json
```diff
@@ -56,6 +56,10 @@
       "type": "boolean",
       "copy_to": "_all_fields"
     },
+    "_markedAsDeprecated": {
+      "type": "boolean",
+      "copy_to": "_all_fields"
```
imsdu, line 61: It does not provide much value to copy a boolean value in the `_all_fields`

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FilesDeletion.scala
imsdu: It can stay as it is, we will change things later if needed with DLM

imsdu (changes requested):

delta/plugins/elasticsearch/src/main/resources/defaults/default-mapping.json
umbreak: Probably. This is the case for the other fields too. We can address the `_all_fields` thing and fix it in a separate PR for all the metadata, not just `_markedForDeletion`

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViewsDeletion.scala
umbreak: The tests I did do delete the indices, so I assume it works. We could do further testing on this later on.

umbreak:
> In the current solution if deletion is disabled and if you try to prune, you will actually deprecate the project which is kind of misleading because you should expect a 403 or something.

Yes. That is always the case. Our API ignores fields that are not being used (there is an issue to address this in a generic manner). When the config is set to `allowResourcesDeletion=false` we just ignore the field. So yes, it is expected to deprecate instead of to delete.

PR #2734, opened 2021-08-30
Reject project deletion if it is referenced in other projects

### LINKED ISSUE #2723: Delete Projects: Check if project is being used anywhere before allowing deletion
If the project requested to be deleted is being used by an `AggregatedView` or by a `CompositeView` or by a `CrossProjectResource` we should reject, pointing out on the details that first the client should modify the `AggregatedView` or `CrossProjectResource` or `CompositeView` to exclude the pointer to the requested project.

Until this is not done, it wouldn't be safe to introduce the deletion feature into staging/production

imsdu:
Fixes #2723

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ProjectReferenceFinder.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ProjectReferenceFinder.scala
@@ -7,7 +7,6 @@ import ch.epfl.bluebrain.nexus.delta.sdk.ProjectReferenceFinder.ProjectReference
 import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceF
 import ch.epfl.bluebrain.nexus.delta.sdk.model.projects.ProjectRef
 import io.circe.Encoder
-import io.circe.generic.semiauto.deriveEncoder
 import monix.bio.UIO
 
 import java.time.Instant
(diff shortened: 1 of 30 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ProjectReferenceFinder.scala
```diff
@@ -0,0 +1,62 @@
...
+
+    implicit val projectReferenceMapMonoid: Monoid[ProjectReferenceMap] =
+      new Monoid[ProjectReferenceMap] {
+        override def empty: ProjectReferenceMap = ProjectReferenceMap.empty
+
+        override def combine(x: ProjectReferenceMap, y: ProjectReferenceMap): ProjectReferenceMap =
+          ProjectReferenceMap(x.value |+| y.value)
+      }
+
+    implicit val projectReferenceMapEncoder: Encoder.AsObject[ProjectReferenceMap] = deriveEncoder[ProjectReferenceMap]
```
umbreak, line 48: do we want the default encoder? I think it would be better to unwrap the value (this is used in the rejection):
```scala
Encoder.encodeMap[ProjectRef, List[Iri]].contramapObject(_.value)
```

umbreak (changes requested):

bogdanromanx (approved):

umbreak (approved):

PR #2750, opened 2021-09-01
Added docs for projects delete

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/2746

screencapture-localhost-4001-docs-delta-api-projects-api-html-2021-09-01-11_12_47


imsdu:
It looks good to me.

Maybe we could just add a word about the fact that the project should not have any other project pointing to it ?

imsdu (approved):

PR #2754, opened 2021-09-02
Global search docs.

wwajerowicz:
screencapture-localhost-63342-nexus-docs-target-paradox-site-main-docs-delta-api-search-api-html-2021-09-03-09_45_21

### DIFF
--- a/docs/src/main/paradox/docs/delta/api/search-api.md
+++ b/docs/src/main/paradox/docs/delta/api/search-api.md
@@ -1,0 +1,39 @@
+# Global search
+
+Nexus provides global search functionality across all projects through the @link:[search plugin](https://github.com/BlueBrain/nexus/tree/master/delta/plugins/search/src){ open=new }.
+
+The plugin is disabled by default and requires following configuration:
+
+- `plugins.search.enabled=true` - enables the plugin. Enabling the plugin will trigger creating a `CompositeView` in each project.
+- `plugins.search.fields={pathToFile}` - specifies which fields are available for searching and/or aggregations. This also serves as configuration returned by the `/search/config` endpoint. Example can be found @link:[here](https://github.com/BlueBrain/nexus/blob/master/tests/docker/config/fields.json){ open=new }.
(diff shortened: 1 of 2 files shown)


docs/src/main/paradox/docs/delta/api/search-api.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Query
+
+```
+POST /v1/search/query
+{payload}
+```
+... where `{payload}` is an @link:[Elasticsearch query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html){ open=new } and the response
+is forwarded from the underlying Elasticsearch indices.
+The endpoint filters the Elasticsearch projections from the underlying composite views and only returns results from indices to which user has access to.
```
imsdu, line 31: We should make explicit the permission needed by the user

docs/src/main/paradox/docs/delta/api/search-api.md
```diff
@@ -0,0 +1,39 @@
+# Global search
+
+Nexus provides global search functionality across all projects through the @link:[search plugin](https://github.com/BlueBrain/nexus/tree/master/delta/plugins/search/src){ open=new }.
+
+The plugin is disabled by default and requires following configuration:
+
+- `plugins.search.enabled=true` - enables the plugin. Enabling the plugin will trigger creating a `CompositeView` in each project.
+- `plugins.search.fields={pathToFile}` - specifies which fields are available for searching and/or aggregations. This also serves as configuration returned by the `/search/config` endpoint. Example can be found @link:[here](https://github.com/BlueBrain/nexus/blob/master/tests/docker/config/fields.json){ open=new }.
```
imsdu, line 8: We should add that this current file is needed to use global search with Fusion.
If they don't use Fusion, this could be optional

It is ok for now as we will have to update it anyway in this current sprint but as it is something custom, we should document more the different values and options

Also add a link to its counterpart in Fusion doc when it gets written

docs/src/main/paradox/docs/delta/api/search-api.md
```diff
@@ -0,0 +1,39 @@
+# Global search
+
+Nexus provides global search functionality across all projects through the @link:[search plugin](https://github.com/BlueBrain/nexus/tree/master/delta/plugins/search/src){ open=new }.
```
imsdu, line 3: We need a warning that this plugin remains at experimental stage

imsdu (changes requested):

imsdu (approved):

bogdanromanx (approved):

Issue #2528, opened 2021-06-08
Record and present deployment usage stats


bogdanromanx:
**Design proposal timeboxed to 2 days**

Expose Delta metrics that allow visualizing the usage of deployment for showcasing Nexus features, for example in the case of a deployment that supports an EDX course.

Metrics over time:

- number of active projects
- project creation
- project deprecation status
- project deletion
- resource count per project
- file count per project
- file size distribution per project
- total file size per project
- ... check current grafana dashboards

Configure two Kibana dashboards (templates to be version controlled) that present:

- a global view on the project statistics (total values for all projects)
- a specific project view (values of a specific project)

Note: considering that the statistics need to be recorded over time, the `influx` projection should be generalized from the CLI and ported as a plugin of Delta instead. An initial implementation can be done without any runtime configuration; static configuration is sufficient.

Tasks:

- collect resource events and move them to ES
- create dashboards in Kibana and save them in git somewhere (GitHub)

Not in scope?

- install Kibana https://bbpteam.epfl.ch/project/issues/browse/NISE-195
- dashboard: file size distribution (Gitlab?)
- dashboard: ACLs information

adulbrich:
@bogdanromanx could we consider in the future having it as part of fusion? have a proper route like /nexus/web/statistics?

adulbrich:
This could be answered by ElsaticSearch instead of InfluxDB @imsdu 
Could we use it to also improve upon project statistics?

adulbrich:
@dhaneshnm @imsdu @nicwells @Stafeeva @umbreak Hey team! Can you please add your [planning poker estimates with ZenHub](https://app.zenhub.com/workspaces/Blue-Brain-Nexus-60ace035034cae00105c9307/issues/BlueBrain/nexus/2528?planning-poker)

PR #2766, opened 2021-09-07
Implementing deployment stats

imsdu:
Address part of #2528

### DIFF
--- a/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/metric/ProjectEventMetricsStream.scala
+++ b/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/metric/ProjectEventMetricsStream.scala
@@ -41,7 +41,7 @@ object ProjectEventMetricsStream {
     .memoize
 
   /**
-    * Default elasticsearch settings for a view
+    * Default elasticsearch settings for metrics
     */
   private val metricsSettings: UIO[JsonObject] = ClasspathResourceUtils
     .ioJsonObjectContentOf("metrics/metrics-settings.json")
--- a/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/metric/ProjectEventMetricsStream.scala
+++ b/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/metric/ProjectEventMetricsStream.scala
@@ -1,0 +1,103 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.metric
+
+import akka.actor.typed.ActorSystem
+import akka.persistence.query.Offset
+import cats.syntax.functor._
+import ch.epfl.bluebrain.nexus.delta.kernel.RetryStrategy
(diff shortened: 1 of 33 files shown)


delta/sourcing/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/projections/ProjectionStream.scala
```diff
@@ -370,6 +370,44 @@ object ProjectionStream {
...
+    def persistProgress(
+        initial: ProjectionProgress[A],
+        persistProgress: ProjectionProgress[A] => Task[Unit],
+        persistErrors: Vector[Message[A]] => Task[Unit]
+    ): Stream[Task, ProjectionProgress[A]] = {
+      stream
+        .evalMapAccumulate(initial) { case (acc, chunk) =>
+          val (progress, errors) = chunk.foldLeft((acc, Vector.empty[Message[A]])) {
+            case ((acc, errors), msg: ErrorMessage) if msg.offset.gt(acc.offset)                               => (acc + msg, errors :+ msg)
+            case ((acc, errors), msg: SuccessMessage[A]) if msg.offset.gt(acc.offset) && msg.warnings.nonEmpty =>
```
bogdanromanx, line 389: this will yield a warning and fail the build

bogdanromanx (changes requested):
Also create a separate issue (or directly a PR) to handle recording metrics for Resources that have a distribution that are not Files in the system (binaries are described but not registered)

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/metric/ProjectEventMetricsStream.scala
```diff
@@ -0,0 +1,103 @@
...
+            client
+              .bulk(
+                metrics.map { e =>
+                  ElasticSearchBulk.Index(index, e.uuid.toString, e.asJson)
+                }
+              )
+              .tapError(e => UIO.pure(println(e)))
+          }
+          .map(_.map(_.void))
+          .persistProgress(progress, projectionId, projection)
```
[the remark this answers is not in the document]
imsdu: I will create an issue to address this in a different PR.
Only views are using kamon for now, the other streams don't.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/metrics/EventMetric.scala
```diff
@@ -147,8 +153,13 @@ object EventMetric {
         "instant" -> e.instant.asJson,
         "subject" -> e.subject.asJson,
         "action"  -> e.action.asJson,
-        "@id"     -> e.id.asJson,
-        "@type"   -> e.types.asJson
+        "@id"     -> e.resourceId.asJson,
+        "@type"   -> e.types.map { tpe =>
+          Json.obj(
+            "raw"   -> tpe.asJson,
+            "short" -> tpe.toUri.toOption.flatMap(_.path.lastSegment).asJson
```
bogdanromanx, line 160: this should be Fragment or last segment

bogdanromanx (changes requested):

bogdanromanx (approved):

PR #2792, opened 2021-09-15
Allow rdf parser to be configurable

### LINKED ISSUE #2760: Invalid IRIs should be properly rejected
Currently when there is a invalid IRI (e.g. that contains a space), no error is returned to the user and the indexing in Blazegraph fails later after a simple log in Delta
```
WARN  org.apache.jena.riot - Bad IRI: < https://doi.org/10.3389/fnsyn.2019.00029> Spaces are not legal in URIs/IRIs.
```
Adding `.checking(true)` to this [line](https://github.com/BlueBrain/nexus/blob/master/delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/api/JsonLdJavaApi.scala#L68) and/or changing the error handler: https://jena.apache.org/documentation/javadoc/arq/org/apache/jena/riot/RDFParserBuilder.html#checking(boolean)
should help to solve this.

imsdu:
Fixes #2760 

* The rdf parser for validating data can be configured to be more or less strict
* The rdf parser for fetching/indexing remains lenient to be able to return a resource no matter the strictness defined for validation (and also because there is no need to validate them again)

### DIFF
--- a/delta/app/src/main/resources/app.conf
+++ b/delta/app/src/main/resources/app.conf
@@ -77,6 +77,23 @@ app {
     }
   }
 
+  # Json-LD Api configuration
+  json-ld-api {
+    # Set the Jena parser to perform extra checks:
+    # * URIs - whether IRs confirm to all the rules of the URI scheme
+    # * Literals: whether the lexical form conforms to the rules for the datatype.
+    # * Triples and quads: check slots have a valid kind of RDF term (parsers usually make this a syntax error anyway).
+    extra-checks: true
+    # Set the Jena parser to strict mode
+    # This mode do the extra-checks described above and reject relative iris/uris
+    strict: false
+    # Define the error handler for the Jena parser
+    # * strict: Both errors and warnings will raise exceptions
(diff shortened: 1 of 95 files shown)


umbreak:
Yep, checking and strict were the solution.

However your test only handles the case expanded -> Graph, as far as I can see.

Does this also throw when converting compacted -> expanded ?

imsdu:
Nope, nothing is thrown at the expanded/compacted level

umbreak:
Ok. So then as far as I can tell this is not going to solve the root problem, which is: people will still be able to create resources with those "wrong" IRIs (for example iris with a space at the beginning). 
We are just going to get another kind of warning/error.

imsdu:
The problem is solved in my PR, we transform the expanded into a graph in the resources state machine (I only had to add this for resources not validated by a schema, we already did it otherwise)
We get an `InvalidJsonLdFormat` rejection if the iri is wrong

umbreak (approved):

bogdanromanx (approved):

PR #2841, opened 2021-10-05
Deleting views is now synchronous + implement deleting progress

### LINKED ISSUE #2840: Views are deleted asynchronously which can lead to errors as project can be deleted first

imsdu:
Fixes #2840 

* Deleting views is now synchronous as project could be deleted before them and it resulted in an error
* Implement projection deletion and add it to clean up so when a view is created again, it starts from the beginning
* Physical file deletion is now handled in `StorageDeletion`
* ProjectCreationCooldown now return an `Instant` and not a `Duration`

### DIFF
--- a/delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsDeletionStream.scala
+++ b/delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsDeletionStream.scala
@@ -85,16 +85,19 @@ class ProjectsDeletionStream(
     Stream
       .eval(projection.progress(projectionId))
       .evalTap { progress =>
+        logger.info(s"Starting projects deletion stream")
         cache.putAll(progress.value.value)
       }
       .flatMap { initialProgress =>
         projects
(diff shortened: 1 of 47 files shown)


delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/projects/ProjectsDeletionStream.scala
```diff
@@ -85,16 +85,19 @@ class ProjectsDeletionStream(
...
       .eval(projection.progress(projectionId))
       .evalTap { progress =>
+        logger.info(s"Starting projects deletion stream")
         cache.putAll(progress.value.value)
       }
       .flatMap { initialProgress =>
         projects
           .events(initialProgress.offset)
           .collect { case env @ Envelope(_: ProjectMarkedForDeletion, _, _, _, _, _) =>
+            println(env)
```
wwajerowicz, line 95: this probably shouldn't be here

wwajerowicz (approved):

wwajerowicz (approved):

bogdanromanx (approved):

PR #2936, opened 2021-11-10
Allow deleting tags on resources.

### LINKED ISSUE #2860: Delete a tag on a given resource
To complete #2793, it should be possible to delete a tag on a given resource even after deprecation.

wwajerowicz:
Fixes #2860 

Screenshot 2021-11-11 at 10 04 41


delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resources/ResourceEventExchange.scala
```diff
@@ -51,10 +52,25 @@ class ResourceEventExchange(resources: Resources)(implicit base: BaseUri) extend
...
-      case ev: ResourceEvent => resourceToValue(resources.fetch(IdSegmentRef.fromTagOpt(ev.id, tag), ev.project, None))
+      case ev: ResourceEvent =>
+        tag match {
+          case None    =>
+            fetch(IdSegmentRef(ev.id), ev.project)
+          case Some(t) =>
+            resources
+              .fetch(IdSegmentRef(ev.id, t), ev.project, None)
+              .map(Resources.eventExchangeValue(_, Some(t)))
+              .redeemWith(
```
bogdanromanx, line 63: what's the reason for this `redeemWith`?

delta/service/src/test/scala/ch/epfl/bluebrain/nexus/delta/service/resources/ResourceEventExchangeSpec.scala
```diff
@@ -106,13 +106,24 @@ class ResourceEventExchangeSpec
...
-    "return the latest resource state from the event at a particular tag" in {
+    "return the resource state from the event at a particular tag" in {
       val result = exchange.toResource(deprecatedEvent, Some(tag)).accepted.value
       result.value.source shouldEqual source
       result.value.resource shouldEqual resRev1
       result.metadata.value shouldEqual ()
+      result.tag.value shouldEqual tag
+    }
+
+    "return the latest resource state if the tag has been deleted" in {
```
bogdanromanx, line 120: Should this fail if the tag was not found?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/resources/ResourceEvent.scala
```diff
@@ -157,6 +157,34 @@ object ResourceEvent {
...
+    * @param tag
+    *   the tag that was deleted
+    * @param rev
+    *   the resource revision
+    * @param instant
+    *   the instant when this event was created
+    * @param subject
+    *   the subject which created this event
+    */
+  final case class ResourceTagDeleted(
```
bogdanromanx, line 178: Missing the serialization tests for both DB and JsonLD.

bogdanromanx (changes requested):
would be useful to also include the doc updates in the PR

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resources/ResourceEventExchange.scala
wwajerowicz: if the the fetch by tag fails, we fetch the latest version and create an `EventExchangeValue` with None passed as a tag.
It's not very efficient, since if we use a tag in a view, we need to delete from the index for each resource without the given tag.

delta/service/src/test/scala/ch/epfl/bluebrain/nexus/delta/service/resources/ResourceEventExchangeSpec.scala
wwajerowicz: The way an `EventExchange` needs to work now:
if the tag is found return resource at the tag with `Some(tag)`,
if the tag is not found, return latest version of resource with `None`
It's not the most efficient way, but we cannot filter out events anymore based on tag, because we need to delete resources from the index.

delta/service/src/main/scala/ch/epfl/bluebrain/nexus/delta/service/resources/ResourceEventExchange.scala
imsdu: I think we have to change the signature in some way (at least temporarilyy) because in this current version, it is difficult to expect for the consumer to be aware that it is in fact the last version and not the tagged one. We will hit bugs in the long run.

wwajerowicz: `EventExchangeValue` now has `Option[TagLabel]` to signal if the resource is the latest revision or a tagged one. Not sure what would be a better/simpler solution.

imsdu: An `Either` or an ADT somewhere would allow to skip that second call and make things explicit.
Not sure about how to do it without stirring up a hornets' nest in terms of refactoring though :).

wwajerowicz: To skip the second call we probably would have to make `ResourceF` abstract or trait and have `LatestResourceF` and `TaggedResourceF`. That would push a lot of this complexity to `Resources`

delta/sdk-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/views/model/IndexingMessage.scala
```diff
@@ -0,0 +1,122 @@
...
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution
+import ch.epfl.bluebrain.nexus.delta.sdk.EventExchange
+import ch.epfl.bluebrain.nexus.delta.sdk.EventExchange.{EventExchangeResult, EventExchangeValue}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{BaseUri, ResourceRef}
+import ch.epfl.bluebrain.nexus.delta.sdk.syntax._
+import io.circe.Json
+import monix.bio.IO
+import org.apache.jena.graph.Node
+
+sealed trait IndexingMessage {
```
imsdu, line 19: We already use the term `Message`, maybe something like `ViewData` ?

delta/sdk-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/views/model/IndexingMessage.scala
```diff
@@ -0,0 +1,122 @@
...
+        resource.schema,
+        resource.types,
+        finalRootGraph,
+        rootMetaGraph,
+        source.removeAllKeys(keywords.context)
+      )
+    }
+  }
+
+  final case class NotFound(id: Iri) extends IndexingMessage
```
imsdu, line 90: `TagNotFound` as it is only for this particular case ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/EventExchange.scala
```diff
@@ -62,24 +63,25 @@ trait EventExchange {
...
     */
-  def toResource(event: Event, tag: Option[TagLabel]): UIO[Option[EventExchangeValue[A, M]]]
+  def toResource(event: Event, tag: Option[TagLabel]): UIO[Option[EventExchangeResult]]
 }
 
 object EventExchange {
 
+  sealed trait EventExchangeResult
+
+  final case class NotFound(id: Iri) extends EventExchangeResult
```
imsdu, line 73: `TagNotFound` ?

imsdu (changes requested):

delta/sdk-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/views/model/IndexingMessage.scala
wwajerowicz: Yes, that makes more sense.

delta/sdk-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/views/model/IndexingMessage.scala
wwajerowicz: `ViewData` is probably better, but also a bit confusing, because it's not always data, sometimes it's not found :)

PR #2954, opened 2021-11-22
Elasticsearch views pipe migration

### LINKED ISSUE #2901: Migrate existing Elasticsearch views to the new model
**Motivation**
Events related to existing elasticsearch views must be rewritten to fit the new model with pipes. 

**Acceptance criteria**
* Existing views events are correctly rewritten with the appropriate pipes
* Indexing behaviour does not change for these views

imsdu:
Fixes #2901

### DIFF
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/Main.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/Main.scala
@@ -11,6 +11,7 @@ import cats.effect.ExitCode
 import ch.epfl.bluebrain.nexus.delta.config.{AppConfig, BuildInfo}
 import ch.epfl.bluebrain.nexus.delta.kernel.kamon.KamonMonitoring
 import ch.epfl.bluebrain.nexus.delta.sdk.error.PluginError
+import ch.epfl.bluebrain.nexus.delta.sdk.migration.Migration
 import ch.epfl.bluebrain.nexus.delta.sdk.model.BaseUri
 import ch.epfl.bluebrain.nexus.delta.sdk.plugin.{Plugin, PluginDef}
 import ch.epfl.bluebrain.nexus.delta.service.plugin.PluginsLoader.PluginLoaderConfig
@@ -132,6 +133,11 @@ object Main extends BIOApp {
       implicit val cfg: AppConfig         = locator.get[AppConfig]
       val logger                          = locator.get[Logger]
       val cluster                         = Cluster(as)
+
+      sys.env.get("MIGRATION_1_7").foreach { _ =>
+        locator.get[Migration].run.runSyncUnsafe()
+        RepairTagViews.repair
+      }
 
       sys.env.get("DELETE_PERSISTENCE_IDS").foreach { persistenceIds =>
         DeletePersistenceIds.delete(persistenceIds.split(",").toSeq)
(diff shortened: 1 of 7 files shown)


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/migration/MigrationV16ToV17.scala
```diff
@@ -0,0 +1,178 @@
...
+/**
+  * Migrate elasticsearch views to new model with pipeline
+  */
+final class MigrationV16ToV17 private (
+    session: CassandraSession,
+    config: CassandraConfig
+) extends Migration {
+
+  private val selectViewEvents =
+    s"SELECT persistence_id, event, sequence_nr, timestamp FROM ${config.keyspace}.tag_views where tag_name = ? limit 1000000 allow filtering;"
```
imsdu, line 41: As there are few events for the tag `elasticsearch`, the query with the allow filtering remains performant

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/Main.scala
```diff
@@ -133,6 +134,11 @@ object Main extends BIOApp {
       val logger                          = locator.get[Logger]
       val cluster                         = Cluster(as)
 
+      sys.env.get("MIGRATION_1_7").foreach { _ =>
```
bogdanromanx, line 137: Can we check if it's a `boolean` and that its value is `true`? It's useful for the deployment to flip true/false.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/migration/MigrationV16ToV17.scala
```diff
@@ -0,0 +1,178 @@
...
+
+  private val updateMessage: String =
+    s"UPDATE ${config.keyspace}.messages set event = ? where persistence_id = ? and partition_nr = 0 and sequence_nr = ? and timestamp = ?"
```
bogdanromanx, line 44: is the partition number always 0?

bogdanromanx (changes requested):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/migration/MigrationV16ToV17.scala
imsdu: For us, yes.

```
The messages table that stores the events is partitioned by (persistence_id, partition_nr). The partition_nr is an artificial partition key to ensure that the Cassandra partition does not get too large if there are a lot of events for a single persistence_id.

akka.persistence.cassandra.journal.target-partition-size controls the number of events that the journal tries to put in each Cassandra partition. It is a target as persistAll calls will have all the events in the same partition even if it will exceed the target partition size to ensure atomicity.

It is not possible to change the value once you have data so consider if the default of 500000...
```

bogdanromanx: is it possible to read it from the table and write it back with the same value, or exclude it from the update statement?

imsdu: We can't as it is part of the PK

bogdanromanx (approved):

PR #3098, opened 2022-02-03
Add second prefix to search views

imsdu:


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeViewProjectionFields.scala
```diff
@@ -90,6 +90,7 @@ object CompositeViewProjectionFields {
   final case class ElasticSearchProjectionFields(
       id: Option[Iri] = None,
       query: SparqlConstructQuery,
+      indexGroup: Option[String],
```
bogdanromanx, line 93: any reason for the name difference? `indexPrefix` vs `indexGroup`?

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViewDecodingSpec.scala
```diff
@@ -111,6 +111,7 @@ class CompositeViewDecodingSpec
       ElasticSearchProjectionFields(
         Some(iri"http://music.com/bands"),
         query1,
+        Some("cv"),
```
bogdanromanx, line 114: We should also put a length limit and sanitization (maybe use a label here?) to this along with tests to avoid very long index names.

bogdanromanx (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeViewProjectionFields.scala
imsdu: It is the same, it meant to name them indexGroup as there is already a prefix. I will fix that

bogdanromanx (approved):

PR #3181, opened 2022-04-05
Remove Migration v1.6 to v1.7

imsdu:


bogdanromanx:
There seems to be a test for the migration that's still in the code base, so it fails the compilation.

bogdanromanx (approved):

PR #3211, opened 2022-04-28
Handle jira errors and return more meaningful error messages

imsdu:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/DeltaDirectives.scala
```diff
@@ -39,13 +39,19 @@ trait DeltaDirectives extends UriDirectives {
...
     * Completes the current Route with the provided conversion to any available entity marshaller
     */
   def emit(response: ResponseToMarshaller): Route =
-    response()
+    response(None)
+
+  /**
+    * Completes the current Route with the provided conversion to any available entity marshaller
+    */
+  def emit(status: StatusCode, response: ResponseToMarshaller): Route =
```
bogdanromanx, line 53: Is this really necessary?
The magnet pattern seems to work with custom status codes, like for example:
https://github.com/BlueBrain/nexus/blob/e6234ca40b8021a1ab05227939f5decae4d4c3a7/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/RealmsRoutes.scala#L98

bogdanromanx (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/DeltaDirectives.scala
imsdu: The existing one only works when we produces json-ld responses, not json ones:
https://github.com/BlueBrain/nexus/blob/a24ac84e3caf26abaa96face2078baadf8ec33dd/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/DeltaDirectives.scala#L71

bogdanromanx (approved):

PR #3367, opened 2022-06-15
Added ability to persist projection offsets

bogdanromanx:


delta/sourcing-psql/src/main/resources/scripts/schema.ddl
```diff
@@ -61,4 +61,13 @@ CREATE TABLE IF NOT EXISTS public.scoped_states(
...
 );
-CREATE INDEX IF NOT EXISTS scoped_states_ordering_idx ON public.scoped_states USING BRIN (ordering);
\ No newline at end of file
+CREATE INDEX IF NOT EXISTS scoped_states_ordering_idx ON public.scoped_states USING BRIN (ordering);
+
+--
+-- Table for projection offsets
+--
+CREATE TABLE if NOT EXISTS public.projection_offsets(
+    name   text,
```
imsdu, line 70: It can be added later but it would be nice to have:
* When it got created
* When it was last saved 
* For views, to have the project/id so that we can fetch them with this columns and avoid to have to fetch the view to build the name to fetch content

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/Elem.scala
```diff
@@ -39,7 +96,15 @@ object Elem {
     * @tparam A
     *   the value type of the element
     */
-  final case class SuccessElem[+A](ctx: ElemCtx, value: A) extends Elem[A]
+  final case class SuccessElem[+A](
```
imsdu, line 99: It feels like we could drop `Envelope` at some point (not now) and just use `SuccessElem` instead

imsdu, line 99: It feels like we could drop `Envelope` at some point (not now) and just use `SuccessElem` instead

imsdu (approved):

PR #3397, opened 2022-06-29
Migrate projects to new sourcing library

imsdu:


delta/app/src/main/resources/app.conf
```diff
@@ -210,58 +210,56 @@ app {
...
+    permissions = ["resources/read", "resources/write", "projects/read"]
+    description = "Auto provisioned project"
+
+    # configuration of realms for automatic provisioning
+    # mapping between realm label and organization in which to provision the project
+    enabled-realms {
+      realm = "users-realm"
     }
-    # Quotas for projects
-    quotas {
```
imsdu, line 247: Same for quotas

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala
```diff
@@ -149,18 +144,10 @@ final class ProjectsRoutes(
                     },
                     // Deprecate/delete project
                     (delete & pathEndOrSingleSlash) {
-                      parameters("rev".as[Long], "prune".?(false)) {
```
imsdu, line 152: Removing more delete stuff (as for example the uuid is computed from the datastax driver library)

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/model/Project.scala
```diff
@@ -46,11 +45,14 @@ final case class Project(
     organizationUuid: UUID,
     description: Option[String],
     apiMappings: ApiMappings,
+    defaultApiMappings: ApiMappings,
```
imsdu, line 48: Default API mappings are now injected here and not in the state as we don't want to persist them

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/quotas/QuotasImpl.scala
```diff
@@ -1,26 +1,26 @@
...
 
 final class QuotasImpl(
-    projects: Projects,
-    projectsCounts: ProjectsCounts
+    projectsCounts: ProjectsStatistics
 )(implicit quotaConfig: QuotasConfig, serviceAccountConfig: ServiceAccountConfig)
     extends Quotas {
 
-  override def fetch(ref: ProjectRef): IO[QuotaRejection, Quota] = {
-    projects.fetchProject[QuotaRejection](ref) >> IO.fromEither(quotasFromConfig(ref))
```
imsdu, line 22: No need to fetch the project before to have it consistent with acls

delta/sourcing-psql/src/main/resources/scripts/schema.ddl
```diff
@@ -58,8 +58,8 @@ CREATE TABLE IF NOT EXISTS public.scoped_states(
     rev      integer      NOT NULL,
     value    JSONB        NOT NULL,
     instant  timestamptz  NOT NULL,
-    PRIMARY KEY(type, org, project, id, tag),
-    UNIQUE (org, project, id, tag)
+    PRIMARY KEY(type, org, project, tag, id),
```
imsdu, line 61: Change the order so that we can rely on this index for project stats queries

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/QuotasRoutes.scala
```diff
@@ -41,7 +38,7 @@ final class QuotasRoutes(
   def routes: Route =
     (baseUriPrefix(baseUri.prefix) & pathPrefix("quotas")) {
       extractCaller { implicit caller =>
-        projectRef(projects).apply { implicit ref =>
+        projectRef.apply { ref =>
```
imsdu, line 41: No need to fetch quotas via project uuid (the idea is also to align quotas on how we deal with acls)

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/syntax/EventLogSyntax.scala
```diff
@@ -41,31 +41,31 @@ final class EventLogOpts[State, E <: Event](private val eventLog: EventLog[Envel
...
+    projects.fetch(projectRef).bimap(rejectionMapper.to, _ => eventLog.currentEventsByTag(module, offset))
 
   /**
     * Fetch project events
     */
   def currentProjectEvents[R](projects: Projects, projectRef: ProjectRef, offset: Offset)(implicit
       rejectionMapper: Mapper[ProjectNotFound, R]
   ): IO[R, Stream[Task, Envelope[E]]] =
-    EventLogUtils.currentProjectEvents(projects, eventLog, projectRef, offset)
+    projects.fetch(projectRef).bimap(rejectionMapper.to, _ => eventLog.currentEventsByTag("", offset))
```
bogdanromanx, line 60: The use of "" for no module is a bit odd.

bogdanromanx (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/syntax/EventLogSyntax.scala
imsdu: I agree but this class will go away soon, the problem is that a lot of classes point to each other and there is no real satisfactory way to handle this

PR #3428, opened 2022-07-19
Updated Store.vocabulary config to conform to forge v0.7.0

MFSY:
- See v0.7.0 release notes https://nexus-forge.readthedocs.io/en/latest/releases/v0.7.0-release-notes.html

- Forge config should be pulled by the tutorial from https://github.com/BlueBrain/nexus-forge so that it is in sync with each release. The version before update is outdated
- Reduced the value of max_connection from 50 to 5 by default


crisely09:
So I tried the changes and the error still persists. I believe the problem is not to conform v0.7.0, is the changes in v0.7.1, particularly to kgforge/specializations/stores/nexus/service.py adding `_schemaProject` to `store_metadata_keys`.

crisely09:
So yes, after updating the path in the tutorial file, it worked.

crisely09 (approved):

PR #3517, opened 2022-10-10
Remove NonEmptySet implementation

olivergrabinski:
Replace existing NonEmptySet implementation with `cats.data.NonEmptySet`


imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/utils/NonEmptySetImplicits.scala
```diff
@@ -0,0 +1,16 @@
...
+import cats.data.NonEmptySet
+
+object NonEmptySetImplicits {
+
+  implicit class NonEmptySetTuple2Ops[K, V](nes: NonEmptySet[(K, V)]) {
+
+    /**
+      * Converts a NonEmptySet of 2-element tuple into a Map
+      */
+    def toMap: Map[K, V] = nes.foldLeft(Map.empty[K, V]) { case (m, elem) => m + elem }
```
imsdu, line 12: I would provided a function as a parameter here like

```scala
implicit class NonEmptySetOps[A](nes: NonEmptySet[A]) {

    /**
      * Converts a NonEmptySet of 2-element tuple into a Map
      */
    def toMap[K, V](a: A => (K, V)): Map[K, V] = nes.foldLeft(Map.empty[K, V]) { a => f(a) }
}
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/utils/NonEmptySetImplicits.scala
```diff
@@ -0,0 +1,16 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.utils
+
+import cats.data.NonEmptySet
+
+object NonEmptySetImplicits {
```
imsdu, line 5: Rename to `NonEmptySetSyntax` and move it to `ch.epfl.bluebrain.nexus.delta.sdk.syntax`, add it to the package object where it will meet some others

imsdu (changes requested):

imsdu (approved):

imsdu (approved):

Issue #2299, opened 2021-04-23
Expose indexing failure information through the API


bogdanromanx:
While the statistics endpoint gives an overview of the current indexing state, it's not sufficient to debug or correct problems.

Additional API endpoints like the following would be useful in investigating and correcting problems:

```
/v1/views/{org}/{proj}/{view_id}/failures
```

* Based on SSEs
* The permissions needed to see the errors is `views/write`
* The stacktrace should NOT be included in the SSE
* The Json-LD context to use is  the existing `error` one

PR #3537, opened 2022-10-25
Add elasticsearch indexing failures API endpoint

olivergrabinski:
Add SSE based `failures` endpoint to the elasticsearch plugin

#2299


docs/src/main/paradox/docs/delta/api/views/elasticsearch-view-api.md
```diff
@@ -410,3 +410,33 @@ Request
...
+id:{id}
+```
+
+where...
+
+- `{payload}`: Json - is the actual payload of the current event
+- `{type}`: String - is a type identifier for the current event. Possible types are related to core resource types (Resouce, Schema, Resolver) and available plugin types
+- `{id}`: String - is the identifier of the resource event. It can be used in the `Last-Event-Id` query parameter
+
+### Fetch failures
```
imsdu, line 430: These changes need to go in the general view page where there are the common operations for views

Fetch indexing failures

Also we need a small description.

imsdu (approved):

imsdu (approved):

Issue #3546, opened 2022-10-28
Add purging of old indexing failures


olivergrabinski:
## Description
As they are used for debugging purposes, indexing failures (exposed by the `/v1/views/{org}/{proj}/{view_id}/failures` endpoint) older than a predefined amount of time should be deleted.

## Acceptance Criteria
- Configuration is done in Delta projection configuration.
- When not configured a default of 2 weeks is applied.
- Ensure that this is applied and that more recent failures are not deleted.

PR #3560, opened 2022-11-02
Purging of old indexing failures

olivergrabinski:
Closing #3546

### DIFF
--- a/delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/ProjectionStore.scala
+++ b/delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/ProjectionStore.scala
@@ -73,7 +73,7 @@ trait ProjectionStore {
     * @param failures
     *   the FailedElem to save
     */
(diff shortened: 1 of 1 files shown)


delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/ProjectionStore.scala
```diff
@@ -73,12 +73,12 @@ trait ProjectionStore {
     * @param failures
     *   the FailedElem to save
     */
-  def saveFailedElems(metadata: ProjectionMetadata, failures: List[FailedElem]): UIO[Unit]
+  def saveFailedElems(metadata: ProjectionMetadata, failures: List[FailedElem])(implicit clock: Clock[UIO]): UIO[Unit]
```
imsdu, line 76: The clock should be already provided by the implementation

imsdu (approved):

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/ProjectionStore.scala
olivergrabinski: Fixed

imsdu (approved):

PR #3564, opened 2022-11-03
Add the ability to restart ES projections

### LINKED ISSUE #3551: Add the ability to restart indexing of Elasticsearch views
## Description
When you have a projection, you can restart indexing from a given offset.

## Acceptance Criteria
- indexing is restarted from the given offset
- don't delete data previously indexed
- If the offset doesn't exist, restart indexing from the start.
- Ensure this will work also for BG views

imsdu:
Fixes #3551


delta/sourcing-psql/src/test/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/SupervisionSuite.scala
```diff
@@ -17,46 +18,91 @@ import scala.concurrent.duration._
...
+                 status = ExecutionStatus.Ignored,
+                 progress = ProjectionProgress.NoProgress
+               )
+             )
+      // The restart has not been acknowledged and can be read by another node
+      _ <- projections.restarts(Offset.start).assertSize(1)
+    } yield ()
   }
 
   test("Cannot fetch ignored projection descriptions (by default)") {
```
olivergrabinski, line 108: This test now fails because the `watch-restarts` projection is there and was previously unexpected. I think it's a similar issue with the other two failing tests in the suite.

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/Supervisor.scala
```diff
@@ -105,23 +117,33 @@ object Supervisor {
     stop = Task.unit
   )
 
+  private[sourcing] val watchRestartMetadata = ProjectionMetadata("system", "watch-restarts", None, None)
```
olivergrabinski, line 120: Should `DeleteExpired`, and `PurgeElemFailures` be integrated inside the supervisor in the future?

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/ProjectionStore.scala
```diff
@@ -171,7 +171,7 @@ object ProjectionStore {
         sql"""SELECT * from public.failed_elem_logs
              |WHERE projection_project = $projectionProject
              |AND projection_id = $projectionId
-             |AND ordering >= $offset
+             |AND ordering > $offset
```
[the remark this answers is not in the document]
imsdu: No problem :)

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/projections/Projections.scala
```diff
@@ -0,0 +1,156 @@
...
+  def deleteExpiredRestarts(): UIO[Unit]
+}
+
+object Projections {
+
+  def apply(xas: Transactors, config: QueryConfig, restartTtl: FiniteDuration)(implicit
+      clock: Clock[UIO]
+  ): Projections =
+    new Projections {
+      private val projectionStore        = ProjectionStore(xas, config)
```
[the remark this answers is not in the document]
imsdu: It is not urgent but it is a good idea yes.
The `ProjectionStore` was growing in complexity when it was meant for data access, that is why I changed the approach

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/Supervisor.scala
imsdu: `PurgeElemFailures` may fall into into it yes.
The idea was that at least the watch-restarts part is mandatory for the supervisor to work properly and I wanted to include it within the supervisor creation itself and not to rely distage magic to assemble it at startup.
It seemed more atomic and more understable that way, what do you think ?

olivergrabinski: I agree, but as I wrote the comment I was just thinking that if other projections get added like this it might be putting to much in the supervisor.

olivergrabinski (approved):

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/Supervisor.scala
imsdu: Yes, I think the ones to be included here must be related to the supervisor.

Issue #3195, opened 2022-04-20
Views - Support 'name' and 'description'


samuel-kerrien:
**Description**
When using Fusion (currently 1.7.0), and trying to work with a project Query panel, available views are presented using their @ID suffix. If you have created your View with an intelligible ID suffix, it is ok, but if not, you are presented with a UUID. Could we support the use of `name` so that a user can improve the presentation of Views, even after creation of the original resource ? The name should be seen in the Browse panel listing and the dropdown in the Query panel.

Here is an example of a project with many ElasticSearchView:
![Screenshot 2022-04-20 at 11.00.03.png](https://images.zenhubusercontent.com/60ad1ee0d2c5ea18af754dc0/7c653ca2-b369-4e90-bbfc-71be056e23f7)

**Work to be done**
- update the schemas (GitHub) for all View in order to allow an **optional** https://schema.org/name and https://schema.org/description properties.
- check that the UI render the name in the Browse listing and the Query dropdown in both Sparql and Elasticsearch panels

*Elasticsearch views:*
* Update the elasticsearch context (`contexts.elasticsearch`) to add the name and the description respectively as `https://schema.org/name` and `https://schema.org/description`
* The new properties must appear both on indexing views and aggregate views
* Update the `ElasticSearchViewJsonLdSourceDecoder` to include the properties in the `ElasticSearchViewFields`
* Update the `ElasticSearchViewValue` to include the properties
* Complete the tests to make sure they are correctly serializes to database for both events and states
* Update `ElasticSearchView` to include the new fields
* Complete tests like `ElasticSearchViewSpec` to make sure the JSON-LD serialization of `ElasticSearchView` is correct

*Blazegraph views:*
* Update the sparql context (`contexts.sparql`) to add the name and the description respectively as `https://schema.org/name` and `https://schema.org/description`
* The new properties must appear both on indexing views and aggregate views
* Update the `BlazegraphViewValue` to include the properties
* Complete the tests to make sure they are correctly serializes to database for both events and states
* Complete `BlazegraphViewDecodingSpec` tests
* Update `BlazegraphView` to include the new fields

*Composite Views*
Out of scope for now

PR #3565, opened 2022-11-07
Add name and description to Blazegraph views

olivergrabinski:
#3195


olivergrabinski:
To review:
1. Both name and description are `String`s.
2. schema.org is not used in the blazegraph json-ld context

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphScopeInitialization.scala
```diff
@@ -32,6 +32,8 @@ class BlazegraphScopeInitialization(views: BlazegraphViews, serviceAccount: Serv
   implicit private val kamonComponent: KamonMetricComponent = KamonMetricComponent(entityType.value)
 
   private val defaultValue: IndexingBlazegraphViewValue = IndexingBlazegraphViewValue(
+    name = None,
```
imsdu, line 35: We need a default value here

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphScopeInitialization.scala
```diff
@@ -32,6 +32,8 @@ class BlazegraphScopeInitialization(views: BlazegraphViews, serviceAccount: Serv
...
+    description = None,
```
imsdu, line 36: Same

imsdu (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphScopeInitialization.scala
olivergrabinski: Added defaults to `model`

imsdu (approved):

Issue #3071, opened 2022-01-28
Add storage and storageType field to all file events


imsdu:
* Retrieve fields from states when missing

PR #3574, opened 2022-11-15
Add storage and storageType field to all file events

olivergrabinski:
#3071

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileEvent.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileEvent.scala
@@ -117,6 +117,10 @@ object FileEvent {
     *   the file identifier
     * @param project
     *   the project the file belongs to
+    * @param storage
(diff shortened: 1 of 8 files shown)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileEvent.scala
```diff
@@ -133,6 +137,8 @@ object FileEvent {
   final case class FileAttributesUpdated(
       id: Iri,
       project: ProjectRef,
+      storage: ResourceRef.Revision,
```
imsdu, line 140: We can also add them in `FileEvent` itself

imsdu (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileEvent.scala
olivergrabinski: Thanks, fixed.

imsdu (approved):

Issue #3456, opened 2022-08-16
Reimplement project event metrics


imsdu:
* Add a typeclass on every resource linked to a project to translate it to a `ProjectScopedMetric`
* Files should also include information to be able to compute the number of files, the file size per storage
```
final private case class FileExtraFields(
      storage: Iri,
      physicalFile: Option[Int],
      bytes: Option[Long],
      mediaType: Option[ContentType],
      origin: Option[FileAttributesOrigin]
  )
```
* Write the SQL query to retrieve all scoped events
* Write the projection (which will persists its offset) and which pushes the metrics into ES
* Write a StorageStatistics that relies on the index populated by this projection

Not in scope: Check the kibana dashboard

PR #3576, opened 2022-11-17
Add `EventMetricsProjection`

olivergrabinski:
Part of #3456 

Adds a projection that reads scoped events and pushes their metrics to ES.

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/metrics/EventMetric.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/metrics/EventMetric.scala
@@ -144,7 +144,7 @@ object EventMetric {
       )
   }
 
-  implicit val eventMetricEncoder: Encoder.AsObject[EventMetric] = {
+  implicit def eventMetricEncoder[E <: EventMetric]: Encoder.AsObject[E] = {
     @nowarn("cat=unused")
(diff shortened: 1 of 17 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/metrics/EventMetric.scala
```diff
@@ -144,7 +144,7 @@ object EventMetric {
       )
   }
 
-  implicit val eventMetricEncoder: Encoder.AsObject[EventMetric] = {
+  implicit def eventMetricEncoder[E <: EventMetric]: Encoder.AsObject[E] = {
```
imsdu, line 147: We need this change ?

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/pipes/AsJson.scala
```diff
@@ -0,0 +1,32 @@
...
+import monix.bio.Task
+import shapeless.Typeable
+
+object AsJson {
+
+  private val label = Label.unsafe("asJson")
+
+  private def elemValueToJson[A: Encoder]: SuccessElem[A] => Task[Elem[Json]] =
+    elem =>
+      Task {
```
imsdu, line 18: `Task.pure`

imsdu (changes requested):

PR #3599, opened 2022-11-28
Add name and description to resolvers

olivergrabinski:


delta/sdk/src/test/resources/resolvers/cross-project-provided-entities-from-value.json
```diff
@@ -4,6 +4,8 @@
     "Resolver",
     "CrossProject"
   ],
+  "name": null,
```
imsdu, line 7: We should not encode the nulls, there is something to fix there

delta/sdk/src/test/resources/resolvers/database/resolver-cross-project-created-1.json
```diff
@@ -14,6 +14,8 @@
   },
   "value": {
     "@type": "CrossProjectValue",
+    "name": null,
```
imsdu, line 17: We should not encode the nulls.
We should have at least one of the examples of serialization to have these values populated

imsdu (changes requested):

delta/sdk/src/test/resources/resolvers/database/resolver-cross-project-created-1.json
olivergrabinski: Added test with resolvers that have non null name/description fields.

imsdu (approved):

PR #3607, opened 2022-12-06
Add composite restarts + enable back composite views in Integration tests

imsdu:


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ElasticSearchViewsSpec.scala
```diff
@@ -228,10 +228,10 @@ class ElasticSearchViewsSpec extends BaseSpec with EitherValuable with CirceEq {
         response.status shouldEqual StatusCodes.OK
         val expected = jsonContentOf(
           "/kg/views/statistics.json",
-          "total"     -> "12",
-          "processed" -> "12",
+          "total"     -> "13",
```
imsdu, line 231: There is 1 more resource as a composite views is created automatically for search

olivergrabinski (approved):

PR #3612, opened 2022-12-21
Add local migration script

olivergrabinski:
Adding scripts, docker-compose files to perform local migration from backup.

Also updated:
* Permissions `entityType`/`label`
* `enrichJsonFileEvent` function to fix a decoding failure
* `toIgnore` fix & tests added


delta/migration/scripts/cassandra-init/init.sh
```diff
@@ -0,0 +1,5 @@
+until cqlsh -f /cassandra-init/cassandra_1_5.ddl && cqlsh -f /cassandra-init/create-materialized-view.ddl && sstableloader --nodes localhost  /cassandra-backups/cassandra-0/delta_1_5/messages && sstableloader --nodes localhost  /cassandra-backups/cassandra-1/delta_1_5/messages && sstableloader --nodes localhost  /cassandra-backups/cassandra-2/delta_1_5/messages; do
```
imsdu, line 1: cassandra-0, cassandra-1 and cassandra-2 have the same messages, no need to load them all

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeViewProjection.scala
```diff
@@ -117,7 +119,7 @@ object CompositeViewProjection {
       resourceTag: Option[UserTag],
       includeMetadata: Boolean,
       includeDeprecated: Boolean,
-      includeContext: Boolean,
+      includeContext: Boolean = false,
```
imsdu, line 122: There should be a better way to handle this

delta/migration/scripts/cassandra-init/init.sh
olivergrabinski: Fixed

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeViewProjection.scala
olivergrabinski: I ended up reverting the decoder changes and writing the function to enrich the `CompositeViewEvents`.

imsdu (approved):

PR #3617, opened 2023-01-10
Evaluate transactions inside the query stream

olivergrabinski:
Previously the pool could run out of connections.
The queries that are streamed now need to have a `LIMIT` clause.


delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/query/StreamingQuery.scala
```diff
@@ -178,7 +179,7 @@ object StreamingQuery {
     * @param query
     *   the query to execute depending on the offset
     * @param extractOffset
-    *   how to extract the offset from an [[A]] to be able to pursue the stream
+    *   how to extract the offset from an [[Json]] to be able to pursue the stream
```
imsdu, line 182: The function is on the A

imsdu (changes requested):

imsdu (approved):

Issue #3569, opened 2022-11-09
Only a subset of changes on a view should trigger reindexing


imsdu:
Only a change on a field affecting reindexing (i.e filtering, ES mapping, ...) should trigger a reindexing.

Others like name, permission or adding a tag should not.

This should not affect the event but the idea is to have a new field in the state `indexingRev` which would be the revision where the last meaningful change for indexing took place

PR #3625, opened 2023-01-19
Use indexing revision to trigger Blazegraph reindexing

olivergrabinski:
#3569


olivergrabinski:
To do:
- [ ] Review `indexingRev` name
- [ ] Namespace should match the projection name

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphCoordinator.scala
```diff
@@ -45,26 +45,38 @@ final class BlazegraphCoordinator private (
...
+                      supervisor.run(
+                        projection,
+                        for {
+                          _ <- createNamespace(active)
+                          _ <- cache.put(active.ref, active)
+                        } yield ()
+                      )
+                  }
+              case (_, deprecated: DeprecatedViewDef)                                              =>
+                cleanupCurrent(deprecated.ref)
```
imsdu, line 77: I missed it in the ES part but you don't have to do another look up to get the value in the cache

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphCoordinator.scala
```diff
@@ -45,26 +45,38 @@ final class BlazegraphCoordinator private (
...
+            (cachedView, v) match {
+              case (Some(cached), active: ActiveViewDef) if cached.projection == active.projection =>
+                for {
+                  _ <- cache.put(active.ref, active)
+                  _ <- Task.delay {
+                         logger.info(s"Index ${active.projection} already exists and will not be recreated.")
+                       }
+                } yield ()
+              case (_, active: ActiveViewDef)                                                      =>
+                IndexingViewDef
```
imsdu, line 59: Maybe create a private method compile here to improve a little readibility ?
Something like:
```scala
private def compile(v: ActiveViewDef) =
    IndexingViewDef.compile(v, compilePipeChain, graphStream, sink(v))
```

imsdu (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/BlazegraphCoordinator.scala
olivergrabinski: Corrected in both

imsdu (approved):

PR #3632, opened 2023-01-23
Provide elem stream as SSEs

imsdu:


delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/EventsRoutesSpec.scala
```diff
@@ -41,9 +40,9 @@ class EventsRoutesSpec extends BaseRouteSpec {
   private val identities = IdentitiesDummy(caller)
   private val asAlice    = addCredentials(OAuth2BearerToken("alice"))
 
-  private val acl      = Label.unsafe("acl")
-  private val project  = Label.unsafe("project")
-  private val resource = Label.unsafe("resource")
+  private val acl       = Label.unsafe("acl")
+  private val project   = Label.unsafe("project")
+  private val resources = Label.unsafe("resources")
```
imsdu, line 45: Renamed because it was a little confusing as we do use the plural one in real situation

PR #3641, opened 2023-01-26
Update dependencies + Scala / SBT version

imsdu:
Updates
* Update most dependencies
* Scala version
* SBT version
* Most SBT plugins
* Docker base image

Does not update:
* RDF/JSON-LD related librairies (need proper dedicated testing)
* AWS client dependency (makes integration tests fail)
* sbt-site, sbt-native-packager, sbt-coverage  plugins


build.sbt
```diff
@@ -473,6 +477,7 @@ lazy val blazegraphPlugin = project
     name                       := "delta-blazegraph-plugin",
     moduleName                 := "delta-blazegraph-plugin",
     libraryDependencies       ++= Seq(
+      scalaXml,
```
olivergrabinski, line 480: Might have to be added to the sdk instead of the blazegraphPlugin

olivergrabinski (approved):

PR #3652, opened 2023-02-02
Handle invalid state occurences during migration

olivergrabinski:
An invalid state occurrence will be logged as a warning, and the migration will proceed.

### DIFF
--- a/delta/migration/src/main/scala/ch/epfl/bluebrain/nexus/migration/Migration.scala
+++ b/delta/migration/src/main/scala/ch/epfl/bluebrain/nexus/migration/Migration.scala
@@ -8,12 +8,13 @@ import ch.epfl.bluebrain.nexus.delta.kernel.database.Transactors
 import ch.epfl.bluebrain.nexus.delta.rdf.syntax.iriStringContextSyntax
 import ch.epfl.bluebrain.nexus.delta.sdk.acls.Acls
 import ch.epfl.bluebrain.nexus.delta.sdk.migration.{MigrationLog, ToMigrateEvent}
+import ch.epfl.bluebrain.nexus.delta.sourcing.EvaluationError.InvalidState
 import ch.epfl.bluebrain.nexus.delta.sourcing.config.BatchConfig
 import ch.epfl.bluebrain.nexus.delta.sourcing.offset.Offset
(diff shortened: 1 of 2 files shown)


delta/migration/src/main/scala/ch/epfl/bluebrain/nexus/migration/Migration.scala
```diff
@@ -132,6 +119,31 @@ object Migration {
     }
   }
 
+  def processEvent(logs: Set[MigrationLog]): ToMigrateEvent => Task[Unit] = { event =>
+    val logMap = logs.map { log => log.entityType -> log }.toMap
```
imsdu, line 123: We don't need to compute the logMap for every event

delta/migration/src/main/scala/ch/epfl/bluebrain/nexus/migration/Migration.scala
```diff
@@ -132,6 +119,31 @@ object Migration {
...
+      case Some(migrationLog) =>
+        migrationLog(event)
+          .tapError { e =>
+            Task.delay {
+              logger.error(s"[${event.persistenceId}] $e")
+            }
+          }
+          .onErrorRecoverWith { case e: InvalidState[_, _] =>
+            Task.delay {
+              logger.warn(s"Invalid state encountered. Proceeding with migration. Original message: ${e.getMessage}")
```
imsdu, line 135: It would be nice to have some details about the failing event

delta/migration/src/main/scala/ch/epfl/bluebrain/nexus/migration/Migration.scala
```diff
@@ -132,6 +119,31 @@ object Migration {
     }
   }
 
+  def processEvent(logs: Set[MigrationLog]): ToMigrateEvent => Task[Unit] = { event =>
+    val logMap = logs.map { log => log.entityType -> log }.toMap
+
+    logMap.get(event.entityType) match {
+      case Some(migrationLog) =>
+        migrationLog(event)
+          .tapError { e =>
```
imsdu, line 128: Not sure we need this log here

imsdu (changes requested):

Issue #3512, opened 2022-10-05
Retrieve Nexus metadata when fetching  resource original payload


MFSY:
**Motivation**
Currently [fetching a resource original payload](https://bluebrainnexus.io/docs/delta/api/resources-api.html#fetch-original-payload) is not retrieving the nexus metadata (ie. _rev, _self, ...). Clients that need to get resources as they were submitted and the Nexus metadata at the same time have to make two http calls slowing down their workflow. It would be good to retrieve at the same time and in one http call the original payload of a resource and the resource nexus metadata

https://bluebrainnexus.io/docs/delta/api/resources-api.html#fetch-original-payload

**Acceptance criteria**
1. When performing this api call GET /v1/resources/{org_label}/{project_label}/{schema_id}/{resource_id}/source?rev={rev}&tag={tag}
then the resource original payload should be retrieved along with the corresponding metadata
2. Only applies to resources (ignore param if on other types)
3. include [documentation](https://bluebrainnexus.io/docs/delta/api/resources-api.html#fetch-original-payload)
4. includes Unit tests
5. includes integration tests

imsdu:
Not to break compatibility, I would enable this via a new parameter `annotated` (any better name is ok) that would allow to add the metadata to the original payload.
Just to avoid any misunderstanding, the result would have that shape:
```json
{
 "name": "...",
 "description": "...",
 "_createdAt": "...",
 "_rev": 5.
  "_self": "..."
}
```

MFSY:
Hi @imsdu ,
Thank you for looking into this. I guess here name and description are user data as registered in the first place. If yes than I'm okay with this.

PR #3695, opened 2023-02-21
Optionally retrieve Nexus metadata when fetching resource original payload (#3512)

shinyhappydan:


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -167,9 +170,20 @@ final class ResourcesRoutes(
...
+                                  resources
+                                    .fetch(id, ref, schemaOpt)
+                                    .flatMap(asSourceWithMetadata)
+                                )
+                              } else {
+                                val sourceIO = resources.fetch(id, ref, schemaOpt).map(_.value.source)
+                                val value    = sourceIO.leftWiden[ResourceRejection]
+                                emit(value.rejectWhen(wrongJsonOrNotFound))
+                              }
+                            }
```
shinyhappydan, line 186: This seems like a lot of logic for a Route, but there is no layer between this and `Resources` so it seems like the best place

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -291,6 +294,13 @@ class ResourcesRoutesSpec extends BaseRouteSpec {
       }
     }
 
+    "fetch a resource original payload with metadata" in {
```
imsdu, line 297: It would be nice to have some tests:
* To validate that we are forbidden to fetch the source (annotated or not) when we don't have the permission
* To check that we have a 404 when the resource does not exist
* To check we can combine annotate with the rev and tag parameters

docs/src/main/paradox/docs/delta/api/resolvers-api.md
```diff
@@ -250,15 +250,18 @@ if the `Accept` header is set to `text/html`, a redirection to the fusion repres
 ## Fetch original payload
 
 ```
-GET /v1/resolvers/{org_label}/{project_label}/{resolver_id}/source?rev={rev}&tag={tag}
+GET /v1/resolvers/{org_label}/{project_label}/{resolver_id}/source?rev={rev}&tag={tag}&annotate={annotate}
```
imsdu, line 253: This should go in `resources-api.md`
Also this is not mandatory for this PR but we try to include a snapshot in the description of the PR to make sure the rendering is ok

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ResourcesSpec.scala
```diff
@@ -140,6 +140,25 @@ class ResourcesSpec extends BaseSpec with EitherValuable with CirceEq {
       }
     }
 
+    "fetch the original payload with metadata" in {
```
imsdu, line 143: Same remark than in the unit tests

imsdu (changes requested):

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -248,4 +263,30 @@ object ResourcesRoutes {
...
+
+  private def metadataJson(resource: ResourceF[Resource])(implicit baseUri: BaseUri, cr: RemoteContextResolution) = {
+    implicit val resourceFJsonLdEncoder: JsonLdEncoder[ResourceF[Unit]] = ResourceF.defaultResourceFAJsonLdEncoder
+    resourceFJsonLdEncoder
+      .compact(resource.void)
+      .map(_.json)
+      .mapError(e => InvalidJsonLdFormat(Some(resource.id), e))
+  }
+
+  private val IdLens = root.`@id`.string
```
imsdu, line 283: We haven't really used `circe-optics` outside of migration code and tests so far.
It has not been really well maintained lately, so it would be better to stick to vanilla circe.

delta/app/src/test/resources/errors/resource-not-found.json
```diff
@@ -0,0 +1,5 @@
+{
+  "@context": "https://bluebrain.github.io/nexus/contexts/error.json",
```
imsdu, line 2: This error is already defined in the `not-found.json` file

docs/src/main/paradox/docs/delta/api/resources-api.md
```diff
@@ -210,15 +210,18 @@ if the `Accept` header is set to `text/html`, a redirection to the fusion repres
...
 ```
 where ...
 
 - `{rev}`: Number - the targeted revision to be fetched. This field is optional and defaults to the latest revision.
 - `{tag}`: String - the targeted tag to be fetched. This field is optional.
+- `{annotate}`: Boolean - annotate the response with the resource metadata. This field only applies to standard resources. This field is optional.
 
 `{rev}` and `{tag}` fields cannot be simultaneously present.
 
+If `{annotate}` is set, fields present in the metadata will override fields with the same name from the payload.
```
imsdu, line 223: A word about how we handle the `@id` field ?

imsdu (changes requested):

shinyhappydan:
image

imsdu (approved):

PR #3696, opened 2023-02-21
Indexing IDs of resources posted without `@id`s in custom ES views

### LINKED ISSUE #3666: Ids of resources with generated ids are not indexed in custom elasticsearch views
So when there is no id in the original source, we should inject the id of the resource

olivergrabinski:
Closes #3666 

The generated ID will be indexed.


imsdu (changes requested):

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/GraphResourceToDocumentSuite.scala
```diff
@@ -0,0 +1,113 @@
...
+          "@type" : "http://schema.org/Person",
+          "name" : "John Doe"
+        }
+      """
+
+    for {
+      json <- graphResourceToDocument(elem).accepted.toOption
+    } yield json.equalsIgnoreArrayOrder(expectedJson)
+  }
+
```
shinyhappydan, line 96: In this test could it be better to only assert on the `@id` field rather than the entire JSON object?

imsdu (approved):

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ElasticSearchViewsSpec.scala
```diff
@@ -222,6 +243,15 @@ class ElasticSearchViewsSpec extends BaseSpec with EitherValuable with CirceEq {
...
     }
 
+    "the person resource created with no id in payload should have the default id in _source" in eventually {
+      deltaClient.post[Json](s"/views/$fullId2/test-resource:people/_search", matchAll, ScoobyDoo) { (json, response) =>
+        response.status shouldEqual StatusCodes.OK
+        val id       = hits(0)._id.string.getOption(json)
+        val sourceId = hits(0)._source.`@id`.string.getOption(json)
+        sourceId shouldEqual id
+      }
+    }
```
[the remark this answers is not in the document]
imsdu: It is not a bad idea for most of the tests but here it may slow down the test execution:
* When we create a view, it takes some time for the indexing routine to be started
* When the routine starts, it processes the other elements and batches things
* Elasticsearch itself has a refresh period to show the results

So do the setup before is in the end not a bad idea :)

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/GraphResourceToDocumentSuite.scala
imsdu: It does not hurt, it shows that injecting the `@id` has no other side effects

PR #3717, opened 2023-02-28
handle elasticsearch bulk update partial failures

### LINKED ISSUE #3677: Improve error reporting when indexing into Elasticsearch
**Motivation**
When any indexed element is rejected by Elasticsearch is rejected, the whole batch is mark as rejected, even those that were properly accepted.

The bulk api from Elasticsearch allow us to process individually there results:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html

**Acceptance criteria**
* The accepted elements are properly indexed
* The rejected elements are rejected and the error associated is the rejection linked to it and not the whole Elasticsearch response

shinyhappydan:
when a bulk elasticsearch update is partially rejected, make the element stream reflect that state accurately

Fixes #3677

### DIFF
--- a/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
+++ b/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
@@ -26,7 +26,7 @@ import ch.epfl.bluebrain.nexus.delta.sdk.model.{BaseUri, Name}
 import ch.epfl.bluebrain.nexus.delta.sdk.syntax._
 import com.typesafe.scalalogging.Logger
 import io.circe.syntax._
-import io.circe.{Decoder, DecodingFailure, Json, JsonObject}
+import io.circe.{Decoder, DecodingFailure, HCursor, Json, JsonObject}
 import monix.bio.{IO, UIO}
 import retry.syntax.all._
 
@@ -576,15 +576,11 @@ object ElasticSearchClient {
     final case class Error(json: JsonObject) extends BulkItemOutcome
   }
 
+  private val Operations                      = List("update", "create", "delete", "index")
   implicit private val bulkItemOutcomeDecoder = Decoder.instance[BulkItemOutcome] { hcursor =>
-    val obj = hcursor
-      .downField("update")
-      .success
-      .orElse(hcursor.downField("create").success)
-      .orElse(hcursor.downField("delete").success)
-      .orElse(hcursor.downField("index").success)
-      .toRight(DecodingFailure("expected item of type update, create, delete or index", hcursor.history))
-    obj
+    Operations
+      .collectFirstSome(hcursor.downField(_).success)
+      .toRight(DecodingFailure(s"operation type was not one of ${Operations.mkString(", ")}", hcursor.history))
       .map(itemCursor =>
         itemCursor.get[JsonObject]("error") match {
           case Left(_)      => BulkItemOutcome.Success
--- a/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
+++ b/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
@@ -10,7 +10,8 @@ import cats.syntax.all._
 import ch.epfl.bluebrain.nexus.delta.kernel.RetryStrategy
 import ch.epfl.bluebrain.nexus.delta.kernel.RetryStrategy.logError
 import ch.epfl.bluebrain.nexus.delta.kernel.utils.UrlUtils
-import ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.client.ElasticSearchClient.BulkResponse.{NoErrors, SomeErrors}
+import ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.client.ElasticSearchClient.BulkResponse.MixedOutcomes.{HasMixedOutcomes, Outcome}
+import ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.client.ElasticSearchClient.BulkResponse.Success
 import ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.client.ElasticSearchClient._
 import ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.model.{emptyResults, ResourcesSearchParams}
 import ch.epfl.bluebrain.nexus.delta.sdk.circe.CirceMarshalling._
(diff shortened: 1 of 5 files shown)


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
```diff
@@ -208,20 +209,24 @@ class ElasticSearchClient(client: HttpClient, endpoint: Uri, maxIndexPathLength:
...
-        .flatMap { json =>
-          IO.unless(json.hcursor.get[Boolean]("errors").contains(false))(
-            IO.raiseError(HttpClientStatusError(req, BadRequest, json.noSpaces))
-          )
+        .map {
+          case BatchPartialFailure(err) => err
+          case _                        => NoErrors
+        }
+        .recover { case BatchPartialFailure(err) =>
+          err
```
shinyhappydan, line 226: It's unclear from the API documentation whether we get 200 or 400 in this situation, so we need to try and extract the partial error in both cases

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
```diff
@@ -558,4 +563,47 @@ object ElasticSearchClient {
...
+    obj
+      .map(itemCursor =>
+        itemCursor.get[JsonObject]("error") match {
+          case Left(_)      => BulkItemOutcome.Success
+          case Right(value) => BulkItemOutcome.Error(value)
+        }
+      )
+  }
+
+  object BatchPartialFailure {
```
imsdu, line 592: Could this go in the companion object of `BulkErrors` (or whatever it is called) ?

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchSink.scala
```diff
@@ -56,12 +58,19 @@ final class ElasticSearchSink private (
...
+            elements.zip(Chunk.seq(items)).map {
+              case (element, BulkItemOutcome.Success)     => element.void
+              case (element, BulkItemOutcome.Error(json)) =>
+                element.failed(BulkUpdateException(json))
+            }
+        }
+        .onErrorHandleWith(err =>
+          UIO
+            .delay(logger.error(s"Indexing in elasticsearch index ${index.value} failed", err))
+            .as(elements.map(_.failed(err)))
```
shinyhappydan, line 73: This is the case of an unknown error. Here it makes sense to still log and fail the elements

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchSink.scala
```diff
@@ -56,12 +58,19 @@ final class ElasticSearchSink private (
...
+        .map {
+          case NoErrors          => elements.map(_.void)
+          case SomeErrors(items) =>
+            elements.zip(Chunk.seq(items)).map {
+              case (element, BulkItemOutcome.Success)     => element.void
+              case (element, BulkItemOutcome.Error(json)) =>
+                element.failed(BulkUpdateException(json))
+            }
+        }
+        .onErrorHandleWith(err =>
```
imsdu, line 70: `tapError` as we only want to log

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchSink.scala
imsdu, line 73: For the other errors (like a timeout), we don't want to mark the elems as failed but we want to let it crash.
The indexing will restart moments later, it will retry the same query

imsdu (changes requested):

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/bio/BioAssertions.scala
```diff
@@ -43,7 +43,7 @@ trait BioAssertions { self: Assertions =>
       )
     }
 
-    def assert(expected: A, clue: Any = "values are not the same"): UIO[Unit] = io.redeemCause(
+    def assert(expected: A, clue: Any = "values are not the same")(implicit loc: Location): UIO[Unit] = io.redeemCause(
```
olivergrabinski, line 46: This might already be coming from the class it is in

shinyhappydan: This just changes the position of the error to be where this assert method is called

shinyhappydan: (You can never not have a location, I believe. But when writing assertion helpers like this, adding this location makes navigation to the "real" error easier)

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
```diff
@@ -558,4 +563,47 @@ object ElasticSearchClient {
     implicit val decodeCount: Decoder[Count] =
       Decoder.instance(_.get[Long]("count").map(Count(_)))
   }
+
+  sealed trait BulkResponse
+  object BulkResponse {
+    final case object NoErrors                                extends BulkResponse
+    final case class SomeErrors(items: List[BulkItemOutcome]) extends BulkResponse
```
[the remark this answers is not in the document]
shinyhappydan: I wanted to communicate that there may also be successes. Perhaps `MixedOutcomes`?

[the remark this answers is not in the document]
imsdu: You are right, `SomeErrors` is ok

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchSink.scala
imsdu: I agree that it changes the previous behaviour but now that we could extract the failures related to bad input, the remaining cases are technical ones: ES down, misconfiguration, network problem, etc...

So those are not related to the elems themselves but to the plaform that got unstable.
If you mark the elems that are marked as failed during this period, you will have to reindex them as they will be missing.
Where as if we let those type of errors propagate, the indexing progress will fail, there will be some backoff, and the process will restart and if ES is back on track, those will eventually pass.

imsdu (approved):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
olivergrabinski: Did you get 400 answers for with mixed cases?

shinyhappydan: I did once during testing.. 😒

PR #3731, opened 2023-03-07
Only use LRU caches for immutable values + fix TTL for project stats cache

### LINKED ISSUE #3730: TTL for cache statistics is too long
**Describe the bug**
The cache for project statistics should be set to a lower value and have a different value than cache for project uuids.
The eviction should also not be on last access

imsdu:
Fixes #3730


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/quotas/QuotasImpl.scala
```diff
@@ -24,10 +24,10 @@ final class QuotasImpl(
   override def reachedForResources(
       ref: ProjectRef,
       subject: Subject
-  ): IO[QuotaReached, Unit] =
-    IO.unless(subject == serviceAccountConfig.value.subject) {
+  ): IO[QuotaReached, Unit] = {
+    val quota = quotasFromConfig(ref).toOption
+    IO.unless(quota.isEmpty || subject == serviceAccountConfig.value.subject) {
```
imsdu, line 29: No need to fetch the stats if there is no quota to compare to

olivergrabinski (approved):

PR #3743, opened 2023-03-15
Add sink for Graph analytics

imsdu:
Part of #3415


delta/plugins/graph-analytics/src/main/resources/elasticsearch/mappings.json
```diff
@@ -9,6 +9,38 @@
     "_rev": {
       "type": "long"
     },
+    "_createdAt": {
```
imsdu, line 12: I added those at it is an trivial addition and it can help to solve some questions asked in past

delta/plugins/graph-analytics/src/test/resources/indexed-documents-after-query.json
```diff
@@ -0,0 +1,272 @@
+{
```
imsdu, line 1: Quite a big chunk but it is meant to be sure that the painless script does not have undesired side effects

delta/plugins/graph-analytics/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsSinkSuite.scala
```diff
@@ -0,0 +1,174 @@
...
+      chunk              = Chunk.seq(List(r1, r2, r3))
+      // We expect no error
+      _                 <- sink(chunk).assert(chunk.map(_.void))
+      // Elasticsearch should have taken into account the bulk query and the update query
+      expectedDocuments <- ioJsonContentOf("indexed-documents.json")
+      sort               = SortList(Sort("@id") :: Nil)
+      _                 <- client
+                             .search(JsonObject.empty, Set(index.value), Uri.Query.Empty)(sort)
+                             .map(_.removeKeys("took", "_shards"))
+                             .eventually(expectedDocuments)
```
shinyhappydan, line 125: Perhaps we can assert on the things we need.. like the hits containing the expected document ids?

Another possibility is making an alias for _.removeKeys("took", "_shards") to explain that you don't want to compare the ES metadata, as it's not clear to someone unfamiliar with ES that this is why you remove those fields

delta/plugins/graph-analytics/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsSinkSuite.scala
```diff
@@ -0,0 +1,174 @@
...
+    )
+
+    for {
+      _                 <- sink(chunk).assert(chunk.map(_.void))
+      expectedDocuments <- ioJsonContentOf("indexed-documents-after-query.json")
+      sort               = SortList(Sort("@id") :: Nil)
+      _                 <- client
+                             .search(JsonObject.empty, Set(index.value), Uri.Query.Empty)(sort)
+                             .map(_.removeKeys("took", "_shards"))
+                             .eventually(expectedDocuments)
```
shinyhappydan, line 170: It's hard to understand here what the expected behaviour is

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchSink.scala
```diff
@@ -58,26 +57,7 @@ final class ElasticSearchSink private (
...
-                          "reason" -> Json.fromString(s"${element.id} was not found in Elasticsearch response")
-                        )
-                      )
-                    )
-                  case Some(MixedOutcomes.Outcome.Success)     => element.void
-                  case Some(MixedOutcomes.Outcome.Error(json)) => element.failed(BulkUpdateException(json))
-                }
-            }
-        }
+        .map(ElasticSearchSink.markElems(_, elements, documentId))
```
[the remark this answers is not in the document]
imsdu: If we follow the Scala vocabulary, it looks more like a zip than a map ?

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsSink.scala
```diff
@@ -0,0 +1,99 @@
...
+object GraphAnalyticsSink {
+
+  private val empty = Acc(List.empty, Map.empty)
+
+  // Accumulator of operations to push to Elasticsearch
+  final private case class Acc(bulk: List[ElasticSearchBulk], updates: Map[Iri, Set[Iri]]) {
+    def add(index: ElasticSearchBulk): Acc    = copy(bulk = index :: bulk)
+    def update(id: Iri, types: Set[Iri]): Acc = copy(updates = updates + (id -> types))
+    def remove(id: Iri): Acc                  = copy(updates = updates - id)
+  }
```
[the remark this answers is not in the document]
imsdu: You mean the `remove` ?
Indeed this is some dead code, I will take care of it

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsResult.scala
```diff
@@ -0,0 +1,84 @@
...
+    * @param value
+    *   the document to push to elasticsearch
+    * @param createdAt
+    *   the instant when the resource was created
+    * @param createdBy
+    *   the identity that created the resource
+    * @param updatedAt
+    *   the instant when the resource was last updated
+    * @param updatedBy
+    *   the identity that last updated the resource
```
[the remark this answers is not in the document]
imsdu: What would you propose ? Just getting rid of them ?

delta/plugins/graph-analytics/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsSinkSuite.scala
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  test("Create the update script and the index") {
+    for {
+      script  <- scriptContent
+      _       <- client.createScript(updateRelationshipsScriptId, script)
+      mapping <- graphAnalyticsMappings
+      _       <- client.createIndex(index, Some(mapping), None).assert(true)
+    } yield ()
+  }
```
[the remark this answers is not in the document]
imsdu: I thought of it when writing tests for the regular `ElasticSearchSinkSuite`, I did not find a nice way to do it as a `Fixture` that would go along well with the ES client one.
Especially that it is very likely that we won't test creating this script in other unit tests

delta/plugins/graph-analytics/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsSinkSuite.scala
imsdu: The result is a bit large but there is only two documents in the response and both are important.
About the ES testing, it can be included in our thinking on how we design tests but the way to go could be at some point to rely on `elastic4s` at least for the tests

imsdu: In the long run, we should avoid to compare json payloads as much as we do

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchSink.scala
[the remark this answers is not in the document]
shinyhappydan: I meant map in a more general sense.. taking the results and "mapping" them to each element

It's a hard one to name maybe

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsResult.scala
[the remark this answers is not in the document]
shinyhappydan: That is better to me. I like to think of comments as only needed when the code cannot explain itself

shinyhappydan (dismissed):

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsResult.scala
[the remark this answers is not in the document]
olivergrabinski: This is a comment style that's used throughout the codebase, so maybe we should agree on some guideline. I know I am also responsible for some self-explaining comments, which I have added to keep in line with what was already here.

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsResult.scala
```diff
@@ -0,0 +1,73 @@
...
+    * @param types
+    *   the types of the resource
+    */
+  final case class UpdateByQuery(id: Iri, types: Set[Iri]) extends GraphAnalyticsResult
+
+  /**
+    * The value is indexed to Elasticsearch and an update by query action is required so that the type of this resource
+    * is propagated to the resources pointing to it.
+    *
+    * The other fields are metadata and matches the same definition as in [[ResourceState]].
```
imsdu, line 40: Proposal to avoid repeating myself

delta/plugins/graph-analytics/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsSinkSuite.scala
```diff
@@ -0,0 +1,172 @@
...
+    for {
+      _         <- sink(chunk).assert(chunk.map(_.void))
+      // The reference to file1 should have been resolved and introduced as a relationship
+      // The update query should not have an effect on the other resource
+      expected1 <- ioJsonContentOf("result/resource1_updated.json")
+      expected2 <- ioJsonContentOf("result/resource2.json")
+      _         <- client.count(index.value).eventually(2L)
+      _         <- client.getSource[Json](index, resource1.toString).eventually(expected1)
+      _         <- client.getSource[Json](index, resource2.toString).eventually(expected2)
+    } yield ()
```
shinyhappydan, line 169: I think here it's probably helpful to clarify in variable names what things are doing. If we called `resource3` `resourceWithRelationshipToResource1` then it's much more understandable what the test is doing.

`resource1_updated.json` could become `resource1_with_extra_relationship`
(or, `expected1` could become `resource1WithExtraRelationship`

Even better, we could say `client.getSource[Json](index, resource1.toString) should haveRelationshipTo(resource3)`. This would really spell out what the test is about

Another point, I don't think it's necessary to have negative assertions. And whenever negative assertions are in an eventually block, they might not work anyway

olivergrabinski (approved):

PR #3744, opened 2023-03-16
Add GraphAnalyticsStream

imsdu:
Part of #3415

* Fetches resources and files
* Change relationship resolution: batch thems to avoid to multiply requests to the db


delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsStream.scala
```diff
@@ -0,0 +1,106 @@
...
+    val groupIds = NonEmptyList.fromList(ids.toList).map(_.grouped(batchSize).toList)
+    val noop     = UIO.pure(emptyMapType)
+    groupIds.fold(noop) { list =>
+      list.foldLeftM(emptyMapType) { case (acc, l) =>
+        query(project, xas)(l).map(_ ++ acc)
+      }
+    }
+  }
+
+  // $COVERAGE-OFF$
```
imsdu, line 78: I don't test that whole part as a unit test here:
* The subparts are already tested independently
* It would need to set up the resources and the files, so the storages and so on for little value
This will be tested through integration tests

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/model/JsonLdDocument.scala
```diff
@@ -41,37 +42,34 @@ object JsonLdDocument {
...
     * @param expanded
     *   the expanded JSON-LD to analyze
-    * @param findRelationship
+    * @param findRelationships
     *   the function to apply to resolve resources/files
     */
   def fromExpanded(
       expanded: ExpandedJsonLd,
-      findRelationship: Iri => UIO[Option[Set[Iri]]]
+      findRelationships: Set[Iri] => UIO[Map[Iri, Set[Iri]]]
```
imsdu, line 58: What changes is that we now attempt to resolve all references in the end instead of one by one
Otherwise the code is the same

delta/plugins/graph-analytics/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsStreamSuite.scala
```diff
@@ -0,0 +1,100 @@
...
+      _ <- findRelationships(project2, Set.empty).assert(Map.empty)
+    } yield ()
+
+  }
+
+}
+
+object GraphAnalyticsStreamSuite {
+
+  final private case class Sample(project: ProjectRef, id: Iri, types: Set[Iri]) extends ScopedState {
```
imsdu, line 71: A simple resource to avoid to avoid the complexity of the real ones

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsStream.scala
```diff
@@ -0,0 +1,106 @@
...
+}
+
+object GraphAnalyticsStream {
+
+  private val emptyMapType = Map.empty[Iri, Set[Iri]]
+
+  /**
+    * We look for a resource with these ids in this project and return their types if it can be found
+    */
+  private def query(project: ProjectRef, xas: Transactors)(nel: NonEmptyList[Iri]): UIO[Map[Iri, Set[Iri]]] = {
```
shinyhappydan, line 43: `queryTypesOfResources`?

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsStream.scala
```diff
@@ -0,0 +1,106 @@
...
+        l.foldLeft(emptyMapType) { case (acc, (id, json)) =>
+          acc ++ json.as[Set[Iri]].toOption.map(id -> _)
+        }
+      }
+  }.hideErrors
+
+  /**
+    * We batch ids and looks for existing resources
+    */
+  private[indexing] def findRelationships(project: ProjectRef, xas: Transactors, batchSize: Int)(
```
shinyhappydan, line 66: It's maybe better to say types of resources here, because relationships is much more general

shinyhappydan: This might be true of everywhere we are saying relationships. It's only one type of relationship

shinyhappydan (approved):
I don't see any problems, but I'm not sure I would notice some things. The code you are changing seems quite complicated :)

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsStream.scala
imsdu: It is a private method and the comment is enough no ?

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsStream.scala
imsdu: It is not only about types, the fact that one of the input ids is returned also means the relationships is resolved

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsStream.scala
shinyhappydan: For it to be easy to read for someone else I think the method name needs more description. For someone who has never seen this class before it will really help

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/GraphAnalyticsStream.scala
shinyhappydan: But is it only the relationships between a resource and the type of that resource, so perhaps something more specific

Issue #3681, opened 2023-02-13
Revalidating a resource with its schema


imsdu:
**Motivation**
If the schema of resource is updated, a resource will be out of sync with it and may not pass validation anymore

**Acceptance criteria**
* This new operation can be called on existing resources
* This operation does NOT update the resource
* The endpoint is:
```
GET /v1/resources/{org_label}/{project_label}/{schema_id}/{resource_id}/validate
```
* The operation return the SHACL validation report with a `200` code if the validation succeeds, `400` otherwise

PR #3749, opened 2023-03-17
#3681 Add resource validation endpoint

shinyhappydan:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourceValidator.scala
```diff
@@ -0,0 +1,75 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceF
+import ch.epfl.bluebrain.nexus.delta.sdk.resolvers.ResolverResolution.ResourceResolution
+import ch.epfl.bluebrain.nexus.delta.sdk.resources.model.ResourceRejection
+import ch.epfl.bluebrain.nexus.delta.sdk.resources.model.ResourceRejection.{InvalidJsonLdFormat, InvalidResource, InvalidSchemaRejection, ReservedResourceId, ResourceShaclEngineRejection, SchemaIsDeprecated}
+import ch.epfl.bluebrain.nexus.delta.sdk.schemas.model.Schema
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.ResourceRef.Latest
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.{ProjectRef, ResourceRef}
+import monix.bio.IO
+
+final class ResourceValidator(resourceResolution: ResourceResolution[Schema])(implicit
```
imsdu, line 19: `ValidateResource` ? 
That  is the name pattern we already use for elasticsearch views for example
Also if you make it a trait, it can make some of the test easier as you will get a SAM (Single abstract Method) and won't need a dependance on `ResourceResolution` or `JsonLdApi` anymore

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourceValidator.scala
```diff
@@ -0,0 +1,75 @@
...
+    jsonLdApi: JsonLdApi
+) {
+  private def toGraph(id: Iri, expanded: ExpandedJsonLd): IO[ResourceRejection, Graph] =
+    IO.fromEither(expanded.toGraph).mapError(err => InvalidJsonLdFormat(Some(id), err))
+
+  def validationReport(
```
imsdu, line 25: If you go for the SAM path, it can just be an apply method

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourceValidator.scala
```diff
@@ -0,0 +1,75 @@
...
+      projectRef: ProjectRef,
+      schemaRef: ResourceRef,
+      caller: Caller,
+      resourceId: Iri,
+      expanded: ExpandedJsonLd
+  ): IO[ResourceRejection, (Option[ValidationReport], ResourceRef.Revision, ProjectRef)] =
```
imsdu, line 31: A case class for the return method would be better

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourceValidator.scala
shinyhappydan: I am planning to change this into two different calls, but it's possible yes

shinyhappydan: I was thinking this: 
`def validationReport(
                        projectRef: ProjectRef,
                        schemaRef: ResourceRef,
                        caller: Caller,
                        resourceId: Iri,
                        expanded: ExpandedJsonLd
                      ): IO[ResourceRejection, Option[ValidationReport]]`

and

`def validate(
                        projectRef: ProjectRef,
                        schemaRef: ResourceRef,
                        caller: Caller,
                        resourceId: Iri,
                        expanded: ExpandedJsonLd
                      ): IO[ResourceRejection, (ResourceRef.Revision, ProjectRef)]`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -267,40 +280,24 @@ object Resources {
...
-                        .mapError(ResourceShaclEngineRejection(id, schemaRef, _))
-          _        <- IO.when(!report.isValid())(IO.raiseError(InvalidResource(id, schemaRef, report, expanded)))
-        } yield (ResourceRef.Revision(schema.id, schema.rev), schema.value.project)
+    ): IO[ResourceRejection, (ResourceRef.Revision, ProjectRef)] = {
+      for {
+        (_, rev, ref) <- resourceValidator.validationReport(projectRef, schemaRef, caller, id, expanded)
+      } yield {
+        (rev, ref)
+      }
+    }
```
shinyhappydan, line 300: This code here is why it would be nicer to have two validation methods for the two different use cases

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/builders/ShapeBuilder.scala
```diff
@@ -0,0 +1,63 @@
...
+      ) ++ optionalField("targetClass", targetClasses) ++ optionalField("datatype", datatype) ++ optionalField(
+        "maxInclusive",
+        maxInclusive
+      )
+    )
+  }
+
+  def datatype(datatype: String)(implicit @unused ev: A <:!< HasDatatype): ShapeBuilder[A with HasDatatype] = {
+    this.copy(datatype = Some(datatype))
+  }
```
shinyhappydan, line 54: I'm using this type bound to stop a test author setting a datatype twice, which would be a mistake. This is why type bounds are used in these builder classes

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -180,6 +180,20 @@ final class ResourcesRoutes(
...
                         }
                       },
+                      (pathPrefix("validate") & get & pathEndOrSingleSlash & idSegmentRef(id)) { id =>
+                        operationName(s"$prefixSegment/resources/{org}/{project}/{schema}/{id}/validate") {
+                          authorizeFor(ref, Read).apply {
+                            emit(
+                              OK,
+                              resources
+                                .validate(id, ref, schemaOpt)
+                                .map(_.map(_.json))
```
imsdu, line 190: It would be nice to return Json-LD and not json
So if you return  a `ValidationResult`, you will need a `JsonLdEncoder` there.

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -394,6 +395,27 @@ class ResourcesRoutesSpec extends BaseRouteSpec {
       }
     }
 
+    "validate a resource successfully against its latest schema" in {
```
imsdu, line 398: We need a test to get the result when we validate with the special `unconstrained` case

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -394,6 +395,27 @@ class ResourcesRoutesSpec extends BaseRouteSpec {
...
+      Get("/v1/resources/myorg/myproject/_/myid2/validate") ~> routes ~> check {
+        status shouldEqual StatusCodes.OK
+        response.asJson should have(`@type`("sh:ValidationReport"))
+      }
+    }
+
+    "validate a resource against a schema that does not exist" in {
+      Get("/v1/resources/myorg/myproject/pretendschema/myid2/validate") ~> routes ~> check {
+        status shouldEqual StatusCodes.NotFound
+        response.asJson should have(`@type`("InvalidSchemaRejection"))
```
imsdu, line 408: I think it is important not to check only for the `@type` and make sure the details are also properly serialized

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -394,6 +395,27 @@ class ResourcesRoutesSpec extends BaseRouteSpec {
...
+      }
+    }
+
+    "validate a resource that does not exist" in {
+      Get("/v1/resources/myorg/myproject/_/pretendresource/validate") ~> routes ~> check {
+        status shouldEqual StatusCodes.NotFound
+        response.asJson should have(`@type`("ResourceNotFound"))
```
imsdu, line 415: There should be other tests where we test this resource not found, they could be put together

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourceValidator.scala
imsdu: I think it would be better to have something like this as a return type:
```
sealed trait ValidationResult
case object NoValidation extends ValidationResult
case object Validated(schema, validationReport)
```
It is valuable in the response to have a proper message if there was no validation too.
And if there is one besides having the validation report to have information about the schema (remember that the schema could live in another project)

NB: feel free to change the names,

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -111,6 +106,24 @@ trait Resources {
...
+    *
+    * @param id
+    *   the identifier that will be expanded to the Iri of the resource
+    * @param projectRef
+    *   the project reference where the resource belongs
+    * @param schemaOpt
+    *   the optional identifier that will be expanded to the schema reference to validate the resource. A None value
+    *   uses the currently available resource schema reference.
+    */
+  def validate(
```
imsdu, line 121: Same here, it should return the `ValidationResult`

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImpl.scala
```diff
@@ -98,6 +98,26 @@ final class ResourcesImpl private (
...
 
+  override def validate(id: IdSegmentRef, projectRef: ProjectRef, schemaOpt: Option[IdSegment])(implicit
+      caller: Caller
+  ): IO[ResourceRejection, Option[ValidationReport]] = {
+    for {
+      projectContext <- fetchContext.onRead(projectRef)
+      resourceRef    <- expandIri(id.value, projectContext)
+      schemaRefOpt   <- expandResourceRef(schemaOpt, projectContext)
+      notFound        = ResourceNotFound(resourceRef, projectRef, schemaRefOpt)
+      state          <- id match {
```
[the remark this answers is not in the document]
shinyhappydan: Unfortunately, it's not quite possible. There is a lot of state that we still need, and we need to not run certain validations

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -180,6 +180,17 @@ final class ResourcesRoutes(
                           }
                         }
                       },
+                      (pathPrefix("validate") & get & pathEndOrSingleSlash & idSegmentRef(id)) { id =>
+                        operationName(s"$prefixSegment/resources/{org}/{project}/{schema}/{id}/validate") {
+                          authorizeFor(ref, Read).apply {
```
imsdu, line 185: This should be the `write` permission

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImpl.scala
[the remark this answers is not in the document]
imsdu: Then they maybe could rely on a common method ?

docs/src/main/paradox/docs/delta/api/resources-api.md
```diff
@@ -130,6 +130,26 @@ Response
 :   @@snip [refreshed.json](assets/resources/updated.json)
 
 
+## Validate
+
+This operation runs validation of a resource against a schema. This would be useful to test whether resources would
+match the shape of a new schema. 
+
+This is equivalent of doing an update with the same source as the last revision of the resource. It is useful when the
```
imsdu, line 138: This needs to be updated

imsdu (changes requested):

PR #3756, opened 2023-03-21
Handle untranslatable `\u0000` character during migration

olivergrabinski:
PostgreSQL does not support JSONs text fields that contain `\u0000`. As a fix, when this error occurs, we try to simply remove this value from the JSON.

### DIFF
--- a/delta/migration/src/test/resources/resource.json
+++ b/delta/migration/src/test/resources/resource.json
@@ -1,0 +1,15 @@
+{
+  "number": 1,
+  "field": "some text ",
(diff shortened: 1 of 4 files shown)


delta/migration/src/test/resources/resource.json
```diff
@@ -0,0 +1,15 @@
+{
```
imsdu, line 1: It is migration code so it is ok, it will be deleted anyway soon but for this kind of payload, we can use `circe-literal`

Issue #3758, opened 2023-03-22
Make suites a global entity


imsdu:
**Motivation**
Suites (see #3753) are currently defined by configuration.
The downside with that is that it requires a configuration change and a restart.
The idea is to delegate this part to users with the appropriate permission.

**Acceptance criteria**
1. Suites can be created/updated and persisted to database
2. Users with the permission (`suites/read`) can get a list of the existing suites via an api endpoint
3. User with the permission (`suites/write`) can create/update a suite

PR #3759, opened 2023-03-22
Add the ability to define suites in the search plugin

### LINKED ISSUE #3753: Allow to target automatically a subset of projects with Global search
**Motivation**
In some cases, we want to be able to query search on a subset of projects automatically without the client to be aware of these projects belonging to this subset.

**Acceptance criteria**
1. When a user hits an endpoint `/search/query/suite/{key}`, the key value will be substituted by this list of projects
2. This list of projects is then automatically filtered according to the caller acls just like in the general case
3. The input and input of this new endpoint are the same as the general `/search/query` endpoint

**Not in scope**
1. The suites will be handled within a configuration for now and not saved in database (this will be handled in database in a later ticket)

imsdu:
Fixes #3753 

First implementation of suites based on a configuration.
This will change with #3758 

Note that:
* Documentation is skipped and will be done in a second PR
* Same for integration tests (difficult to handle by configuration when project ids are dynamic)

### DIFF
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
@@ -180,6 +180,17 @@ final class ResourcesRoutes(
                           }
                         }
                       },
+                      (pathPrefix("validate") & get & pathEndOrSingleSlash & idSegmentRef(id)) { id =>
+                        operationName(s"$prefixSegment/resources/{org}/{project}/{schema}/{id}/validate") {
+                          authorizeFor(ref, Write).apply {
+                            emit(
+                              resources
+                                .validate(id, ref, schemaOpt)
+                                .leftWiden[ResourceRejection]
+                            )
+                          }
+                        }
+                      },
                       // Fetch a resource original source
                       (pathPrefix("source") & get & pathEndOrSingleSlash & idSegmentRef(id)) { id =>
                         operationName(s"$prefixSegment/resources/{org}/{project}/{schema}/{id}/source") {
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
@@ -80,20 +80,22 @@ class DeltaModule(appCfg: AppConfig, config: Config)(implicit classLoader: Class
 
   make[RemoteContextResolution].named("aggregate").fromEffect { (otherCtxResolutions: Set[RemoteContextResolution]) =>
     for {
-      errorCtx    <- ContextValue.fromFile("contexts/error.json")
-      metadataCtx <- ContextValue.fromFile("contexts/metadata.json")
-      searchCtx   <- ContextValue.fromFile("contexts/search.json")
-      pipelineCtx <- ContextValue.fromFile("contexts/pipeline.json")
-      tagsCtx     <- ContextValue.fromFile("contexts/tags.json")
-      versionCtx  <- ContextValue.fromFile("contexts/version.json")
+      errorCtx      <- ContextValue.fromFile("contexts/error.json")
+      metadataCtx   <- ContextValue.fromFile("contexts/metadata.json")
+      searchCtx     <- ContextValue.fromFile("contexts/search.json")
+      pipelineCtx   <- ContextValue.fromFile("contexts/pipeline.json")
+      tagsCtx       <- ContextValue.fromFile("contexts/tags.json")
+      versionCtx    <- ContextValue.fromFile("contexts/version.json")
+      validationCtx <- ContextValue.fromFile("contexts/validation.json")
     } yield RemoteContextResolution
       .fixed(
-        contexts.error    -> errorCtx,
-        contexts.metadata -> metadataCtx,
-        contexts.search   -> searchCtx,
-        contexts.pipeline -> pipelineCtx,
-        contexts.tags     -> tagsCtx,
-        contexts.version  -> versionCtx
+        contexts.error      -> errorCtx,
+        contexts.metadata   -> metadataCtx,
+        contexts.search     -> searchCtx,
+        contexts.pipeline   -> pipelineCtx,
+        contexts.tags       -> tagsCtx,
+        contexts.version    -> versionCtx,
+        contexts.validation -> validationCtx
       )
       .merge(otherCtxResolutions.toSeq: _*)
   }
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/MigrationModule.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/MigrationModule.scala
@@ -18,11 +18,10 @@ import ch.epfl.bluebrain.nexus.delta.sdk.projects.Projects
 import ch.epfl.bluebrain.nexus.delta.sdk.projects.model.{ProjectCommand, ProjectEvent, ProjectRejection, ProjectState}
 import ch.epfl.bluebrain.nexus.delta.sdk.realms.Realms
 import ch.epfl.bluebrain.nexus.delta.sdk.realms.model.{RealmCommand, RealmEvent, RealmRejection, RealmState}
-import ch.epfl.bluebrain.nexus.delta.sdk.resolvers.ResolverResolution.ResourceResolution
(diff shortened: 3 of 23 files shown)


delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchSpec.scala
```diff
@@ -159,14 +170,28 @@ class SearchSpec
...
+      extractSources(results).toSet shouldEqual project1Documents
+    }
+
+    "search within an unknown suite" in {
+      search.query(Label.unsafe("xxx"), payload, queryParameters)(bob).rejectedWith[UnknownSuite]
+    }
+
+    "search within a suite as Bob" in {
+      val results = search.query(proj2Suite, payload, queryParameters)(bob).accepted
+      extractSources(results).toSet shouldEqual project2Documents
```
shinyhappydan, line 189: I would propose:

`search all indexes on a user with full access`
`search a suite on a user with full access`
`search all indexes on a user with restricted access`
`search a suite on a user with restricted access`

You would need either 3 projects to achieve this I think

As a reader I don't know who "Bob" or "alice" are

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchSpec.scala
```diff
@@ -145,7 +149,14 @@ class SearchSpec
   private val prefix = "prefix"
 
   "Search" should {
-    lazy val search = Search(listViews, aclCheck, esClient, prefix)
+    lazy val search = Search(listViews, aclCheck, esClient, prefix, allSuites)
+
+    val payload         = jobj"""{"size": 100}"""
+    val queryParameters = Query.Empty
```
shinyhappydan, line 155: If you called this "get everything" then the intention is much clearer when reading the test

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/Search.scala
```diff
@@ -69,19 +81,37 @@ object Search {
...
                                  allProjections,
                                  p => ProjectAcl(p.view.project) -> p.projection.permission,
                                  p => projectionIndex(p.projection, p.view.uuid, p.rev, prefix).value
                                )
           results           <- client.search(payload, accessibleIndices, qp)().mapError(WrappedElasticSearchClientError)
         } yield results
-      }
+
+      override def query(payload: JsonObject, qp: Uri.Query)(implicit caller: Caller): IO[SearchRejection, Json] =
+        query(_ => true, payload, qp)
```
shinyhappydan, line 103: Labelling this `allProjects` would show the intent here more clearly

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/model/SearchConfigSpec.scala
```diff
@@ -44,6 +44,12 @@ class SearchConfigSpec extends AnyWordSpecLike with Matchers with Inspectors wit
         |    name = ${defaults.name}
         |    description = ${defaults.description}
         |  }
+        |
+        |  suites {
+        |    my-suite   = [ "myorg/myproject", "myorg/myproject2" ]
+        |    my-suite-2 = [ "myorg2/myproject", "myorg2/myproject" ]
```
shinyhappydan, line 50: I don't think this is used

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/model/SearchConfig.scala
```diff
@@ -3,33 +3,53 @@ package ch.epfl.bluebrain.nexus.delta.plugins.search.model
...
+      case orgStr :: projectStr :: Nil =>
+        (Label(orgStr), Label(projectStr))
+          .mapN(ProjectRef(_, _))
+          .leftMap(err => CannotConvert(value, classOf[ProjectRef].getSimpleName, err.getMessage))
+      case _                           =>
+        Left(CannotConvert(value, classOf[ProjectRef].getSimpleName, "Wrong format"))
+    }
+  }
+
+  type Suite = Map[Label, Set[ProjectRef]]
```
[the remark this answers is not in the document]
olivergrabinski: I think it is rather supposed to be a Map of several suites

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/model/SearchConfig.scala
```diff
@@ -3,33 +3,53 @@ package ch.epfl.bluebrain.nexus.delta.plugins.search.model
...
 
 import java.nio.file.{Files, Path}
 import scala.util.Try
 
 final case class SearchConfig(
     indexing: IndexingConfig,
     fields: Option[JsonObject],
-    defaults: Defaults
+    defaults: Defaults,
+    suites: Map[Label, Set[ProjectRef]]
```
olivergrabinski, line 27: Type alias could be used?

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/model/SearchConfig.scala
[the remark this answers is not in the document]
imsdu: Yep, I renamed it with a plural to make it clearer

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/Search.scala
shinyhappydan: I meant this: `_ => true`

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchSpec.scala
```diff
@@ -159,14 +170,28 @@ class SearchSpec
...
+      val results = search.query(proj2Suite, matchAll, noParameters)(bob).accepted
+      extractSources(results).toSet shouldEqual project2Documents
     }
 
-    "search only indices the user has access to" in {
-      val results = search.query(jobj"""{"size": 100}""", Query.Empty)(alice).accepted
-      extractSources(results).toSet shouldEqual createDocuments(projectionProj1).toSet
+    "search within a suite accordingly to Alice's restricted access" in {
+      val results = search.query(proj2Suite, matchAll, noParameters)(alice).accepted
+      extractSources(results).toSet shouldEqual Set.empty
```
shinyhappydan, line 194: I think it's better to have a positive assertion here, otherwise this could pass if a user with restricted access always returns no results

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/model/SearchConfig.scala
```diff
@@ -38,6 +58,7 @@ object SearchConfig {
       query         <- loadSparqlQuery(pluginConfig.getString("indexing.query"))
       context       <- loadOption(pluginConfig, "indexing.context", loadExternalConfig[JsonObject])
       defaults      <- loadDefaults(pluginConfig)
+      suites        <- IO.fromEither(suiteSource.load[Suites]).mapError(InvalidSuites)
```
shinyhappydan, line 61: the `loadXXX` methods are making this very readable, it's probably better to match that style

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchSpec.scala
```diff
@@ -159,15 +174,38 @@ class SearchSpec
...
+
+    List(
+      (allSuite, project1Documents),
+      (proj2Suite, Set.empty)
+    ).foreach { case (suite, expected) =>
+      s"search within suite $suite accordingly to Alice's restricted access" in {
+        val results = search.query(suite, matchAll, noParameters)(alice).accepted
+        extractSources(results).toSet shouldEqual expected
+      }
+    }
```
shinyhappydan, line 209: The test you had previously was much better than this, I think if you just add a third project then you can easily cover all cases without needing this

shinyhappydan (approved):

PR #3769, opened 2023-03-28
Add the ability to search by id or _self via the listing API

### LINKED ISSUE #3747: Adding the ability to find a resource by its `@id` or its `_self` from the listing API
# Description
In order to support Fusion give access to more meaningful search capabilities spanning all defaultElasticSearch Views, we would like to add support for new fields that are already indexed.

# `@id` OR `_self`
A new `locate` parameter will be introduced.
This parameter should be an Iri.

This new parameter will allow to filter results by the properties `@id` OR `_self` with an exact match.

TODO: maybe `qualifier` would be a better name ?

imsdu:
Fixes #3747


delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchQuerySuite.scala
shinyhappydan, line 1: This test is really hard to follow

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsDirectivesSpec.scala
```diff
@@ -68,14 +82,45 @@ class ElasticSearchViewsDirectivesSpec
...
+          IncludedType(iri"${vocab}A"),
+          IncludedType(iri"${vocab}B"),
+          ExcludedType(iri"${vocab}C")
+        ),
+        schema = Some(ResourceRef.Latest(iri"${base}mySchema")),
+        q = Some("something")
+      )
+
+      Get(s"/search/org/project?$query") ~> Accept(`*/*`) ~> route ~> check {
+        response.as[ResourcesSearchParams] shouldEqual expected
```
shinyhappydan, line 117: I think if you parse with an internal class then you won't catch breaking output changes

shinyhappydan: You can achieve something similar by making a function which constructs the expected CSV

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchQuerySuite.scala
imsdu: It was even worse before if you look at the previous class :)
It is coming from the fact that we test indexing views, aggregate views that can point to other aggregate views, ...
I think the test would benefit greatly of a redesign of the code that is tested but it is a bit out of scope of this PR

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsDirectivesSpec.scala
imsdu: The class is not internal, only the codec is.
The class contains options and an array so CSV does not really fit.
Besides, it would be better to keep the door leading to CSV closed and locked, we already have enough formats to support.

shinyhappydan: `ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.model.ResourcesSearchParams` is this not a class used by the server?

I'm confused because the previous test was asserting that the result was a CSV. Are you getting back a different response type now?

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchQuerySuite.scala
shinyhappydan: I think design is a thing that can be looked at, but I don't think we're prevented from making it more readable if we're already re-writing the test. I think both before and after they are very difficult to follow

docs/src/main/paradox/docs/delta/api/resources-api.md
```diff
@@ -327,6 +335,8 @@ GET /v1/resources?from={from}
 
 - `{from}`: Number - is the parameter that describes the offset for the current query; defaults to `0`
 - `{size}`: Number - is the parameter that limits the number of results; defaults to `20`
+- `{locate}`: Iri - can be used to find a resource by its `@id` or its address (`_self`)
+- `{id}`: Iri - can be used to find a resource by its `@id`
```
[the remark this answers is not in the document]
imsdu: Ok, so the idea iwould be to let it and mark it as deprecated and delete it in 1.9.
I will remove it from the docs then

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsDirectivesSpec.scala
imsdu: Sorry, I thought you meant internal to the test class.
Here, the test is about test the directive, not the route.
The approach when testing a directive is to insert it in a bogus route which exists only for this test.
The CSV choice was only made in the context of the text.
I changed that for a json with circe codecs because it makes things clearer for a data structure with options and a list

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchQuerySuite.scala
imsdu: I already created a ticket for reworking the implementation of `ElasticSearchViewsQuery`:
https://github.com/BlueBrain/nexus/issues/3770
We can extend the scope of this refactoring and would do this then ?
I wouldn't spend too much more time on it for now as we won't get clear gains without this.

shinyhappydan (approved):

olivergrabinski (approved):

PR #3804, opened 2023-04-05
Add project deletion plugin

shinyhappydan:
Probably needs a single integration test as the unit test misses a lot of integration points


delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,164 @@
...
+    def `<&&>`(right: Task[Boolean]): Task[Boolean] = {
+      if (left) {
+        right
+      } else {
+        Task.pure(false)
+      }
+    }
+  }
+
+  def projectDeletionPass(
```
imsdu, line 59: Usually when there is a function which allows to do actions on a list/stream of elements, it makes things easier to decompose it with a function that processes a single element.
It could look like this:
```scala
def processProject(project) =
   isCandidateForDeletion(project).flatMap {
     case true => deleteProject(project) >> log(s"$project has automatically been marked deleted")
     case false => Task.unit
  }

def processProjects(projects) =
   projects.traverse(processProject) // <<-- Traverse FTW
```

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,164 @@
...
+      allProjects: UIO[Seq[ProjectResource]],
+      deleteProject: ProjectResource => IO[ProjectRejection, Unit],
```
imsdu, line 61: We need to have an UIO there to properly handle if a deletion fails.
For example, it may fail if it is referenced by other projects (via a resolver for example)
In this case, we just handle the error by logging the error.

The reason behind that is because if it fails for this reason:
* This is due to a configuration that must be fixed
* We don't want this to fail as it will block other projects to be eventually marked for deletion

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    def allProjectsNotAlreadyDeleted: IO[Nothing, Seq[ProjectResource]] = {
+      allProjects
+        .map(_.filter(!_.value.markedForDeletion))
+    }
+
+    def shouldBeDeleted(now: Instant): ProjectResource => Task[Boolean] = { pr =>
+      {
+        (isIncluded(pr) && notExcluded(pr)) <&&> (deletableDueToDeprecation(pr) <||> deletableDueToBeingIdle(pr, now))
```
imsdu, line 97: If all the conditions returns a `UIO[Boolean]`,  cats can help there with something like (the or condition can be handled before):
```
implicit val semigroup: Semigroup[Boolean] = (x: Boolean, y: Boolean) => x && y
val isCandidate = (isIncluded(pr), notExcluded(pr), ...).reduce
```

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,164 @@
...
+        Task.pure(false)
+      }
+    }
+  }
+
+  def projectDeletionPass(
+      allProjects: UIO[Seq[ProjectResource]],
+      deleteProject: ProjectResource => IO[ProjectRejection, Unit],
+      config: ProjectDeletionConfig,
+      lastEventTime: ProjectResource => Task[Instant]
```
imsdu, line 63: Same, this should be an UIO.
In the end, you should not have any Task in the conditions as none of them should fail as this would prevent to process other projects

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,164 @@
...
+  def started(
+      projects: Projects,
+      config: ProjectDeletionConfig,
+      projectStatistics: ProjectsStatistics
+  ): Task[ProjectDeletionPlugin] = {
+
+    def lastEventTime(pr: ProjectResource): Task[Instant] = {
+      projectStatistics.get(pr.value.ref).flatMap {
+        case Some(stats) => IO.pure(stats.lastEventTime)
+        case None        => IO.raiseError(new IllegalArgumentException(s"Project '${pr.value.ref}' was not found."))
```
imsdu, line 131: We don't want to raise an error there (even if it should never happen)

imsdu (changes requested):

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
shinyhappydan: Should we use `now` if there is no stat?

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
shinyhappydan: The problem here is that we have && and || behaving differently, however both would be viable semigroup implementations for boolean. I think the solution here is clear, although I do agree it should be `UIO[Boolean]`

delta/plugins/project-deletion/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionSpec.scala
```diff
@@ -0,0 +1,193 @@
...
+import monix.bio.{IO, UIO}
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+import java.time.{Duration, Instant}
+import scala.collection.mutable
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
+import scala.util.matching.Regex
+
+class ProjectDeletionSpec extends AnyWordSpec with Matchers with IOValues {
```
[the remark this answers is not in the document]
shinyhappydan: Is it possible to use scalatest assertions? I think they are important for readability of the tests.

Processing projects individually does make sense I think

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,164 @@
...
+
+    def projectIsIdle(pr: ProjectResource, now: Instant) = {
+      (now diff pr.updatedAt).toSeconds > config.idleInterval.toSeconds
+    }
+
+    def resourcesAreIdle(pr: ProjectResource, now: Instant) = {
+      lastEventTime(pr).map(_.isBefore(now.minus(Duration.ofMillis(config.idleInterval.toMillis))))
+    }
+
+    def allProjectsNotAlreadyDeleted: IO[Nothing, Seq[ProjectResource]] = {
```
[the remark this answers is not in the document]
shinyhappydan: Not sure what you mean by this, could you expand?

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,164 @@
...
+    def deleteProjects(projects: Seq[ProjectResource]): Task[Unit] = {
+      Task
+        .traverse(projects)(
+          deleteProject(_).void.mapError(e => new RuntimeException(s"error deleting project from plugin: ${e}"))
+        )
+        .void
+    }
+
+    for {
+      allProjects      <- allProjectsNotAlreadyDeleted
```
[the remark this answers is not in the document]
shinyhappydan: I'm not sure how this is a second pass, could you explain?

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  def projectDeletionPass(
+      allProjects: UIO[Seq[ProjectResource]],
+      deleteProject: ProjectResource => IO[ProjectRejection, Unit],
+      config: ProjectDeletionConfig,
+      lastEventTime: ProjectResource => Task[Instant]
+  ): Task[Unit] = {
+
+    def isIncluded(pr: ProjectResource): Boolean = {
```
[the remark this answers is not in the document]
imsdu: Let's first see how it looks like after introducing the processing per project :)

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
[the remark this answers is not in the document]
imsdu: We should rather have a
```
def projectIsNotAlreadyMarked(pr: ProjectResource)
```
to combine with the others

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
imsdu: I know but the idea is to combine the or before with:
```
val idleOrDeprecated = ???
val isCandidate = (isIncluded(pr), notExcluded(pr), ..., idleOrDeprecated).reduce
```

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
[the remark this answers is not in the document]
imsdu: Projects are filtered a first time here and then filtered again with the other conditions
If the `markAsDeleted` condition is handled like the other conditions we can only filter once

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
imsdu: I think this is a good way to handle this yes.
When this happens, we should also log it

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    def deleteProject(pr: ProjectResource): IO[ProjectRejection, Unit] = {
+      implicit val caller: Subject = Identity.Anonymous
+      projects
+        .delete(pr.value.ref, pr.rev)
+        .void
+    }
+
+    val continuousStream = Stream
```
imsdu, line 153: I forgot about it but we need the supervisor here to run the stream so that it gets restarted if it fails (due for example to a Network error)

imsdu (changes requested):

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
shinyhappydan: Yeah this is currently done in a weird way by using recover and passing back the original stream.. Do you have an example of where we use a supervisor I can look at?

olivergrabinski: One simple example is `EventMetricsProjection`. Or you can look for usages of the `Supervisor.run` function

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeleter.scala
```diff
@@ -0,0 +1,99 @@
...
+      if (left) {
+        right
+      } else {
+        UIO.pure(false)
+      }
+    }
+  }
+}
+
+class ProjectDeleter(
```
imsdu, line 52: We don't really need a class here, everything could be handled in the object

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,97 @@
...
+        .map(_.results)
+        .map(_.map(_.source))
+    }
+
+    def deleteProject(pr: ProjectResource): UIO[Unit] = {
+      implicit val caller: Subject = Identity.Anonymous
+      projects
+        .delete(pr.value.ref, pr.rev)
+        .void
+        .onErrorHandle(e => logger.error(s"error deleting project from plugin: ${e}"))
```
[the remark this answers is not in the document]
shinyhappydan: What is the difference here?

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,97 @@
...
+      .fixedRate[Task](config.idleCheckPeriod)
+      .evalMap(_ => projectDeletionPass(allProjects, deleteProject, config, lastEventTime))
+      .drain
+
+    val compiledProjection =
+      CompiledProjection.fromStream(projectionMetadata, ExecutionStrategy.TransientSingleNode, _ => continuousStream)
+
+    supervisor
+      .run(compiledProjection)
+      .map(_ => new ProjectDeletionPlugin(supervisor.destroy(projectionMetadata.name).void))
```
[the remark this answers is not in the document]
shinyhappydan: So the plugin stop method should just do nothing?

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeleter.scala
```diff
@@ -0,0 +1,99 @@
...
+import ch.epfl.bluebrain.nexus.delta.kernel.syntax.instantSyntax
+import ch.epfl.bluebrain.nexus.delta.kernel.utils.IOUtils
+import ch.epfl.bluebrain.nexus.delta.plugins.projectdeletion.model.ProjectDeletionConfig
+import ch.epfl.bluebrain.nexus.delta.sdk.ProjectResource
+import monix.bio.UIO
+import ProjectDeleter._
+
+import java.time.{Duration, Instant}
+
+object ProjectDeleter {
```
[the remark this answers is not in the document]
shinyhappydan: I don't think this makes sense as a class name, which would usually be a verb

delta/plugins/project-deletion/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionLogicSpec.scala
```diff
@@ -0,0 +1,221 @@
...
+                             markedForDeletion: Boolean
+  ) {
+    val resource = {
+      val project = ProjectGen.project(
+        orgLabel = org,
+        label = label,
+        markedForDeletion = markedForDeletion
+      )
+
+      ResourceF[Project](
```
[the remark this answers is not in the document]
shinyhappydan: I don't think this method is good, it's taking what is in the project and changing it based on the arguments. We could make another version, but I think this version should not be used

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeleter.scala
```diff
@@ -0,0 +1,99 @@
...
+  }
+}
+
+class ProjectDeleter(
+    deleteProject: ProjectResource => UIO[Unit],
+    config: ProjectDeletionConfig,
+    lastEventTime: (ProjectResource, Instant) => UIO[Instant]
+) {
+
+  def processProject(pr: ProjectResource, now: Instant) = {
```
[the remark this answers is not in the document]
shinyhappydan: I don't think this is a big deal. Especially because we don't have an integration test, I'm not sure of the value in this. The tests are already clear and have good coverage

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeleter.scala
shinyhappydan: I think the encapsulation here provides value

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
```diff
@@ -0,0 +1,97 @@
...
+object ProjectDeletionPlugin {
+
+  private val logger: Logger = Logger[ProjectDeletionPlugin]
+
+  private val projectionMetadata: ProjectionMetadata = ProjectionMetadata("system", "project-deletion", None, None)
+
+  /**
+    * Constructs a stoppable ProjectDeletion process that is started in the background.
+    */
+  def started(
```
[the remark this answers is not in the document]
shinyhappydan: I think we need to talk about this, it doesn't make sense to me for this logic to be in the deleter, as the functionality is different. This feels like the responsibility of the plugin rather than the deletion logic

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionModule.scala
```diff
@@ -31,4 +34,13 @@ class ProjectDeletionModule(priority: Int) extends ModuleDef {
   many[PriorityRoute].add { (route: ProjectDeletionRoutes) =>
     PriorityRoute(priority, route.routes, requiresStrictEntity = true)
   }
+
+  many[Plugin].addEffect {
```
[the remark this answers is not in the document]
shinyhappydan: I think I might need more detail, I don't understand how the search plugin is equivalent to this

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
shinyhappydan: The options are:

`val intermediate = (deletableDueToDeprecation(pr), deletableDueToBeingIdle(pr, now)).reduce(orSemigroup)
(isIncluded(pr), notExcluded(pr), !alreadyDeleted(pr), intermediate).reduce(andSemigroup)`

or

`(isIncluded(pr), notExcluded(pr), !alreadyDeleted(pr), (deletableDueToDeprecation(pr), deletableDueToBeingIdle(pr, now)).reduce(orSemigroup)).reduce(andSemigroup)`

Both of these seem much less readable than the original. Any suggestions for improvement?

imsdu: I would go with the first one without introducing a semigroup for the `or` part as it is a bit overkill for only two conditions

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionModule.scala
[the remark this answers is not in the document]
imsdu: We don't load plugins directly with Distage as there is some class loading logic involved.
What happens is that at startup, we load with the `PluginLoader` the different enabled plugins and then we load and wire together the core module and the plugin modules.

The idea here is also to isolate the classes implementing `PluginDef` and `Plugin` from the business logic like it is done in the other plugins (like the Search one)

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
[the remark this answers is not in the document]
imsdu: We always wrap up logs as it is an effect
For the second one, if we rely on the toString method of the exception, we may lose some information

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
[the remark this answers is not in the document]
imsdu: All of this is already handled in the supervisor, yes

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
shinyhappydan: There needs to be a semigroup for the or because one of them is a UIO

If I didn't have an or semigroup then the semigroup logic would just be inline, which would be even worse I think

IMO the custom operators much more readable than these options, I would rather not change to this

Imagine if we need to change the logic to be more complicated. It's going to be much more difficult if we are are using reduce because the reduces will start getting nested

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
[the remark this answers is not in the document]
shinyhappydan: Ah I didn't see the second part :)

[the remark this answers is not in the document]
shinyhappydan: Does it still make sense if e is a ProjectRejection rather than an exception?

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionModule.scala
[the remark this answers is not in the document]
shinyhappydan: I think I see what you mean :+1:

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeleter.scala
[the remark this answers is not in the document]
imsdu: It could help reduce the boilerplate in the tests

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeleter.scala
imsdu: It looks less consistent that way as it is not the same syntax if you want to process one project or a list of projects

shinyhappydan: Isn't the difference just the new keyword?

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
imsdu: The semigroup is just a way to combine things, there are other alternatives for only two conditions (keep the or semigroup if you want) and we really need a good reason to introduce custom operators.
Semigroups is something we all need to understand.
The logic is not going to get more complicated (we want the conditions for deleting a project easy enough to understand to avoid accidents)

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeleter.scala
imsdu: In the second case, the parameters are scattered and you also need to provide the now

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionPlugin.scala
shinyhappydan: I understand perfectly how semigroups work, but that doesn't change that when presented in terms of semigroups, this simple boolean expression becomes much more difficult to read. I don't see why making a function that encapsulates a really simple concept and enhances readability / changeability is a bad thing worthy of not merging this PR.

I will make the change in order to get this PR merged

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionRunner.scala
```diff
@@ -0,0 +1,98 @@
...
+      .map(_.results)
+      .map(_.map(_.source))
+  }
+
+  private def deleteProject(pr: ProjectResource): UIO[Unit] = {
+    implicit val caller: Subject = Identity.Anonymous
+    projects
+      .delete(pr.value.ref, pr.rev)
+      .void
+      .onErrorHandleWith(e => UIO.eval(logger.error(s"Error deleting project from plugin: $e")))
```
imsdu, line 47: `UIO.delay`

delta/plugins/project-deletion/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ProjectDeletionRunner.scala
```diff
@@ -0,0 +1,98 @@
...
+
+    def possiblyDelete(project: ProjectResource): UIO[Unit] = {
+      shouldDeleteProject(project).flatMap {
+        case true  => deleteProject(project)
+        case false => UIO.unit
+      }
+    }
+
+    for {
+      allProjects <- allProjects
```
imsdu, line 63: allProjects.flatMap(_traverse(possiblyDelete))
(or at least the for comprehension is not needed)

delta/plugins/project-deletion/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ShouldDeleteProjectSuite.scala
```diff
@@ -0,0 +1,199 @@
...
+      markedForDeletion: Boolean
+  ) {
+    val resource = {
+      val project = ProjectGen.project(
+        orgLabel = org,
+        label = label,
+        markedForDeletion = markedForDeletion
+      )
+
+      ResourceF[Project](
```
imsdu, line 124: I think ProjectGen can generate the `ResourceF`

delta/plugins/project-deletion/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ShouldDeleteProjectSuite.scala
```diff
@@ -0,0 +1,199 @@
...
+      idleInterval,
+      idleCheckPeriod = 1.day,
+      deleteDeprecatedProjects,
+      includedProjects,
+      excludedProjects
+    )
+  }
+
+  def addTo(deletedProjects: mutable.Set[ProjectResource]): ProjectResource => UIO[Unit] = { pr =>
+    IO.evalTotal {
```
imsdu, line 171: UIO.delay

imsdu (changes requested):

delta/plugins/project-deletion/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/projectdeletion/ShouldDeleteProjectSuite.scala
shinyhappydan: The method which generates the resource works in a strange way, I don't think it's good for this situation

shinyhappydan: To explain the problem with `ProjectGen`:
```
def resourceFor(
    project: Project,
    rev: Int = 1,
    subject: Subject = Anonymous,
    deprecated: Boolean = false,
    markedForDeletion: Boolean = false
): ProjectResource
```
The problem is that `markedForDeletion` is set in Project, so it’s a duplication here and it’s unclear which one would be used. We should change this method to not take that as an argument, but that feels out of scope for this PR

imsdu (approved):

PR #3823, opened 2023-04-18
log slow blazegraph queries

### LINKED ISSUE #3716: Log slow queries submitted to Blazegraph
**Motivation**
Users run sometimes complex queries that can timeout and in some cases crash the Blazegraph instance.
It would be a good thing to log for these queries:
* The project and view
* The query itself
* How much time it took (or the timeout)
* When it happened
* Who submitted it

**Acceptance criteria**
1. When a query timeouts or exceeds a configured execution time
2. We log it in a dedicated table
3. The queries older than a given ttl will be automatically deleted

shinyhappydan:
Fixes #3716


delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphSlowQueryLogger.scala
```diff
@@ -0,0 +1,42 @@
...
+import scala.concurrent.duration.{Duration, FiniteDuration}
+
+trait BlazegraphSlowQueryLogger  {
+  def logSlowQueries[E, A](context: BlazegraphQueryContext, query: IO[E, A]): IO[E, A]
+}
+object BlazegraphSlowQueryLogger {
+  def noop: BlazegraphSlowQueryLogger = new NoopBlazegraphSlowQueryLogger
+}
+
+class NoopBlazegraphSlowQueryLogger extends BlazegraphSlowQueryLogger {
```
imsdu: Those would be better in the companion object

imsdu (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViewsQuery.scala
```diff
@@ -169,6 +170,8 @@ object BlazegraphViewsQuery {
       )(implicit caller: Caller): IO[BlazegraphViewRejection, R] =
         for {
           view    <- viewsStore.fetch(id, project)
+          p       <- fetchContext.onRead(project)
+          iri     <- expandIri(id, p)
```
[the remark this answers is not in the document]
shinyhappydan: No, because view can be an `AggregateView`

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphSlowQueryStore.scala
```diff
@@ -0,0 +1,61 @@
...
+  def save(query: BlazegraphSlowQuery): Task[Unit]
+}
+
+case class BlazegraphQueryContext(viewId: Iri, project: ProjectRef, query: SparqlQuery, subject: Subject)
+case class BlazegraphSlowQuery(
+    viewId: Iri,
+    project: ProjectRef,
+    query: SparqlQuery,
+    duration: FiniteDuration,
+    occurredAt: Instant,
```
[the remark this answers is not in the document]
shinyhappydan: It's better if stores do nothing but save/load, it's small but still business logic

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryLogger.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import cats.effect.Clock
+import ch.epfl.bluebrain.nexus.delta.kernel.utils.IOUtils
+import ch.epfl.bluebrain.nexus.delta.plugins.blazegraph.BlazegraphQueryContext
+import com.typesafe.scalalogging.Logger
+import monix.bio.{IO, UIO}
+
+import scala.concurrent.duration.{Duration, FiniteDuration}
+
+trait BlazegraphSlowQueryLogger {
```
imsdu, line 11: Drop the slow keyword here too

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryLogger.scala
```diff
@@ -0,0 +1,48 @@
...
+  def logSlowQueries[E, A](context: BlazegraphQueryContext, query: IO[E, A]): IO[E, A]
```
imsdu, line 12: apply is ok here as we have a logger that logs

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryLogger.scala
```diff
@@ -0,0 +1,48 @@
...
+  def logSlowQueries[E, A](context: BlazegraphQueryContext, query: IO[E, A]): IO[E, A] = {
+    query.attempt.timed
+      .flatMap { case (duration, outcome) =>
+        UIO
+          .when(duration >= longQueryThreshold)(logSlowQuery(context, duration))
+          .flatMap(_ => IO.fromEither(outcome))
+      }
+  }
+
+  private def logSlowQuery(context: BlazegraphQueryContext, duration: FiniteDuration): UIO[Unit] = {
```
imsdu, line 41: Can we also have some traditional logging there (without the query not to spam but just a line)

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryLogger.scala
```diff
@@ -0,0 +1,48 @@
...
+    override def logSlowQueries[E, A](context: BlazegraphQueryContext, query: IO[E, A]): IO[E, A] = query
+  }
+  def store(store: BlazegraphSlowQueryStore, longQueryThreshold: Duration)(implicit
+      clock: Clock[UIO]
+  ): BlazegraphSlowQueryLogger = {
+    new BlazegraphSlowQueryLoggerImpl(store, longQueryThreshold)
+  }
+}
+
+class BlazegraphSlowQueryLoggerImpl(store: BlazegraphSlowQueryStore, longQueryThreshold: Duration)(implicit
```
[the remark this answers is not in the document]
shinyhappydan: I don't think that's necessary. Having the class makes discoverability of this code easier and doesn't cost anything

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphPluginModule.scala
```diff
@@ -55,8 +56,22 @@ class BlazegraphPluginModule(priority: Int) extends ModuleDef {
       HttpClient()(cfg.indexingClient, as.classicSystem, sc)
   }
 
+  make[BlazegraphSlowQueryStore].from { (xas: Transactors) =>
```
imsdu, line 59: We don't need the store to be directly accessible there

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphPluginModule.scala
```diff
@@ -55,8 +56,22 @@ class BlazegraphPluginModule(priority: Int) extends ModuleDef {
...
+    BlazegraphSlowQueryStore(
+      xas
+    )
+  }
+
+  make[BlazegraphSlowQueryLogger].from { (cfg: BlazegraphViewsConfig, store: BlazegraphSlowQueryStore) =>
+    new BlazegraphSlowQueryLoggerImpl(store, cfg.slowQueryThreshold)
```
imsdu, line 66: Add an apply method in `BlazegraphSlowQueryLogger` and make `BlazegraphSlowQueryLoggerImpl` private

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViewsQuery.scala
```diff
@@ -87,13 +89,16 @@ trait BlazegraphViewsQuery {
   )(implicit caller: Caller): IO[BlazegraphViewRejection, R]
 }
 
+case class BlazegraphQueryContext(view: ViewRef, query: SparqlQuery, subject: Subject)
```
imsdu, line 92: This should go in the companion object of this class + final

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/config/BlazegraphViewsConfig.scala
```diff
@@ -48,6 +48,7 @@ final case class BlazegraphViewsConfig(
     indexingClient: HttpClientConfig,
     queryClient: HttpClientConfig,
     queryTimeout: Duration,
+    slowQueryThreshold: Duration,
```
imsdu, line 51: Docs (this is valid everywhere in other parts too)

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryLogger.scala
```diff
@@ -0,0 +1,50 @@
...
+      .flatMap { case (duration, outcome) =>
+        UIO
+          .when(duration >= longQueryThreshold)(logSlowQuery(context, duration))
+          .flatMap(_ => IO.fromEither(outcome))
+      }
+  }
+
+  private def logSlowQuery(context: BlazegraphQueryContext, duration: FiniteDuration): UIO[Unit] = {
+    IOUtils.instant
+      .tapEval(_ => UIO.delay(logger.warn(s"slow blazegraph query recorded: duration $duration, view ${context.view}")))
```
imsdu, line 43: Capital letter + we tend to put things like duration and view here between quotes

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryLogger.scala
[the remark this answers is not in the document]
imsdu: Make it private and in the companion object

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryStore.scala
```diff
@@ -0,0 +1,93 @@
...
+          subject.as[Subject] match {
+            case Right(value) => value
+            case Left(e)      => throw e
+          }
+        )
+    }
+  }
+}
+
+case class BlazegraphSlowQuery(
```
imsdu, line 70: To move in a model subpackage + final
We also need to know if the query was able to complete or not

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryStore.scala
```diff
@@ -0,0 +1,93 @@
...
+          .transact(xas.read)
+          .compile
+          .toList
+      }
+    }
+  }
+}
+
+object BlazegraphSlowQuery {
+  implicit val projectionProgressRowRead: Read[ProjectionProgressRow] = {
```
imsdu: To remove

imsdu (changes requested):

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryStore.scala
```diff
@@ -0,0 +1,55 @@
...
+import monix.bio.Task
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.Identity.Database._
+
+import java.time.Instant
+
+trait BlazegraphSlowQuerySink {
+  def save(query: BlazegraphSlowQuery): Task[Unit]
+}
+
+trait BlazegraphSlowQueryStore extends BlazegraphSlowQuerySink {
```
imsdu, line 19: What is the value to separate the save operation ? 
It makes you create two different instances and complicates things even more.

My guess is partly to do the test where we want to write in database and it fails.
I would rather see this check go for the reason that if postgres is down, we don't bother about logging a slow query because we won't even be able to perform the query at all (we can't fetch the acls, we can't fetch the view, etc...)

imsdu (changes requested):

delta/plugins/blazegraph/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryLoggerSuite.scala
```diff
@@ -0,0 +1,131 @@
...
+class BlazegraphSlowQueryLoggerSuite extends BioSuite {
+  private val LongQueryThreshold = 100.milliseconds
+  private val SinkWhichFails: BlazegraphSlowQuerySink = (_: BlazegraphSlowQuery) => Task.raiseError(new RuntimeException("error saving slow log"))
+
+  private val view        = ViewRef(ProjectRef.unsafe("epfl", "blue-brain"), Iri.unsafe("hippocampus"))
+  private val sparqlQuery = SparqlQuery("")
+  private val user        = Identity.User("Ted Lasso", Label.unsafe("epfl"))
+
+  private def inMemorySink(): (BlazegraphSlowQuerySink, () => List[BlazegraphSlowQuery]) = {
+    val saved   = new util.ArrayList[BlazegraphSlowQuery]()
```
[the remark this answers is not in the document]
shinyhappydan: I used ListBuffer because Vector wasn't mutable, but fixed

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryStore.scala
shinyhappydan: Only one instance is created. This interface segregation removes boilerplate from the tests

The postgres check is a feature we could add, but I don't know if it's a valuable optimisation

If postgres is down I'm not sure the query would be slow enough to be logged anyway

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryDeleter.scala
```diff
@@ -0,0 +1,41 @@
...
+
+class BlazegraphSlowQueryDeleter(store: BlazegraphSlowQueryStore, deletionThreshold: FiniteDuration) {
+  def deleteOldQueries: Task[Unit] = {
+    IOUtils.instant.flatMap { now =>
+      store.removeQueriesOlderThan(now.minusMillis(deletionThreshold.toMillis))
+    }
+  }
+}
+
+object BlazegraphSlowQueryDeleter {
```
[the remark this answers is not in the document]
shinyhappydan: With this it would be nice if no class was needed at all. The code located in this class doesn't really have anything to do with logging, so it's clearer if they are separate. Merging is just going to add boilerplate to the tests, as well as coupling these actions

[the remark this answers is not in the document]
imsdu: You can handle this with different apply methods for example one that just initiates the logger and the other that also starts the routine on top of that

delta/plugins/blazegraph/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryStoreSuite.scala
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  test("Remove old queries") {
+    for {
+      _       <- store.save(OldQuery)
+      _       <- store.save(RecentQuery)
+      _       <- store.removeQueriesOlderThan(OneWeekAgo)
+      results <- store.listForTestingOnly(view)
+    } yield {
+      assert(results.contains(RecentQuery), "recent query was deleted")
```
imsdu, line 53: The clues are inverted

shinyhappydan: I think it's correct, if results doesn't contain recent query, then it was deleted

shinyhappydan:
@imsdu I can't reply to the comment directly for some reason, but with regard to the project deletion two apply methods: I tried to do what you've suggested, but I think the situations are just different here? I can't see a way to do it that makes sense. I think in your example, the runner vs noop means that it does make sense, but here we don't have that so we would be casting for no reason. I might have misunderstood, you can show me what you mean when you get back

Issue #3049, opened 2022-01-28
Delete project and related resources


imsdu:
* Stop running streams
* Lock the project during deletion
* Delete files
* Delete entries in DB
* Delete indices in ES
* Delete namespaces in BG

PR #3834, opened 2023-04-20
Add project deletion task for composite views

imsdu:
Related to #3049

### DIFF
--- a/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/deletion/CompositeViewsDeletionTask.scala
+++ b/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/deletion/CompositeViewsDeletionTask.scala
@@ -1,0 +1,50 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.deletion
+
+import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.CompositeViews
+import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.deletion.CompositeViewsDeletionTask.{init, logger}
+import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.indexing.CompositeViewDef
+import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.indexing.CompositeViewDef.{ActiveViewDef, DeprecatedViewDef}
+import ch.epfl.bluebrain.nexus.delta.sdk.deletion.ProjectDeletionTask
+import ch.epfl.bluebrain.nexus.delta.sdk.deletion.model.ProjectDeletionReport
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.Identity.Subject
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.ProjectRef
+import com.typesafe.scalalogging.Logger
+import fs2.Stream
+import monix.bio.{Task, UIO}
+
+final class CompositeViewsDeletionTask(
+    currentViews: ProjectRef => Stream[Task, CompositeViewDef],
+    deprecate: (ActiveViewDef, Subject) => UIO[Unit]
+) extends ProjectDeletionTask {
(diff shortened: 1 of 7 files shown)


delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/deletion/CompositeViewsDeletionTaskSuite.scala
```diff
@@ -0,0 +1,52 @@
...
+  private val viewStream: Stream[Task, CompositeViewDef] =
+    Stream(
+      activeView(active1),
+      DeprecatedViewDef(deprecated),
+      activeView(active2)
+    )
+
+  test("Deprecate all active views for project") {
+    for {
+      deprecated   <- Ref.of[Task, Set[ViewRef]](Set.empty)
```
shinyhappydan, line 43: Small thing, but perhaps this should be a list? Not sure if duplication would indicate an issue

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/deletion/CompositeViewsDeletionTask.scala
```diff
@@ -0,0 +1,50 @@
...
+import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.indexing.CompositeViewDef.{ActiveViewDef, DeprecatedViewDef}
+import ch.epfl.bluebrain.nexus.delta.sdk.deletion.ProjectDeletionTask
+import ch.epfl.bluebrain.nexus.delta.sdk.deletion.model.ProjectDeletionReport
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.Identity.Subject
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.ProjectRef
+import com.typesafe.scalalogging.Logger
+import fs2.Stream
+import monix.bio.{Task, UIO}
+
+final class CompositeViewsDeletionTask(
```
shinyhappydan, line 15: I think it's fine without, but you would usually have a doc here

olivergrabinski (dismissed):

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/deletion/CompositeViewsDeletionTaskSuite.scala
imsdu: Duplication can't happen here as view ref is the identifier of the view and we only process each view only once.

shinyhappydan: Well, the set here represents calls to the function, and you could call the function twice if you had a bug. I agree it's unlikely, but List would be more accurate than Set

shinyhappydan (approved):

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/deletion/CompositeViewsDeletionTaskSuite.scala
imsdu: Things can only be deprecated once

shinyhappydan: `deprecate(view, subject) >> deprecate(view, subject)` would be a bug but this test wouldn't fail

imsdu: The deprecate function in the test is idempotent just like the real one (and the other deletion tasks), so it wouldn't be a bug.

PR #3836, opened 2023-04-21
Add SearchConfigUpdater to update composite view at startup

### LINKED ISSUE #3798: Allow to update search at startup
**Motivation**
Update search configuration could be simplified.
At the moment, configuration has to be updated and an update must be explicitly performed through the Scala CLI.
Using the Scala CLI implies building it locally and retrieving the token to be able to do the API calls.

**Acceptance criteria**
1. When starting Delta on the search plugin, we loop against all composite views in the system
2. We filter out those belonging to deprecated projects
3. We compare the source of the remaining ones with the one loaded with the config and we update the views where we detect a change

olivergrabinski:
Closes #3798


delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdater.scala
```diff
@@ -0,0 +1,97 @@
...
+      }
+      .compile
+      .drain
+      .doOnFinish {
+        case Some(cause) => UIO.delay(logger.error("Updating default composite views failed.", cause.toThrowable))
+        case None        => UIO.delay(logger.info("Stopping stream. All default composite views have been updated."))
+      }
+
+  private def configHasChanged(v: ActiveViewDef)(implicit baseUri: BaseUri): Boolean =
+    CompositeViewFields.fromValue(v.value).toJson(v.ref.viewId) != defaultSearchViewFields.toJson(v.ref.viewId)
```
imsdu, line 54: We can just compare the two CompositeViewFields no ?

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdater.scala
```diff
@@ -0,0 +1,97 @@
...
+
+object SearchConfigUpdater {
+
+  private val logger: Logger = Logger[SearchConfigUpdater]
+
+  /**
+    * Creates a [[SearchConfigUpdater]] and returns the [[Task]] that updates all default composite view that are not in
+    * line with the given search config.
+    */
+  def apply(
```
imsdu, line 69: We need to use the supervisor to run it:
This way it is done asynchronously and does not slow down startup

imsdu (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeView.scala
```diff
@@ -81,11 +80,7 @@ object CompositeView {
   /**
     * Rebuild strategy defining rebuilding at a certain interval.
     */
-  final case class Interval private[model] (value: FiniteDuration) extends RebuildStrategy
-  object Interval {
```
olivergrabinski, line 85: Seemingly dead code

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdateFunctionSuite.scala
```diff
@@ -0,0 +1,127 @@
...
+  test("Sources, projections, and rebuild strategy should be updated") {
+    for {
+      views   <- compositeViews
+      updated <- views.fetch(id, projectRef)
+      _        = assertEquals(updated.rev, 2)
+      _        = assertEquals(updated.value.sources.map(_.toFields), newSources)
+      _        = assertEquals(updated.value.projections.map(_.toFields), newProjections)
+      _        = assertEquals(updated.value.rebuildStrategy, newRebuildStrategy)
+    } yield ()
+  }
```
shinyhappydan, line 89: This is really just one test case. Consider using methods rather than comments to abstract some of the setup, that way it won't be so much boilerplate to read. If it's not clear what I mean I'd be happy to show you

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdaterSuite.scala
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  test("A non-default active view should not be updated") {
+    assert(!updatedViews.exists(_.viewDef == customOutdatedView))
+  }
+
+  test("Deprecated view should not be updated") {
+    assert(!updatedViews.exists(_.viewDef == deprecatedView))
+  }
+
```
shinyhappydan, line 113: It would also be nicer to have each of these test running individually, rather than having to all be ran in order

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdater.scala
imsdu: One of the problem with json is also that it has no notion of set so it could bring bad surprises.

imsdu (changes requested):

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/model/SearchConfig.scala
```diff
@@ -60,6 +62,7 @@ object SearchConfig {
       settings      <- loadOption(pluginConfig, "indexing.settings", loadExternalConfig[JsonObject])
       query         <- loadSparqlQuery(pluginConfig.getString("indexing.query"))
       context       <- loadOption(pluginConfig, "indexing.context", loadExternalConfig[JsonObject])
+      rebuild        = Try(pluginConfig.getDuration("indexing.rebuild-strategy", MILLISECONDS)).toOption
```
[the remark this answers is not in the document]
olivergrabinski: I've set it so it defaults to no strategy in case of problems 🤔

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdater.scala
```diff
@@ -0,0 +1,112 @@
...
+            update(view, defaultSearchViewFields)
+          case _                                             =>
+            Task.unit
+        }
+      }
+
+  private def configHasChanged(
+      v: ActiveViewDef
+  )(implicit eq: Eq[CompositeViewValue] = indexingEq): Boolean =
+    v.value =!= defaultSearchViewValue(v)
```
shinyhappydan, line 53: If the equality is never different then it's probably better to remove it from the method signature, just to keep the surface area smaller

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/model/SearchConfig.scala
```diff
@@ -107,12 +111,31 @@ object SearchConfig {
...
+    * Load the rebuild strategy from the search config. If either of the required fields is null, missing, or not a
+    * correct finite duration, there will be no rebuild strategy. If both finite durations are present, then the rebuild
+    * strategy interval is used only it is greater or equal to the min-interval-rebuild; if not then there is no
+    * strategy.
+    */
+  private def loadRebuildStrategy(config: Config): Option[RebuildStrategy] =
+    for {
+      rebuild            <- readFiniteDuration(config, "indexing.rebuild-strategy")
+      minIntervalRebuild <- readFiniteDuration(config, "indexing.min-interval-rebuild")
+      rebuildStrategy    <- Option.when(rebuild gteq minIntervalRebuild)(Interval(rebuild))
```
imsdu, line 124: It would be better to fail if the condition is not met

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdateFunctionSuite.scala
```diff
@@ -0,0 +1,131 @@
...
+      _       <- viewCreated(views)
+      _       <- updateView(views, viewDef, newViewFields)
+      updated <- views.fetch(id, projectRef)
+      _        = assertEquals(updated.rev, 2)
+      _        = assertEquals(updated.value.sources, newSources.map(_.toSource(uuid, id)))
+      _        = assertEquals(updated.value.projections, newProjections.map(_.toProjection(uuid, id)))
+      _        = assertEquals(updated.value.rebuildStrategy, newRebuildStrategy)
+    } yield ()
+  }
+
```
shinyhappydan, line 94: This is only a suggestion, but this is a little hard to follow because some of the things are happening implicitly. Consider this alternative:

```
for {
      views   <- thereIsAView(viewRef, viewFields)
      _       <- updateView(views, viewRef, newViewFields)
      updated <- views.fetch(viewRef)
      _        = assertEquals(updated.rev, 2)
      _        = assertEquals(updated.value.sources, newSources.map(_.toSource(uuid, id)))
      _        = assertEquals(updated.value.projections, newProjections.map(_.toProjection(uuid, id)))
      _        = assertEquals(updated.value.rebuildStrategy, newRebuildStrategy)
    } yield ()
  }
```

shinyhappydan: The main change here is to be explicit about the view you are creating/updating/reading, it makes reading clearer even though you don't need to pass the values in to make the function work

Similar to this, if the new fields were more obviously about the sources, projections and strategy then that would make it easier to see why the assertions are happening

I'm only posting this to show my perspective, the test is not a problem

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdater.scala
```diff
@@ -0,0 +1,112 @@
...
+    (viewDef, fields) =>
+      views
+        .update(
+          viewDef.ref.viewId,
+          viewDef.ref.project,
+          viewDef.rev,
+          fields
+        )
+        .void
+        .onErrorHandle(e => logger.error(s"Could not update view ${viewDef.ref}. Reason: ${e.reason}"))
```
imsdu, line 111: We need an UIO there

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdateFunctionSuite.scala
```diff
@@ -0,0 +1,131 @@
...
+  /**
+    * Using the SearchConfigUpdater update function update the given [[ActiveViewDef]] using the new
+    * [[CompositeViewFields]]
+    */
+  private def updateView(views: CompositeViews, viewDef: ActiveViewDef, newFields: CompositeViewFields) = {
+    val update = SearchConfigUpdater.update(views)
+    update(viewDef, newFields)
+  }
+
+  test("Sources, projections, and rebuild strategy should be updated") {
```
imsdu, line 82: This is already tested in the composite views test

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdater.scala
```diff
@@ -0,0 +1,112 @@
...
+    */
+  def apply(
+      views: ElemStream[CompositeViewDef],
+      update: (ActiveViewDef, CompositeViewFields) => Task[Unit]
+  ): Stream[Task, Elem[CompositeViewDef]] =
+    views
+      .filter(_.id == defaultViewId)
+      .evalTap { elem =>
+        elem.traverse {
+          case view: ActiveViewDef if configHasChanged(view) =>
```
imsdu, line 43: The test on the id could come here

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdaterSuite.scala
```diff
@@ -0,0 +1,114 @@
...
+  private val update: (ActiveViewDef, CompositeViewFields) => Task[Unit] =
+    (viewDef, fields) => Task { updatedViews.add((viewDef, fields)) }.void
+
+  test("Update the views") {
+    new SearchConfigUpdater(defaults, indexingConfig)(views, update).compile.drain
+  }
+
+  test("A default active view whose sources do not match the current config should be updated") {
+//    assert(updatedViews.contains((defaultOutdatedView, fromValue(upToDateView.value))))
```
imsdu, line 98: To remove

imsdu (changes requested):

delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchConfigUpdaterSuite.scala
```diff
@@ -0,0 +1,117 @@
...
+  test("A default view whose sources do not match the current config should be updated") {
+    for {
+      views   <- compositeViews
+      default <- defaultViewValue
+      _       <- views
+                   .fetch(defaultViewId, proj1.ref)
+                   .map(view => toValue(view.value))
+                   .eventually(default)
+    } yield ()
+  }
```
shinyhappydan, line 101: These tests should really be run independently of each other

olivergrabinski: I made an update

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/model/SearchConfig.scala
```diff
@@ -107,12 +111,35 @@ object SearchConfig {
...
     )
 
+  /**
+    * Load the rebuild strategy from the search config. If either of the required fields is null, missing, or not a
+    * correct finite duration, there will be no rebuild strategy. If both finite durations are present, then the
+    * specified rebuild strategy must be greater or equal to the min rebuild interval.
+    */
+  private def loadRebuildStrategy(config: Config): IO[SearchConfigError, Option[RebuildStrategy]] = {
+    import cats.implicits._
+    val strategy = for {
```
imsdu, line 121: traverse ? :)

imsdu (changes requested):

imsdu (approved):

PR #3840, opened 2023-04-21
deal with missing files in DiskStorageFetchFile 

### LINKED ISSUE #3817: Report a clean error message when a File is missing from Storage
## Description
Current Delta returns a broken stream, we should have a proper HTTP status code returned and a meaningful error message.

shinyhappydan:
Fixed #3817 

previously returned a stream which would break

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/DiskStorageFetchFile.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/DiskStorageFetchFile.scala
@@ -7,7 +7,7 @@ import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.FetchFi
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.StorageFileRejection.FetchFileRejection
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.StorageFileRejection.FetchFileRejection.UnexpectedLocationFormat
 import ch.epfl.bluebrain.nexus.delta.sdk.AkkaSource
-import monix.bio.{IO, UIO}
+import monix.bio.IO
 
 import java.net.URI
 import java.nio.file.Paths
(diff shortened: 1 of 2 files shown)


delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/DiskStorageFetchFileSpec.scala
```diff
@@ -30,6 +31,14 @@ class DiskStorageFetchFileSpec
...
 
+    "deal with a missing file" in {
+      val volume       = Files.createTempDirectory("disk-access")
+      val file         = volume.resolve("my/file.txt")
+      Files.createDirectories(file.getParent)
+      val fullFilePath = file.toString
+
+      DiskStorageFetchFile(Uri.Path(fullFilePath)).rejected shouldEqual FetchFileRejection.FileNotFound(fullFilePath)
+    }
+  }
```
shinyhappydan, line 43: I would have converted this test to MUnit, however it's using an Akka trait and it didn't seem worth it for this story

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/DiskStorageFetchFile.scala
```diff
@@ -20,7 +20,9 @@ object DiskStorageFetchFile extends FetchFile {
 
   override def apply(path: Uri.Path): IO[FetchFileRejection, AkkaSource] =
     Try(Paths.get(URI.create(s"file://$path"))) match {
-      case Failure(err)   => IO.raiseError(UnexpectedLocationFormat(s"file://$path", err.getMessage))
-      case Success(value) => UIO.pure(FileIO.fromPath(value))
+      case Failure(err)  => IO.raiseError(UnexpectedLocationFormat(s"file://$path", err.getMessage))
+      case Success(path) =>
+        IO.raiseWhen(!path.toFile.exists())(FetchFileRejection.FileNotFound(path.toString))
```
shinyhappydan, line 25: Would want to double check this is how we interact with the file system, I wonder if there is a different library we might use. FileIO didn't seem to have this but perhaps I missed it

PR #3843, opened 2023-04-24
reject empty ids when creating a resource

### LINKED ISSUE #3355: Empty ids when creating a resource should be rejected
When passed empty, the `@id` should be considered as invalid
```
{
  "@id": "",
  "@type": "ElasticSearchView",
  "includeDeprecated": false,
  "includeMetadata": true,
  "mapping": {},
  "resourceSchemas": [],
  "resourceTypes": [],
  "sourceAsText": false
}
```

Please note that using the URL segment to query Delta you could also provide an empty ID. This would need to be caught too.

shinyhappydan:
Fixes #3355

### DIFF
--- a/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
+++ b/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
@@ -244,6 +244,23 @@ final class CompositeViews private (
   }.span("deprecateCompositeView")
 
   /**
+    * Deprecates an existing composite view without applying preliminary checks on the project status
+    *
+    * @param id
+    *   the view identifier
+    * @param project
+    *   the view parent project
+    * @param rev
+    *   the current view revision
+    * @param subject
+    *   the subject that initiated the action
+    */
+  private[compositeviews] def internalDeprecate(id: Iri, project: ProjectRef, rev: Int)(implicit
+      subject: Subject
+  ): IO[CompositeViewRejection, Unit] =
+    eval(DeprecateCompositeView(id, project, rev, subject)).void
+
+  /**
     * Retrieves a current composite view resource.
     *
     * @param id
@@ -402,6 +419,12 @@ final class CompositeViews private (
   }
 
   /**
+    * Return all existing views for the given project in a finite stream
+    */
+  def currentViews(project: ProjectRef): ElemStream[CompositeViewDef] =
+    log.currentStates(Predicate.Project(project)).map(toCompositeViewDef)
+
+  /**
     * Return all existing indexing views in a finite stream
     */
   def currentViews: ElemStream[CompositeViewDef] =
@@ -418,11 +441,14 @@ final class CompositeViews private (
       CompositeViewDef(v)
     }
 
+  private def eval(cmd: CompositeViewCommand) =
+    log.evaluate(cmd.project, cmd.id, cmd)
+
   private def eval(
       cmd: CompositeViewCommand,
       pc: ProjectContext
   ): IO[CompositeViewRejection, ViewResource] =
-    log.evaluate(cmd.project, cmd.id, cmd).map(_._2.toResource(pc.apiMappings, pc.base))
+    eval(cmd).map(_._2.toResource(pc.apiMappings, pc.base))
 
 }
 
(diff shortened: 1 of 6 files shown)


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ResourcesSpec.scala
```diff
@@ -113,6 +113,19 @@ class ResourcesSpec extends BaseSpec with EitherValuable with CirceEq {
...
+      deltaClient
+        .post[Json](s"/resources/$id1/_/", jsonContentOf("/kg/resources/simple-resource-with-blank-id.json"), Rick) {
+          (json, response) =>
+            response.status shouldEqual StatusCodes.BadRequest
+            json shouldEqual jsonContentOf(
+              "/kg/resources/invalid-resource-id.json",
+              "id" -> ""
+            )
+        }
+    }
```
shinyhappydan, line 127: Arguably there should be a second test here, however I could not get the route to fire when specifying the id in the url. The best we could do is have a test which says that a resource is not created. This would pass because the route is rejected. Given the issue we are dealing with, it doesn't seem useful to have this test

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdSourceProcessor.scala
```diff
@@ -52,6 +52,12 @@ sealed abstract class JsonLdSourceProcessor(implicit api: JsonLdApi) {
       case payloadIri: Iri => IO.raiseError(UnexpectedId(iri, payloadIri))
     }
 
+  protected def validateIdNotBlank(source: Json): IO[BlankId.type, Unit] = {
+    IO.raiseWhen(
+      source.hcursor.downField("@id").as[String].map(_.trim).contains("")
```
imsdu, line 57:
```
source.hcursor.downField("@id").as[String].exists(_.isBlank)
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImpl.scala
```diff
@@ -43,20 +43,28 @@ final class ResourcesImpl private (
       source: Json
   )(implicit caller: Caller): IO[ResourceRejection, DataResource] = {
     for {
+      _                          <- validateIdNotBlank(source)
```
imsdu, line 46: The `JsonLdSourceProcessor` already takes care of it

tests/src/test/resources/kg/resources/invalid-resource-id.json
```diff
@@ -0,0 +1,5 @@
+{
+  "@context": "https://bluebrain.github.io/nexus/contexts/error.json",
+  "@type": "InvalidResourceId",
+  "reason": "Resource identifier '{{id}}' cannot be expanded to an Iri."
```
imsdu, line 4:
```
Resource identifier '{{id}}' cannot be expanded to an Iri or is empty.
```
So update the message of the rejection or introduce a new one

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ResourcesSpec.scala
```diff
@@ -113,6 +113,19 @@ class ResourcesSpec extends BaseSpec with EitherValuable with CirceEq {
         )
       }
     }
+
+    "fail if the id is blank" in {
```
imsdu, line 117: A test in `ResourcesImplSpec` would be nice too

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveRejection.scala
```diff
@@ -166,6 +166,7 @@ object ArchiveRejection {
     case JsonLdRejection.UnexpectedId(id, sourceId)        => UnexpectedArchiveId(id, sourceId)
     case JsonLdRejection.InvalidJsonLdFormat(id, rdfError) => InvalidJsonLdFormat(id, rdfError)
     case JsonLdRejection.DecodingFailed(error)             => DecodingFailed(error)
+    case JsonLdRejection.BlankId                           => InvalidArchiveId("")
```
olivergrabinski, line 169: Question: all these InvalidId rejections here and below, they are defined to handle the case where the provided `@id` cannot be resolved/expanded to an IRI. Isn't it possible that the empty string can be resolved, and it is just that we decide not to allow it? Then the error would not be entirely correct.

imsdu (changes requested):

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveRejection.scala
shinyhappydan: This is now a new error

imsdu (approved):

PR #3865, opened 2023-05-03
Add integration tests for project deletion

### LINKED ISSUE #3049: Delete project and related resources

imsdu:
Fixes #3049


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala
```diff
@@ -137,11 +137,11 @@ final class ProjectsRoutes(
                     // Deprecate/delete project
                     (delete & pathEndOrSingleSlash) {
                       parameters("rev".as[Int], "prune".?(false)) {
-                        case (rev, true) if config.deletion.enabled =>
+                        case (rev, true)  =>
```
imsdu, line 140: I removed it from there as the deletion can be triggered from the deletion plugin and other places, so I included it in the state machine as it is way safer

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/ProjectsModule.scala
```diff
@@ -59,7 +61,7 @@ object ProjectsModule extends ModuleDef {
       Task.pure(
         ProjectsImpl(
           organizations.fetchActiveOrganization(_).mapError(WrappedOrganizationRejection),
-          ProjectReferenceFinder(xas),
+          ValidateProjectDeletion(xas, config.projects.deletion.enabled),
```
imsdu, line 64: Renamed as we now also check if the deletion is enabled there

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/GraphAnalytics.scala
```diff
@@ -86,7 +86,7 @@ object GraphAnalytics {
     }
 
   private[analytics] def index(prefix: String, ref: ProjectRef): IndexLabel =
-    IndexLabel.unsafe(s"${prefix}_ga_${ref.organization}_${ref.project}}")
+    IndexLabel.unsafe(s"${prefix}_ga_${ref.organization}_${ref.project}")
```
imsdu, line 89: There was an extra curly brace 🎪

delta/plugins/project-deletion/src/main/resources/project-deletion.conf
```diff
@@ -5,7 +5,7 @@ plugins.project-deletion {
   priority = 8
   # after which period of inactivity are projects considered for deletion; activity is defined as any event scoped to
   # the project recorded in the event log
-  idle-interval = 3 months
+  idle-interval = 90 days
```
imsdu, line 8: month is not a valid keyword

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/deletion/ProjectDeletionStore.scala
```diff
@@ -20,6 +20,7 @@ final private[deletion] class ProjectDeletionStore(xas: Transactors) {
   def deleteAndSaveReport(report: ProjectDeletionReport): Task[Unit] =
     (
       deleteProject(report.project) >>
+        EntityDependencyStore.deleteAll(report.project) >>
```
imsdu, line 23: There was another bug there

tests/docker/config/delta-postgres.conf
```diff
@@ -10,10 +10,18 @@ app {
         host = "postgres"
       }
     }
+
+    query {
```
imsdu, line 14: To make tests faster

tests/docker/config/delta-postgres.conf
```diff
@@ -10,10 +10,18 @@ app {
...
+      refresh-strategy = 1s
+    }
   }
 
   database {
     tables-autocreate = true
+
+    cache {
+      expire-after = 1s
```
imsdu, line 23: Same

tests/docker/docker-compose.yml
```diff
@@ -14,10 +14,14 @@ services:
...
     image: bluebrain/nexus-delta:latest
-    entrypoint: [ "bin/wait-for-it.sh", "-s", "-t", "0", "postgres:5432", "--",
-                  "bin/wait-for-it.sh", "-s", "-t", "0", "elasticsearch:9200", "--",
-                  "./bin/delta-app",
-                  "-Xmx512m" ]
+    entrypoint:
+      - '/bin/bash'
+      - '-c'
+      - |
+        ln -s /opt/docker/plugins/disabled/project-deletion.jar /opt/docker/plugins/project-deletion.jar &&
```
imsdu, line 21: The project deletion plugin is bundled in a different path to avoid enabling it by mistake so this extra step is needed to enable it

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/Projects.scala
```diff
@@ -263,7 +261,7 @@ object Projects {
           IO.raiseError(ProjectIsMarkedForDeletion(c.ref))
         case Some(s)                        =>
           // format: off
-          referenceFinder.raiseIfAny(c.ref) >>
+          validateDeletion.apply(c.ref) >>
```
shinyhappydan, line 264: Maybe scala is forcing you, but .apply?

shinyhappydan (dismissed):

shinyhappydan:
The only thing to be careful of is the change of behaviour on the delete endpoint. If that's intended then all good

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala
imsdu: The behaviour changes yes, as you will now get an error if project deletion is disabled.
But it is cleaner that way in my opinion as delete and deprecate are not the same thing

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/Projects.scala
imsdu: IntelliJ is not always very good at renaming things

PR #3879, opened 2023-05-12
Add tag latest when fetching dependencies

### LINKED ISSUE #3849: The recursive query for aggregate views is slow
**Describe the bug**
The query related to aggregate views tends to be slow in production

```
2023-04-27 07:13:01 WARN  k.i.j.JdbcInstrumentation$LoggingProcessors$WarnOnSlowStatement - Query execution exceeded the [2000000000 nanoseconds] threshold and lasted [2050037418 nanoseconds]. The query was: [
 
 WITH RECURSIVE recursive_dependencies(org, project, id) AS (
 SELECT target_org, target_project, target_id
 FROM entity_dependencies
 WHERE org = ? and project = ? and id = ?
 UNION
 SELECT e.target_org, e.target_project, e.target_id
 FROM entity_dependencies e, recursive_dependencies r
 WHERE e.org = r.org
 AND e.project = r.project
 AND e.id = r.id
 ) CYCLE id SET is_cycle USING path 
 SELECT s.value
 FROM recursive_dependencies d, scoped_states s
 WHERE s.org = d.org
 AND s.project = d.project
 AND s.id = d.id
       ]
```

**Expected behavior**
The query should be reworked to be faster

imsdu:
Fixes #3849

### DIFF
--- a/delta/sourcing-psql/src/test/scala/ch/epfl/bluebrain/nexus/delta/sourcing/EntityDependencyStoreSuite.scala
+++ b/delta/sourcing-psql/src/test/scala/ch/epfl/bluebrain/nexus/delta/sourcing/EntityDependencyStoreSuite.scala
@@ -7,8 +7,8 @@ import ch.epfl.bluebrain.nexus.delta.sourcing.PullRequest.PullRequestState.PullR
 import ch.epfl.bluebrain.nexus.delta.sourcing.config.QueryConfig
 import ch.epfl.bluebrain.nexus.delta.sourcing.implicits._
 import ch.epfl.bluebrain.nexus.delta.sourcing.model.EntityDependency.{DependsOn, ReferencedBy}
-import ch.epfl.bluebrain.nexus.delta.sourcing.model.Identity.Anonymous
-import ch.epfl.bluebrain.nexus.delta.sourcing.model.ProjectRef
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.Identity.{Anonymous, User}
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.{Label, ProjectRef}
 import ch.epfl.bluebrain.nexus.delta.sourcing.model.Tag.UserTag
 import ch.epfl.bluebrain.nexus.delta.sourcing.query.RefreshStrategy
 import ch.epfl.bluebrain.nexus.delta.sourcing.state.ScopedStateStore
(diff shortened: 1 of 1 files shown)


shinyhappydan (approved):
It feels like this could have a test

delta/sourcing-psql/src/test/scala/ch/epfl/bluebrain/nexus/delta/sourcing/EntityDependencyStoreSuite.scala
```diff
@@ -58,7 +59,9 @@ class EntityDependencyStoreSuite extends BioSuite with Doobie.Fixture {
   private val dependencyId5 = DependsOn(proj2, id5)
 
   test("Save the different states") {
-    List(state1, state2, state3, state5).traverse(stateStore.unsafeSave(_)).transact(xas.write)
+    val saveLatestStates = List(state1, state2, state3, state5).traverse(stateStore.unsafeSave(_))
+    val saveTaggedStates = List(state2, state3).traverse(stateStore.unsafeSave(_, UserTag.unsafe("my-tag")))
+    (saveLatestStates >> saveTaggedStates).transact(xas.write)
```
shinyhappydan, line 64: I can't really tell what this test is doing, is this affecting the other tests somehow?

imsdu: It is populating the database with tagged states that should never been returned in this context

shinyhappydan: The problem I see is that it's unclear from reading the test how those tagged states would affect anything. It would be better if there was a test that was clearly saying that the tagged versions should not appear (or whatever the assertion should be)

shinyhappydan (approved):

delta/sourcing-psql/src/test/scala/ch/epfl/bluebrain/nexus/delta/sourcing/EntityDependencyStoreSuite.scala
```diff
@@ -109,7 +113,7 @@ class EntityDependencyStoreSuite extends BioSuite with Doobie.Fixture {
       )
   }
 
-  test("Fetch values for direct dependencies of id1") {
+  test("Fetch latest state values for direct dependencies of id1") {
```
shinyhappydan, line 116: This doesn't check that the latest state is returned, as both the states appear the same in the result (so you could be returning the tagged version rather than latest)

To communicate the intent of the test better, it would be nice to express "I want the latest version of state 2, and I don't want the other version of state 2"

olivergrabinski (approved):

PR #3914, opened 2023-05-31
Fix composite views clean up

### LINKED ISSUE #3913: On a database outage, composite views start reindexing from the start
**Describe the bug**
When there is a database outage, the coordinator responsible for composite views gets restarted and start reindexing from the beginning.

imsdu:
Fixes #3913

### DIFF
--- a/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
+++ b/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
@@ -96,17 +96,17 @@ object CompositeViewsCoordinator {
       (cachedOpt, viewDef) match {
         case (Some(cached), active: ActiveViewDef) if cached.projection == active.projection =>
           Task.delay(logger.info(s"Projection '${cached.projection}' is already running and will not be recreated."))
-        case (Some(cached), _: ActiveViewDef)                                                              =>
+        case (Some(cached), _: ActiveViewDef)                                                =>
           Task.delay(
             logger.info(s"View '${ref.project}/${ref.viewId}' has been updated, cleaning up the current one.")
           ) >>
             destroy(cached)
-        case (Some(cached), _: DeprecatedViewDef)                                                          =>
+        case (Some(cached), _: DeprecatedViewDef)                                            =>
           Task.delay(
             logger.info(s"View '${ref.project}/${ref.viewId}' has been deprecated, cleaning up the current one.")
           ) >>
             destroy(cached)
-        case (None, _)                                                                                     =>
+        case (None, _)                                                                       =>
           Task.delay(
             logger.debug(s"View '${ref.project}/${ref.viewId}' is not referenced yet, cleaning is aborted.")
           )
--- a/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
+++ b/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
@@ -4,7 +4,7 @@ import cats.syntax.all._
 import ch.epfl.bluebrain.nexus.delta.kernel.cache.KeyValueStore
 import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.CompositeViews
 import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.indexing.CompositeViewDef.{ActiveViewDef, DeprecatedViewDef}
-import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.indexing.CompositeViewsCoordinator.logger
+import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.indexing.CompositeViewsCoordinator.cleanupCurrent
 import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.projections.CompositeProjections
 import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.stream.CompositeGraphStream
 import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution
@@ -52,7 +52,7 @@ final class CompositeViewsCoordinator(
             // Compiling and validating the view
             projection <- CompositeViewDef.compile(active, spaces, compilePipeChain, graphStream, compositeProjections)
             // Stopping the previous version
-            _          <- cleanupCurrent(active.ref)
+            _          <- cleanupCurrent(cache, active, destroy)
             // Init, register and run the new version
             _          <- supervisor.run(
                             projection,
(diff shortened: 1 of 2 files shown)


delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinatorSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+                        cachedViews.traverse { v => c.put(v.ref, v) }
+                      }
+      // The destroy action
+      destroyed    <- Ref.of[Task, Option[ActiveViewDef]](None)
+      destroy       = (active: ActiveViewDef) => destroyed.set(Some(active))
+      _            <- CompositeViewsCoordinator.cleanupCurrent(cache, newView, destroy)
+      destroyedOpt <- destroyed.get
+    } yield destroyedOpt
+
+  }
```
shinyhappydan, line 42: A couple of things
-  Is cache a good name here? My understanding is that this is the set of running views
- If destroy is called multiple times (which could happen if there is a bug) this would not be reflected by this fixture

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinatorSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+    val existingView = activeView(existingViewRef, 1)
+    val deprecated   = deprecatedView(existingViewRef)
+    cleanup(deprecated, List(existingView)).assertSome(existingView)
+  }
+
+  test("Clean up if the view is updated") {
+    val existingView = activeView(existingViewRef, 1)
+    val updatedView  = activeView(existingViewRef, 2)
+    cleanup(updatedView, List(existingView)).assertSome(existingView)
+  }
```
shinyhappydan, line 71: These are a bit hard to read

It's going to be clearer if there is more of a given/when/then structure to tests

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
```diff
@@ -103,6 +86,34 @@ object CompositeViewsCoordinator {
...
+      destroy: ActiveViewDef => Task[Unit]
+  ): Task[Unit] = {
+    val ref = viewDef.ref
+    cache.get(ref).flatMap { cachedOpt =>
+      (cachedOpt, viewDef) match {
+        case (Some(cached), active: ActiveViewDef) if cached.projection == active.projection =>
+          Task.delay(logger.info(s"Projection '${cached.projection}' is already running and will not be recreated."))
+        case (Some(cached), _: ActiveViewDef)                                                              =>
+          Task.delay(
+            logger.info(s"View '${ref.project}/${ref.viewId}' has been updated, cleaning up the current one.")
```
shinyhappydan, line 101: We really need a logger which returns tasks because the sclafmt here is painful

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinatorSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+import ch.epfl.bluebrain.nexus.testkit.bio.BioSuite
+import monix.bio.Task
+
+import java.util.UUID
+
+class CompositeViewsCoordinatorSuite extends BioSuite with CompositeViewsFixture {
+
+  private val existingViewRef                    = ViewRef(projectRef, nxv + "my-view")
+  private val anotherView                        = ViewRef(projectRef, nxv + "another-view")
+  private def activeView(ref: ViewRef, rev: Int) = ActiveViewDef(
```
shinyhappydan, line 19: `activeViewWithRevision`?

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
olivergrabinski: log4cats

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinatorSuite.scala
imsdu: * It is the same name as in the implementation
* destroy like almost all operations to start/stop indexing is meant to be idempotent

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
```diff
@@ -103,6 +86,34 @@ object CompositeViewsCoordinator {
...
+          ) >>
+            destroy(cached)
+        case (Some(cached), _: DeprecatedViewDef)                                                          =>
+          Task.delay(
+            logger.info(s"View '${ref.project}/${ref.viewId}' has been deprecated, cleaning up the current one.")
+          ) >>
+            destroy(cached)
+        case (None, _)                                                                                     =>
```
olivergrabinski, line 109: Probably out-of-scope but we are missing this now in the ES and BG coordinators

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
```diff
@@ -62,39 +62,22 @@ final class CompositeViewsCoordinator(
...
-              _ <- cache.remove(v.ref)
-            } yield ()
-          )
-          .void
-      case None    =>
-        Task.delay(
-          logger.debug(s"View '${ref.project}/${ref.viewId}' is not referenced yet, cleaning is aborted.")
-        )
-    }
+  private def destroy(active: ActiveViewDef) =
```
olivergrabinski, line 70: If we don't have any abstraction between coordinators, it would be nice to ultimately have the destroy function in ES and BG coordinators as well. The bug was easy to spot when comparing the coordinators as they had the same structure.

olivergrabinski (approved):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
imsdu: We have too look into log4cats or some similar library for this.
But in the end, yes

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinatorSuite.scala
imsdu: An active view always include a revision

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
imsdu: Ok, we can have a look at that in another PR

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
imsdu: We can see if it is worth it even if if an abstraction for 3 cases may be overkill and might be not bring a real gain in the end (the previous implementation in 1.7 had one and it was not a real win).
We must also do this refactor for the construct query before looking for it

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinatorSuite.scala
shinyhappydan: Ok but that doesn't mean these tests are easy to read. Comments should really be a last resort when we can't communicate intent through the code

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinatorSuite.scala
imsdu: * I updated the test so that the value can only be set once
* `KeyValueStore` has now one implementation which is a in-memory cache
* This PR aims to fix an issue in production so it is better to keep the change minimal

PR #3918, opened 2023-05-31
Retry on the destroy method in supervisor in case of failure

### LINKED ISSUE #3894: Investigate why some projections are not properly cleaned up
When we deploy a new version of a composite views, the former version of its progress, its namespace and its index are not properly destroyed.

We should make sure that:
* An error log is displayed whenever this situation happens
* We should retry when this fails (whenever we update the search config, a lot of pressure is put in Blazegraph which is the main suspect here on why the deletion is not properly executed)

imsdu:
Fixes #3894


delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/ExecutionStrategy.scala
```diff
@@ -21,7 +21,7 @@ sealed trait ExecutionStrategy extends Product with Serializable {
     *   the cluster configuration
     */
   def shouldRun(name: String, cluster: ClusterConfig): Boolean = this match {
-    case _: SingleNode => name.hashCode % cluster.size == cluster.nodeIndex
+    case _: SingleNode => Math.abs(name.hashCode) % cluster.size == cluster.nodeIndex
```
imsdu, line 24: Hashcode can be negative

delta/sourcing-psql/src/test/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/SupervisionSuite.scala
```diff
@@ -353,4 +348,60 @@ class SupervisionSuite extends BioSuite with SupervisorSetup.Fixture with Doobie
...
+  }
+
+  test("Run and properly destroy a projection with an failing destroy method") {
+    val projection = ProjectionMetadata("test", "unstable-global-projection", None, None)
+    for {
+      // Starting the projection
+      started <- Ref.of[Task, Boolean](false)
+      compiled = CompiledProjection.fromStream(projection, ExecutionStrategy.EveryNode, evalStream(started.set(true)))
+      _       <- sv.run(compiled).eventually(ExecutionStatus.Running)
+      _       <- started.get.eventually(true)
```
shinyhappydan, line 375: Would be nice to hide this boilerplate

delta/sourcing-psql/src/test/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/SupervisionSuite.scala
```diff
@@ -353,4 +348,60 @@ class SupervisionSuite extends BioSuite with SupervisorSetup.Fixture with Doobie
...
+      // Destroy the projection with a destroy method that will always fail
+      failing  = Task.raiseError(new IllegalStateException("Fail !"))
```
shinyhappydan, line 377: if you name this variable "alwaysFails" then there is no need for this comment

delta/sourcing-psql/src/test/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/SupervisionSuite.scala
```diff
@@ -353,4 +348,60 @@ class SupervisionSuite extends BioSuite with SupervisorSetup.Fixture with Doobie
...
+      _       <- sv.destroy(projection.name, failing)
+      // The projection should have stopped
+      _       <- sv.describe(projection.name).eventuallyNone
```
shinyhappydan, line 380: If you make a method assertStopped(projection) then it's more readable and you don't need this comment

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/Supervisor.scala
```diff
@@ -356,7 +360,13 @@ object Supervisor {
                               _      <- Task.delay(log.info(s"Destroying '${metadata.module}/${metadata.name}'..."))
                               _      <- stopProjection(s)
                               _      <- Task.when(s.executionStrategy == PersistentSingleNode)(projections.delete(name))
-                              _      <- clear
+                              _      <- onDestroy
+                                          .retryingOnSomeErrors(
+                                            retryStrategy.retryWhen,
+                                            retryStrategy.policy,
+                                            retryStrategy.onError
+                                          )
```
shinyhappydan, line 368: Can we make it so that this method just takes a retry strategy? Would be much more readable

shinyhappydan (dismissed):

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/Supervisor.scala
imsdu: I created an issue to introduce this method and replace it everywhere #3928

delta/sourcing-psql/src/test/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/SupervisorSuite.scala
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  test("Run and properly destroy a projection with an failing destroy method") {
+    val projection = ProjectionMetadata("test", "unstable-global-projection", None, None)
+    for {
+      _         <- assertStart(projection, EveryNode)
+      alwaysFail = Task.raiseError(new IllegalStateException("Fail !"))
+      _         <- assertDestroy(projection, alwaysFail)
+    } yield ()
+  }
```
shinyhappydan, line 282: This is pretty nice, but a small change like this can help make it more like given/when/then

```
  test("Run and properly destroy a projection with an unstable destroy method") {
    val projection = ProjectionMetadata("test", "unstable-global-projection", None, None)
    val destroyThatAlwaysFails = Task.raiseError(new IllegalStateException("Fail !"))

    for {
      _               <- startProjection(projection, EveryNode)
      _               <- destroyProjection(projection, destroyThatFailsThenSucceeds)
      _               <- assertProjectionIsDestroyed(projection)
    } yield ()
  }
```

I realise that technically we are asserting during the startup, but it's not necessary to communicate this, only that the start is what is happening

shinyhappydan (dismissed):

olivergrabinski (approved):

PR #3929, opened 2023-06-06
Do not retry when retrieving groups fail

### LINKED ISSUE #3864: Do not retry when retrieving groups from the OpenID provider fails
Failing fast and returning an error to the user immediately would be a better option in this case.

imsdu:
Fixes #3864


shinyhappydan (approved):
I think as long as there is never a situation where retrying makes sense, then this PR is good

PR #3957, opened 2023-06-14
Add search config integration tests

olivergrabinski:
Notes:
* Some tests are ignored and are to be completed 
* Some resources used for the tests have been altered in order to cover more test with one resource (e.g. adding a `weight` field to a subject).


tests/src/test/resources/application.conf
```diff
@@ -5,6 +5,17 @@ tests {
   clean-up = false
 }
 
+akka {
+  http {
+    sse {
+      # The maximum size for parsing server-sent events (96KiB).
+      max-event-size = 98304
+      # The maximum size for parsing lines of a server-sent event (48KiB).
+      max-line-size = 49152
```
olivergrabinski, line 14: Due to the change in the search config, the default search view resource gets bigger, and could not be streamed as sse without this config change.

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigSpec.scala
```diff
@@ -0,0 +1,596 @@
...
+
+class SearchConfigSpec extends BaseSpec {
+
+  private val orgId    = genId()
+  private val projId1  = genId()
+  private val id1      = s"$orgId/$projId1"
+  private val projects = List(id1)
+
+  private val neuronMorphologyId =
+    "https://bbp.epfl.ch/neurosciencegraph/data/neuronmorphologies/da3b1e42-5f7f-4065-8be3-d2132c219bc2"
```
imsdu, line 20: We can use explicit ids instead of ones with uuids
And make them a little shorter

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigSpec.scala
```diff
@@ -0,0 +1,596 @@
...
+      eventually {
+        deltaClient.post[Json]("/search/query", json"""{"size": 100}""", Rick) { (body, response) =>
+          response.status shouldEqual StatusCodes.OK
+          val sources = Json.fromValues(body.findAllByKey("_source"))
+          sources.asArray.get.size shouldBe mainResources.size
+        }
+      }
+    }
+
+    "index project" in {
```
imsdu, line 81: This is not really about indexing there but asserting the property is correctly indexed

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigSpec.scala
```diff
@@ -0,0 +1,596 @@
...
+import akka.http.scaladsl.model.StatusCodes
+import cats.implicits._
+import ch.epfl.bluebrain.nexus.tests.BaseSpec
+import ch.epfl.bluebrain.nexus.tests.Identity.resources.Rick
+import ch.epfl.bluebrain.nexus.tests.iam.types.Permission.{Organizations, Resources}
+import io.circe.Json
+import monix.bio.Task
+import org.scalatest.Assertion
+
+class SearchConfigSpec extends BaseSpec {
```
imsdu, line 12: It looks ok for now but we may want to break it further down  at some point.
In other tests, we may also want to run some aggregations too (again, not in this PR)

imsdu (changes requested):

imsdu (approved):

PR #3965, opened 2023-06-15
Fix docker-compose file used for integration tests

olivergrabinski:
Related to #3746
Allows to run the `RemoteStorageSpec` without errors


shinyhappydan (approved):
🥳

PR #3982, opened 2023-06-16
Introduce aggregations on default views

imsdu:
This PR include a start about introducing aggregations on default views in the listing operations.

A lot of things are WIP especially on the naming.

One of the purposes here is also to break `ElasticSearchViewsQuery` that does a lot of things:
* Listing operations on default views
* "Free search query" on all views
* Expand the schema reference to inject it in the params

Another one is to make things more modular by  having a store dedicated for default views, to avoid having a default dependency to ES when we can avoid it, and to expand the schema reference before submitting a search.

There is also a change of behaviour when performing a change at the org and root level: if a user has not access to a single view, a authorization error will be raised (instead of returning an empty result). 
The idea is to make it more consistent with the project level where it is already the case

TODO:
* Create tests for the aggregations themselves
* Replace `ElasticsearchViewsQuery` operations related to listing by `DefaultViewsQuery` in `ElasticsearchRoutes`
* Handle the expansion of the resource ref of the schema in the routes before passing it on to `DefaultViewsQuery`
* Delete the list methods of `ElasticSearchViewsQuery`
* In the Elasticsearch routes, remove the check when perfoming a search at the project level (it is handled by `DefaultViewsQuery` so that the ACL logic is now performed in one single place instead of 2)
* One option to explore may be to split `ElasticsearchRoutes` in 2:  one for the query operations, the other for the rest

### DIFF
--- a/delta/plugins/elasticsearch/src/main/resources/elasticsearch.conf
+++ b/delta/plugins/elasticsearch/src/main/resources/elasticsearch.conf
@@ -45,4 +45,7 @@ plugins.elasticsearch {
   }
   # the query config used when fetching all scoped events to obtain metrics
   metrics-query = ${app.defaults.query}
+  # when performing aggregations for default views, defines the maximum number
+  # of buckets for the aggregation
+  listing-bucket-size = 500
 }
--- a/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
+++ b/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
@@ -21,7 +21,7 @@ import ch.epfl.bluebrain.nexus.delta.sdk.model.ComponentDescription.ServiceDescr
 import ch.epfl.bluebrain.nexus.delta.sdk.model.ComponentDescription.ServiceDescription.ResolvedServiceDescription
 import ch.epfl.bluebrain.nexus.delta.sdk.model.search.ResultEntry.{ScoredResultEntry, UnscoredResultEntry}
 import ch.epfl.bluebrain.nexus.delta.sdk.model.search.SearchResults.{ScoredSearchResults, UnscoredSearchResults}
-import ch.epfl.bluebrain.nexus.delta.sdk.model.search.{Pagination, ResultEntry, SearchResults, SortList}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.search.{AggregationResult, Pagination, ResultEntry, SearchResults, SortList}
 import ch.epfl.bluebrain.nexus.delta.sdk.model.{BaseUri, Name}
 import ch.epfl.bluebrain.nexus.delta.sdk.syntax._
 import com.typesafe.scalalogging.Logger
@@ -491,6 +491,29 @@ class ElasticSearchClient(client: HttpClient, endpoint: Uri, maxIndexPathLength:
   }
 
   /**
+    * Perform an aggregation for the ''query'' inside the given indices
+    *
+    * @param params
+    *   the filter parameters
+    * @param indices
+    *   the indices to use (if empty, searches in all the indices)
+    * @param qp
+    *   the query parameters
+    * @param bucketSize
+    *   the maximum number of terms returned by a term aggregation
+    * @see
+    *   https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html#search-aggregations-bucket-terms-aggregation-size
+    */
+  def aggregate(params: ResourcesSearchParams, indices: Set[String], qp: Query, bucketSize: Int)(implicit
+      base: BaseUri
+  ): HttpResult[AggregationResult] = {
+    val query          = QueryBuilder(params).aggregation(bucketSize)
+    val (indexPath, q) = indexPathAndQuery(indices, query)
+    val searchEndpoint = (endpoint / indexPath / searchPath).withQuery(Uri.Query(defaultQuery ++ qp.toMap))
+    client.fromJsonTo[AggregationResult](Post(searchEndpoint, q.build).withHttpCredentials)
+  }
+
+  /**
     * Refresh the given index
     */
   def refresh(index: IndexLabel): HttpResult[Boolean] =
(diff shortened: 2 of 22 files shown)


delta/plugins/elasticsearch/src/main/resources/elasticsearch.conf
```diff
@@ -45,4 +45,7 @@ plugins.elasticsearch {
   }
   # the query config used when fetching all scoped events to obtain metrics
   metrics-query = ${app.defaults.query}
+  # when performing aggregations for default views, defines the maximum number
+  # of buckets for the aggregation
+  listing-bucket-size = 500
```
imsdu, line 50: By default on aggregations, ES returns the top 10 of values.
This allows to configure it.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/ElasticSearchClient.scala
```diff
@@ -490,6 +490,29 @@ class ElasticSearchClient(client: HttpClient, endpoint: Uri, maxIndexPathLength:
...
+    * @param indices
+    *   the indices to use (if empty, searches in all the indices)
+    * @param qp
+    *   the query parameters
+    * @param bucketSize
+    *   the maximum number of terms returned by a term aggregation
+    * @see
+    *   https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html#search-aggregations-bucket-terms-aggregation-size
+    */
+  def aggregate(params: ResourcesSearchParams, indices: Set[String], qp: Query, bucketSize: Int)(implicit
```
imsdu, line 507: New method to perform aggregations with the same params as the `search` method

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/query/DefaultSearchRequest.scala
```diff
@@ -0,0 +1,119 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.projects.FetchContext
+import ch.epfl.bluebrain.nexus.delta.sdk.projects.model.{ApiMappings, ProjectBase}
+import ch.epfl.bluebrain.nexus.delta.sourcing.Predicate
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.{Label, ProjectRef, ResourceRef}
+import monix.bio.IO
+
+/**
+  * Search request on default elasticsearch views
+  */
+sealed trait DefaultSearchRequest extends Product with Serializable {
```
imsdu, line 17: This allows to avoid to have a method for search and aggregate for each project/org/level

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/query/DefaultSearchRequest.scala
```diff
@@ -0,0 +1,119 @@
...
+  /**
+    * Search to be performed on a project
+    */
+  case class ProjectSearch(ref: ProjectRef, params: ResourcesSearchParams, pagination: Pagination, sort: SortList)
+      extends DefaultSearchRequest {
+    override def predicate: Predicate = Predicate.Project(ref)
+  }
+
+  object ProjectSearch {
+    def apply(
```
imsdu, line 52: Apply method to use when the schema is passed as a segment

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/query/DefaultSearchRequest.scala
```diff
@@ -0,0 +1,119 @@
...
+  /**
+    * Search to be performed on an org
+    */
+  case class OrgSearch(label: Label, params: ResourcesSearchParams, pagination: Pagination, sort: SortList)
+      extends DefaultSearchRequest {
+    override def predicate: Predicate = Predicate.Org(label)
+  }
+
+  object OrgSearch {
+    def apply(label: Label, params: ResourcesSearchParams, pagination: Pagination, sort: SortList, schema: IdSegment)(
```
imsdu, line 78: Same but for org

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/query/DefaultSearchRequest.scala
```diff
@@ -0,0 +1,119 @@
...
+  /**
+    * Search to be performed on all default views
+    */
+  case class RootSearch(params: ResourcesSearchParams, pagination: Pagination, sort: SortList)
+      extends DefaultSearchRequest {
+    override def predicate: Predicate = Predicate.Root
+  }
+
+  object RootSearch {
+    def apply(params: ResourcesSearchParams, pagination: Pagination, sort: SortList, schema: IdSegment)(
```
imsdu, line 95: Same but for root

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/query/DefaultViewsQuery.scala
```diff
@@ -0,0 +1,99 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.model.search.{AggregationResult, SearchResults}
+import ch.epfl.bluebrain.nexus.delta.sdk.views.View.IndexingView
+import ch.epfl.bluebrain.nexus.delta.sourcing.Predicate
+import io.circe.JsonObject
+import monix.bio.{IO, UIO}
+
+/**
+  * Allow to list resources from the default elasticsearch views
+  */
+trait DefaultViewsQuery[Result, Aggregate] {
```
imsdu, line 22: I made the search and aggregate result generic because returning json in tests makes it difficult to correctly assert things and because the main purpose of this class is in the end to properly filter the views the user has access to

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/query/ElasticSearchQueryError.scala
```diff
@@ -0,0 +1,54 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.error.ServiceError
+import ch.epfl.bluebrain.nexus.delta.sdk.http.HttpClientError
+import ch.epfl.bluebrain.nexus.delta.sdk.marshalling.HttpResponseFields
+import io.circe.syntax.KeyOps
+import io.circe.{Encoder, JsonObject}
+
+/**
+  * Enumeration of errors raised while querying the Elasticsearch indices
+  */
+sealed abstract class ElasticSearchQueryError(val reason: String) extends Product with Serializable
```
imsdu, line 15: The idea is to separate the rejections raised at query time from the other elasticsearch rejections

imsdu:
Closed in favour of #4004

PR #4004, opened 2023-06-26
Add aggreations on default views

olivergrabinski:
Introduces aggregations:

On `/v1/resources` it is now possible to use the `aggregations` Boolean parameter to request `@type` and `project` bucket aggregations on the underlying resources (at the same scope that listings).

[Aggregations doc changes](https://github.com/BlueBrain/nexus/assets/20188437/1e8948cb-238f-428b-9a69-2af3641ff10d)


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsRoutes.scala
```diff
@@ -277,108 +283,132 @@ final class ElasticSearchViewsRoutes(
...
+                  }
+                )
+              }
+            }
           }
         )
       }
     }
 
+  private val aggregationResourceRoutes: Route =
```
imsdu, line 327: What about putting the aggregations under:
`/aggregations/{org}/{project}`
It feels a bit weird with a get parameter, it feels more than a whole new operation

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsRoutes.scala
```diff
@@ -277,108 +283,132 @@ final class ElasticSearchViewsRoutes(
...
+    pathPrefix("resources") {
+      extractCaller { implicit caller =>
+        (searchParametersAndSortList(baseUri) & paginated & aggregated) { (params, sort, page, _) =>
```
imsdu, line 330: We don't need pagination
The `resourceSchema` is missing

imsdu (changes requested):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsRoutes.scala
olivergrabinski: It’s a possibility but then it’s not possible to merge the current generic resource listing and aggregation routes anymore 🤔

olivergrabinski: Reason I kept it under resources is that we are aggregating the same resources that we list with /resources/{org}/{project}.

Also briefly considered /resources/{org}/{project}/aggregate but I don’t think it’s possible without forbidding to use “aggregate” as org/project name.

imsdu (approved):

PR #4018, opened 2023-07-05
Improve documentation on Elasticsearch pipelines

imsdu:
image
image


docs/src/main/paradox/docs/delta/api/views/elasticsearch-view-api.md
```diff
@@ -11,15 +11,43 @@ A default view gets automatically created when the project is created but other
...
+* The original payload of the resource which has been submitted by the user
+* The metadata graph of the resource (the fields that have been generated from Nexus)
+* The data graph of the resource (that has been computed from the original payload at ingestion time)
+
+### Reconciliation and indexing
+
+When a resource makes it to the end of the pipeline, we produce a json document by merging:
+
+* the original payload,
+* the data and metadata graphs that have been translated to the compacted JSON-LD format with a default context and the one provided by the view
```
olivergrabinski, line 48: Maybe adding a link to the default context?

imsdu: The thing is that there are two of them that we merge at runtime: 
https://github.com/BlueBrain/nexus/blob/c85a0bc056d6e6833c8df93ef9020280fa7df96f/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/IndexingViewDef.scala#L37

olivergrabinski: Ok, let's see then if there are any questions about it later

olivergrabinski (approved):

PR #4025, opened 2023-07-05
Apply fixes to file self in archives

### LINKED ISSUE #4016: FileAddress in archives
**Describe the bug**
https://github.com/BlueBrain/nexus/issues/3924 must be improved on a few points:

* `FileLink` should be renamed to `FileSelf` or `FileAddress` as this is the vocabulary we already use in Nexus
* Use an Uri in the `FileSelf`
* The current implementation does not handle the case when the `_self` include a rev / a tag
* The error message is not precise enough as it won't give information about the invalid `_self` while it is possible to provide a list of them

imsdu:
Fixes #4016


delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/ArchiveDownload.scala
```diff
@@ -87,7 +87,7 @@ object ArchiveDownload {
       aclCheck: AclCheck,
       fetchResource: (ResourceRef, ProjectRef) => UIO[Option[JsonLdContent[_, _]]],
       fetchFileContent: (ResourceRef, ProjectRef, Caller) => IO[FileRejection, FileResponse],
-      resolveSelf: ResolveFileSelf
+      fileSelf: FileSelf
```
imsdu, line 90: I removed the resolve part as in Nexus, it also means actually find the thing

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/ArchiveReferenceType.scala
```diff
@@ -18,7 +18,7 @@ object ArchiveReferenceType {
   final case object File extends ArchiveReferenceType
 
   /**
-    * An archive file reference, but a link rather than a full reference
+    * An archive file reference, but a self of a file rather than a full reference
     */
-  final case object FileLink extends ArchiveReferenceType
+  final case object FileSelf extends ArchiveReferenceType
```
olivergrabinski, line 23: I don't get where `ArchiveReferenceType` is used. If it is used then I think something goes untested 🤔

olivergrabinski (approved):

PR #4030, opened 2023-07-06
Improve retry strategy syntax

### LINKED ISSUE #3928: Improve retry strategy syntax
With applying a retry startegy, we currently rely on the `cats.retry` syntax:
```scala
.retryingOnSomeErrors(
  retryStrategy.retryWhen,
  retryStrategy.policy,
  retryStrategy.onError
)
```
It would be nice to improve it in nexus to make it shorter:
```scala
.retryingOnSomeErrors(retryStrategy)
```

Introduce the new syntax and replace the other method with it

imsdu:
Fixes #3928


delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/RetryStrategy.scala
```diff
@@ -147,22 +158,41 @@ object RetryStrategyConfig {
...
   }
 
+  /**
+    * Retry with a constant delay until the total delay reaches the limit
+    * @param threshold
+    *   the maximum cumulative delay
+    * @param delay
+    *   the delay between each try
+    */
+  final case class MaximumCumulativeDelayConfig(threshold: FiniteDuration, delay: FiniteDuration)
```
imsdu, line 168: As it can also be useful outside of tests

olivergrabinski (approved):

PR #4031, opened 2023-07-06
Add possibility to disable indexing via configuration

### LINKED ISSUE #3942: Allow the possibility to disable indexing by configuration
**Motivation**
It proved useful to be able to disable BG/ES/Composite views / graph analytics indexing to manage the pressure on the different components during the migration.
It would be nice to keep this possibility for maintenance operations

**Acceptance criteria**
1. Prefer HOCON configuration to environment variables

olivergrabinski:
This adds the possibilities to:

* Disable all indexing (ES, BG, CV, GA) by setting `app.defaults.indexing.enable` to `false` (defaults to `true`).
* Disable indexing for only one plugin by setting one (or more) of the following to `false` (defaults to the value of `app.defaults.indexing.disable`)
    * `plugins.elasticsearch.indexing-enabled`
    * `plugins.blazegraph.indexing-enabled`
    * `plugins.composite-views.indexing-enabled`
    * `plugins.graph-analytics.indexing-enabled`

Closes #3942


delta/app/src/main/resources/app.conf
```diff
@@ -360,6 +360,8 @@ app {
     indexing {
       # default prefix to use when creating indices
       prefix = "delta"
+      # turn off all indexing
+      disable = false
```
imsdu, line 364: If we use `enabled` here, we can get rid of the negation

imsdu (changes requested):

delta/app/src/main/resources/app.conf
olivergrabinski: The idea was to avoid confusion and for the user to `enable` the plugin, but `disable` the indexing. But using `enable` for indexing will indeed remove the need for the method there, so we can do that

imsdu (approved):

PR #4060, opened 2023-07-11
Introduce log4cats for logging

olivergrabinski:
* Add log4cats 1.7.0 as a dependency to the kernel
* Add a `Logger` class to the kernel. This is wrapping the log4cats logger in order to be able to have a `Logger[UIO]`.
* Use the new `Logger` in the coordinators


shinyhappydan (dismissed):
I still hate the double parameter list but sure

imsdu (approved):

PR #4061, opened 2023-07-12
Add dedicated clients for composite views

olivergrabinski:
* Adds the possibility of pointing to a different instance for composite views than for blazegraph views.
* By default the same instance is used for both

### DIFF
--- a/delta/plugins/composite-views/src/main/resources/composite-views.conf
+++ b/delta/plugins/composite-views/src/main/resources/composite-views.conf
@@ -3,6 +3,13 @@ plugins.composite-views {
   enabled = true
   # the priority of the plugin
   priority = 5
+  # blazgraph instance for composite views
+  # by default it is the same as the one defined in the blazegraph plugin
+  blazegraph = {
+    base = ${plugins.blazegraph.base}
+    # credentials = ${plugins.blazegraph.credentials}
+    query-timeout = ${plugins.blazegraph.query-timeout}
+  }
   # the configuration of the composite views sources
   sources {
     max-sources = 5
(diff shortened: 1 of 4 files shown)


delta/plugins/composite-views/src/main/resources/composite-views.conf
```diff
@@ -3,6 +3,13 @@ plugins.composite-views {
   enabled = true
   # the priority of the plugin
   priority = 5
+  # blazgraph instance for composite views
+  # by default it is the same as the one defined in the blazegraph plugin
+  blazegraph = {
```
olivergrabinski, line 8: Another possibility is to group `base`, `credentials`, and `query-timeout` at blazegraph plugin level into a single object.

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/config/CompositeViewsConfig.scala
```diff
@@ -40,6 +45,7 @@ import scala.concurrent.duration.FiniteDuration
   */
 final case class CompositeViewsConfig(
     sources: SourcesConfig,
+    blazegraph: BlazegraphInstanceConfig,
```
imsdu, line 48: `access: BlazegraphAccess` ?
We do have `DatabaseAccess` already

imsdu (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/config/CompositeViewsConfig.scala
olivergrabinski: I used `blazegraphAccess` for the field name since `access` on it's own kind of begs the question "access to what?". Works like this?

imsdu (approved):

PR #4067, opened 2023-07-14
Update search config

olivergrabinski:
* Explicit mapping for values of `neuronDensity`, `layerThickness`, and `boutonDensity`. This is so that there are no indices where the value has a different numeric type across shards.
* Update to the latest neuroshapes context used in production
    * In order to frame `startedAt` and `endedAt` fields correctly, the context needs to specify the `@type`.
    * Due to this change, the expected values for these fields must be updated (they will have the same representation as `createdAt` and `updatedAt` in the search indices now).

### DIFF
--- a/tests/src/test/resources/kg/search/activity.json
+++ b/tests/src/test/resources/kg/search/activity.json
@@ -7,7 +7,7 @@
   ],
   "endedAtTime": {
     "@type": "xsd:dateTime",
-    "value": "2015-05-01T00:00:00"
+    "@value": "2015-05-01T00:00:00"
   },
   "generated": {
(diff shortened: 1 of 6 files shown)


tests/src/test/resources/kg/search/activity.json
```diff
@@ -27,7 +27,7 @@
   "reconstructionCompressionCorrected": "yes",
   "startedAtTime": {
     "@type": "xsd:dateTime",
-    "value": "2015-01-21T00:00:00"
```
olivergrabinski, line 30: `startedAtTime` and `endedAtTime` can no longer be `schema:value`s

imsdu (approved):

Issue #3948, opened 2023-06-12
Make the update of composite views more intelligent


imsdu:
**Motivation**
Whenever we update a composite view, we currently start from scratch and reindex everything from the beginning:
* Index in Blazegraph
* Index in Elasticsearch

However, 99% the changes only have an impact on the second part and not the first one while reindexing in Blazegraph makes things longer and generates fragmentation and excessive disk usage

**Acceptance criteria**
* Whenever we bring a change to the source, everything should be reindexed
* Whenever it is on a projection, only the projection part should be reindexed

PR #4073, opened 2023-07-18
Implement search updater as a hook

### LINKED ISSUE #3891: Make the search updater run before the composite views coordinator
To avoid starting the indexing process of composite views and restart it when a change is detected, we should run this update before starting their indexing process.

We have to pay attention to properly delete the existing progress/namespaces/indices.

One approach can be:
* The updater would not be a projection anymore but a trait created in the composite views plugin looking like
* 
``` 
trait UpdateBeforeRun {
 def apply(destroyIfUpdateFound): Task
}
```
* Create the implementation of  UpdateBeforeRun in the search plugin, make it available via distage
* In the composite views coordinator, inject the update runner, run the update with a `fs2.Stream.eval` before streaming the views

imsdu:
Fixes #3891

Coordinating two projections that may run on different nodes proved difficult and not really worth the effort so the idea is to run both actions as a hook:
* If the view and the config are aligned, we start to index
* If not, the update of the view is run as a hook and the indexing will start of the new revision that will come afterwards


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeProjectionLifeCycle.scala
```diff
@@ -0,0 +1,100 @@
...
+import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.stream.CompositeGraphStream
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution
+import ch.epfl.bluebrain.nexus.delta.sourcing.stream.ExecutionStrategy.TransientSingleNode
+import ch.epfl.bluebrain.nexus.delta.sourcing.stream.{CompiledProjection, PipeChain}
+import monix.bio.Task
+
+/**
+  * Handle the different life stages of a composite view projection
+  */
+trait CompositeProjectionLifeCycle {
```
imsdu, line 14: To separate the coordinator to the actions it triggers

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
```diff
@@ -36,42 +30,30 @@ object CompositeViewsCoordinator {
...
-    * @param compilePipeChain
-    *   how to compile pipe chains
-    * @param graphStream
-    *   to provide the data feeding the indexing
-    * @param buildSpaces
-    *   to create the necessary namespaces / indices and the pipes/sinks to interact with them
-    * @param compositeProjections
-    *   to fetch/save progress as well as handling restarts
+    * @param lifecycle
+    *   how to pouet
```
olivergrabinski, line 34: 😅

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeProjectionLifeCycleSuite.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  private def createHook(name: String, test: ViewRef => Boolean, ref: Ref[Task, Set[String]]): Hook =
+    (view: CompositeViewDef.ActiveViewDef) => {
+      Option.when(test(view.ref)) {
+        ref.update {
+          _ + name
+        }
+      }
+    }
+  private def assertView(view: ActiveViewDef, hooks: Set[String])(implicit loc: Location)           =
```
olivergrabinski, line 38: `assertViewHooks`?

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
```diff
@@ -36,42 +30,30 @@ object CompositeViewsCoordinator {
...
-      compilePipeChain: PipeChain.Compile,
-      graphStream: CompositeGraphStream,
-      buildSpaces: ActiveViewDef => CompositeSpaces,
-      compositeProjections: CompositeProjections
-  )(implicit cr: RemoteContextResolution)
-      extends CompositeViewsCoordinator {
+      lifecycle: CompositeProjectionLifeCycle
+  ) extends CompositeViewsCoordinator {
 
     def run(offset: Offset): Stream[Task, Elem[Unit]] = {
```
olivergrabinski, line 43: I like the result result on the `run` function, it's simpler now

olivergrabinski (changes requested):

olivergrabinski (approved):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewsCoordinator.scala
imsdu: I think it was needed yes as the coordinator knew too much.
And this change should also be helpful to implement #3948

PR #4075, opened 2023-07-19
Introduce a simple versioning system for database schema changes

imsdu:
As it is not clear for now if we need to invest time in setting up a tool such as Flyway, this is a first try to version our changes in the schema of the database without it.

* All database scripts are stored in the same directory
* They follow a naming convention allowing to run them in order


delta/sourcing-psql/src/main/resources/scripts/postgres/drop/drop-tables.ddl
olivergrabinski, line 1: It could now drop the jira tables as well, no?

olivergrabinski (approved):

olivergrabinski (approved):

PR #4077, opened 2023-07-19
Introduce `CompositeSink`

olivergrabinski:
Introduces the `CompositeSink`

* Previously the `QueryGraph`, the transform pipe, and the sink were separate pipes merged together.
* Now the `CompositeSink` is a single pipe that encompasses these operations.
* The goal is to make it easier to now introduce a second version of the `CompositeSink`


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeSink.scala
```diff
@@ -0,0 +1,54 @@
...
+    override val maxWindow: FiniteDuration
+) extends Sink {
+
+  override type In = GraphResource
+  override def inType: Typeable[GraphResource] = Typeable[GraphResource]
+
+  private def query: Elem[GraphResource] => Task[Elem[GraphResource]] =
+    elem => queryGraph(elem)
+
+  private def liftedTransform: Elem[GraphResource] => Task[Elem[SinkFormat]] =
```
imsdu, line 43: Has `query` and `transform` are two tasks, it should be possible to combine them this way no ?
`elem.evalMap(queryTransform)` so we would not have to do this

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/QueryGraph.scala
```diff
@@ -43,7 +34,7 @@ final class QueryGraph(client: BlazegraphClient, namespace: String, query: Sparq
         Some(g.replaceRootNode(id))
       }
 
-  override def apply(element: SuccessElem[GraphResource]): Task[Elem[GraphResource]] =
+  def apply(element: Elem[GraphResource]): Task[Elem[GraphResource]] =
```
imsdu, line 37: We can get rid of the elem here as it is not a pipe anymore

imsdu (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeSink.scala
olivergrabinski: I made some changes, hopefully it's simpler now.

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/QueryGraph.scala
olivergrabinski: Removed

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/indexing/GraphResourceToNTriples.scala
```diff
@@ -19,16 +19,17 @@ object GraphResourceToNTriples extends Pipe {
...
-        err => element.failed(err),
-        {
-          case ntriples if ntriples.isEmpty => element.dropped
-          case ntriples                     => element.success(ntriples)
-        }
-      )
+    element
+      .evalMap(graphToNTriples)
+      .map {
+        case ntriples: SuccessElem[NTriples] if ntriples.value.isEmpty => element.dropped
```
imsdu, line 31: This part should also be in `graphToNTriples` to have a consistent behaviour when using the pipe or not.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/GraphResourceToDocument.scala
```diff
@@ -31,22 +31,30 @@ final class GraphResourceToDocument(context: ContextValue, includeContext: Boole
...
         .map(json => injectId(json, id))
-        .map(json => if (json.isEmpty()) element.dropped else element.success(json))
     }
   }
 
+  override def apply(element: SuccessElem[GraphResource]): Task[Elem[Json]] =
+    element
+      .evalMap(graphToDocument)
+      .map {
+        case json: SuccessElem[Json] if json.value.isEmpty() => element.dropped
```
imsdu, line 54: Same as the other case, it needs to be included in the previous method

imsdu (changes requested):

imsdu (approved):

PR #4101, opened 2023-07-21
Add a dedicated worflow for the app module

### LINKED ISSUE #4068: Add a dedicated worflow for the app module
**Motivation**
Whenever only the app module gets updated, we also rebuild and test the core and plugins modules when it is not necessary.

**Acceptance criteria**
1. The workflows related to core and plugin modules are not triggered
2. A dedicated workflow is triggered

shinyhappydan:
Fixes #4068

### DIFF
--- a/.github/workflows/ci-delta-app.yml
+++ b/.github/workflows/ci-delta-app.yml
@@ -1,4 +1,4 @@
-name: Delta Core
+name: Delta App
 on:
   pull_request:
     paths:
--- a/.github/workflows/ci-delta-app.yml
+++ b/.github/workflows/ci-delta-app.yml
@@ -1,0 +1,49 @@
+name: Delta Core
+on:
+  pull_request:
+    paths:
+      - 'delta/**'
+      - 'build.sbt'
+      - 'project/**'
+      - '.github/workflows/ci-delta-app.yml'
+jobs:
+  static-analysis:
+    if: github.event_name == 'pull_request'
+    runs-on: it
+    timeout-minutes: 10
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Static Analysis
+        run: |
+          sbt -Dsbt.color=always -Dsbt.supershell=false \
+            "project app" \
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-delta-core.yml
```diff
@@ -23,11 +28,8 @@ jobs:
           sbt -Dsbt.color=always -Dsbt.supershell=false \
             "project delta" \
             clean \
-            scalafmtCheck \
-            Test/scalafmtCheck \
-            scalafmtSbtCheck \
-            scapegoat \
-            doc
+            core-static-analysis
```
olivergrabinski, line 31: The `core-static-analysis` should run also for plugins here. But it can't just be added to the `coreModules` as is now because that would also run the unit tests for `plugins` through `core-unit-tests-with-coverage`

olivergrabinski (changes requested):

.github/workflows/ci-delta-core.yml
shinyhappydan: Perhaps we can add it to the plugins build instead?

olivergrabinski: That should work yes

.github/workflows/ci-delta-app.yml
```diff
@@ -0,0 +1,42 @@
+name: Delta App
+on:
+  pull_request:
+    paths:
+      - 'delta/**'
```
olivergrabinski, line 5: Is this correct? Shouldn't it be `delta/app/**`?

shinyhappydan: No, this should be triggered on all changes, as it's dependent on all parts of the codebase

build.sbt
```diff
@@ -1012,3 +1012,25 @@ addCommandAlias(
 )
 addCommandAlias("build-docs", ";docs/clean;docs/makeSite")
 addCommandAlias("preview-docs", ";docs/clean;docs/previewSite")
+
+val coreModules = List("kernel", "rdf", "sdk", "sourcing-psql", "testkit")
```
olivergrabinski, line 1016: `sourcingPsql`, cf the failed pipeline

imsdu (changes requested):

imsdu (approved):

PR #4108, opened 2023-07-25
fix bug where lenient decoding would never occur

### LINKED ISSUE #4107: Lenient decoding of resources does not work
**Describe the bug**
When I start delta with the `lenient` decoding option for resources set, the decoding happens as if I had set `strict`

- [ ] Delta

**To Reproduce**
1. Start delta with `lenient` option set for resource decoding
2. Post a resource with underscore fields in it (eg `_self`)
3. See that the resource is rejected

**Expected behavior**
On step 3 above, the resource should be accepted

shinyhappydan:
Fixes #4107

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/NexusSource.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/NexusSource.scala
@@ -47,7 +47,6 @@ object NexusSource {
 
     override def apply(c: HCursor): Result[NexusSource] = {
       decoder(c).flatMap { json =>
-
         val underscoreFields = json.asObject.toList.flatMap(_.keys).filter(_.startsWith("_"))
         Either.cond(
           underscoreFields.isEmpty,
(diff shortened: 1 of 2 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/NexusSource.scala
```diff
@@ -42,29 +42,32 @@ object NexusSource {
...
-          if (underscoreFields.nonEmpty) {
-            Left(
-              DecodingFailure(
-                s"Field(s) starting with _ found in payload: ${underscoreFields.mkString(", ")}",
-                c.history
-              )
+    override def apply(c: HCursor): Result[NexusSource] = {
+      decoder(c).flatMap { json =>
+        val underscoreFields = json.asObject.toList.flatMap(_.keys).filter(_.startsWith("_"))
+        if (underscoreFields.nonEmpty) {
```
imsdu, line 51: `Either.cond` could make it a bit nicer to read

imsdu (approved):

PR #4114, opened 2023-07-31
Add documentation for indexing failures

### LINKED ISSUE #3979: List indexing errors by view
**Motivation**
Getting indexing errors as SSE is not practical enough for users so a listing approach would be better.
The endpoint should be similar to the one provided for resources or for projects/resolvers/..., the main difference would be that it provides less filtering options and that it targets Postgres instead of Elasticsearch

**Acceptance criteria**
1. An index is added on the instant column
2. The errors are filtered are for a given view in a given project and sorted by date asc
3. They can be further filtered by date (reusing the `TimeRange` filter introduced lately)
4. Pagination is possible using the existing `Pagination`
5. It will return `SearchResults[IndexingError]`, there may be the need to create a new context/update an existing one
6. The API endopoint would look like:
```
GET /v1/views/{org_label}/{project_label}/{view_id}/errors
?from={from}
 &size={size}
&instant=date..*
```
6. The structure of the response would look like the outcome of the project listing: https://bluebrainnexus.io/docs/delta/api/projects-api.html#list (where we would have indexing errors instead of projects and the contexts will be partly different)

imsdu:
Fixes #3979 

image

image


docs/src/main/paradox/docs/delta/api/views/index.md
```diff
@@ -165,9 +154,52 @@ Request
...
+GET /v1/views/{org_label}/{project_label}/{view_id}/failures/sse
+```
+
+**Example**
+
+Request
+:   @@snip [failures.sh](../assets/views/sse-failures.sh)
+
+Response
+:   @@snip [failures.json](../assets/views/sse-failures.json)
```
olivergrabinski, line 198: I think that the content of the files (or the names) should be switched. The `sse-failures.json` contains what should be in `failures.json` and vice-versa.

olivergrabinski (changes requested):

olivergrabinski (approved):

PR #4131, opened 2023-08-03
Remove dependency on rdf from storage

### LINKED ISSUE #4099: Remove the dependency between the storage app and rdf
**Motivation**
The storage app only depends on a couple of small things from the rdf module.
Yet, it forces us to :
* rebuild and test the storage app every time the rdf or the kernel module are updated.
* have a compatibility with 1.8 for those
* will prevent us to move rdf away from some of the akka library (ex: Uri)

**Acceptance criteria**
1. The storage app does not depend anymore on the rdf or the kernel modules
2. The worflows for storages is not triggered anymore when they are updated
3. We drop the compatibility for 1.8 for kernel and rdf modules

shinyhappydan:
Fixes #4099

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -745,7 +745,6 @@ lazy val storage = project
     buildInfoKeys            := Seq[BuildInfoKey](version),
     buildInfoPackage         := "ch.epfl.bluebrain.nexus.storage.config",
     Docker / packageName     := "nexus-storage",
-    javaSpecificationVersion := "1.8",
     libraryDependencies     ++= Seq(
       apacheCompress,
       apacheIo,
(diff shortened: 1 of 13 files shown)


build.sbt
```diff
@@ -724,7 +726,6 @@ lazy val storage = project
   .in(file("storage"))
   .enablePlugins(UniversalPlugin, JavaAppPackaging, JavaAgent, DockerPlugin, BuildInfoPlugin)
   .settings(shared, compilation, assertJavaVersion, kamonSettings, storageAssemblySettings, coverage, release, servicePackaging)
-  .dependsOn(rdf)
```
imsdu, line 727: You should be able to remove:
```
javaSpecificationVersion := "1.8"
```
From the kernel, rdf, testkit (?) modules.
Also the worflow for storage should not be triggered anymore if those change

imsdu (changes requested):

imsdu (approved):

PR #4135, opened 2023-08-04
Add a dedicated user for aggregations integration tests

imsdu:
Using the same user for the listing and aggregation tests makes tests flaky as some of the tests are made on all orgs and all projects and as both tests create resources, projects and orgs, they interact with each other


shinyhappydan (approved):
Nice spot

olivergrabinski (approved):

PR #4138, opened 2023-08-04
Add documentation for Composite Sink

olivergrabinski:


docs/src/main/paradox/docs/delta/api/views/composite-sink.md
```diff
@@ -0,0 +1,163 @@
+# Composite Sinks
```
olivergrabinski, line 1: Previously when thinking of the projection side of a composite view, we use the term "projection". Now there is much more overlap between "composite sink" and "projection" since the composite sink now regroups pretty much everything. There can also be the confusion with the actual target sink (ES, BG).

docs/src/main/paradox/docs/delta/api/views/composite-sink.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Configuring the Batch Composite Sink
+
+In order to enable the Batch Composite Sink, configure the following Nexus Delta property
+
+`plugins.composite-views.sink-config = batch`
+
+Furthermore, you can configure the maximum size of a batch and the maximum interval using
+
+`plugins.{plugin}.batch.max-elements={max-elements}`
```
imsdu, line 27: Those are the default values but it should be possible to tune them at the composite views level without affecting the ES and BG level no ?

imsdu (changes requested):

docs/src/main/paradox/docs/delta/api/views/index.md
```diff
@@ -6,6 +6,7 @@
 * @ref:[Sparql view](sparql-view-api.md)
 * @ref:[Aggregated Sparql view](aggregated-sparql-view-api.md)
 * @ref:[Composite view](composite-view-api.md)
+* @ref:[Composite sink](composite-sink.md)
```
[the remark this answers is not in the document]
olivergrabinski: I added a paragraph in the `composite-view-api.md` and a link to the new page. The `composite-sink.md` still needs to be referenced in an index for the docs to compile I think 🤔

docs/src/main/paradox/docs/delta/api/views/composite-sink.md
olivergrabinski: Updated

docs/src/main/paradox/docs/delta/api/views/composite-sink.md
```diff
@@ -0,0 +1,173 @@
...
+* `{max-interval}` is the maximum interval of time to wait for `{max-elements}` elements
+
+### How to write a SPARQL construct query for the Batch Composite Sink
+
+In order to use the Batch Composite Sink successfully, it is necessary to rework some aspects of a regular SPARQL
+construct query. We explain the changes through an example.
+
+### Example
+
+Suppose we are in a situation where Composite Views are using the Single Composite Sink and have the following query
```
imsdu, line 58: :

docs/src/main/paradox/docs/delta/api/views/composite-sink.md
```diff
@@ -0,0 +1,173 @@
...
+  BIND({resource_id} AS ?id) .
+  ?id schema:name ?name .
+  
+  OPTIONAL { ?id schema:age ?age }
+  OPTIONAL { ?id schema:parent ?parent . }
+}
+```
+
+Using the default Single Composite Sink, Nexus Delta will query the resources Alice and Bob individually and obtain the
+following n-triples from Blazegraph
```
imsdu, line 77: :

docs/src/main/paradox/docs/delta/api/views/composite-sink.md
```diff
@@ -0,0 +1,173 @@
...
+<http://people.com/Alice> <http://schema.org/parent> <http://people.com/Bob>
+<http://people.com/Bob> <http://schema.org/name> <Bob>
+<http://people.com/Bob> <http://schema.org/age> <42>
+```
+
+Note how the results are the merged result of the individual queries. While we were able to query several resources
+simultaneously, we are now facing a framing problem. If we try to frame `http://people.com/Alice`, its graph now
+contains more information than before; it will now include the age of Bob, something that we did not request.
+
+In order to solve this problem, we will introduce aliasing for the root resource IDs. The query will now become
```
imsdu, line 125: :

docs/src/main/paradox/docs/delta/api/views/composite-sink.md
```diff
@@ -0,0 +1,173 @@
...
+  BIND(IRI(CONCAT(STR(?id), '/alias')) AS ?alias) .
+  
+  ?id schema:name ?name .
+  
+  OPTIONAL { ?id schema:age ?age }
+  OPTIONAL { ?id schema:parent ?parent }
+}
+```
+
+With this query, a batch query for both Alice and Bob will now yield
```
imsdu, line 145: :

imsdu (approved):

docs/src/main/paradox/docs/delta/api/views/composite-sink.md
```diff
@@ -0,0 +1,173 @@
...
+We recommend reading through the @ref:[search configuration example use case](../../../getting-started/running-nexus/search-configuration.md#example-use-case) and the @ref[Composite View API reference](composite-view-api.md) to learn more about Composite Views.
+
+## Batch Composite Sink
+
+@@@ note { .warning }
+
+We recommend to start using Composite Views with the default @ref:[Single Composite Sink](#single-composite-sink). Once
+you have a good understanding of it, the Batch Composite Sink can be used to enhance the performance of your deployment.
+
+@@@
```
shinyhappydan, line 28: I think the warning needs to come after the description

shinyhappydan (dismissed):
I think the only problem is that it's called composite sink:
- It's not really a sink
- It's not a composite of multiple 'sinks'

I think this makes it harder to understand than it should be. The documentation is good

olivergrabinski:
Creating #4190 for the CompositeSink name issue

docs/src/main/paradox/docs/delta/api/views/composite-sink.md
olivergrabinski: Changed

imsdu (approved):

PR #4151, opened 2023-08-09
Add partial reindexing to composite views

imsdu:
Related to #3948 

This first part prepares the model for partial reindexing when updating a composite view:
* The indexingRev represents the last revision where a change impacted the indexing
* Whenever a source is updated, the indexingRev is updated for both sources and projections and it will trigger a full reindexing
* If a projection is updated, only this projection will be restarted

The changes in the indexing process will come in a second PR


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeView.scala
```diff
@@ -55,8 +57,8 @@ import scala.concurrent.duration.FiniteDuration
 final case class CompositeView(
     id: Iri,
     project: ProjectRef,
-    sources: NonEmptySet[CompositeViewSource],
-    projections: NonEmptySet[CompositeViewProjection],
+    sources: NonEmptyMap[Iri, CompositeViewSource],
```
imsdu, line 60: `NonEmptySet` were kind of a pain here and as we mostly want to access sources and projections individually, a `NonEmptyMap` makes more sense

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeViewFields.scala
```diff
@@ -28,8 +27,8 @@ import scala.concurrent.duration.FiniteDuration
 final case class CompositeViewFields(
     name: Option[String],
     description: Option[String],
-    sources: NonEmptySet[CompositeViewSourceFields],
-    projections: NonEmptySet[CompositeViewProjectionFields],
+    sources: NonEmptyList[CompositeViewSourceFields],
```
imsdu, line 30: Same, here, we check for duplication before creating the composite view anyway

shinyhappydan: Is this an advantage? Set is telling us that ordering doesn't matter and the duplicates don't have an effect, which seems good?

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchViewFactory.scala
```diff
@@ -0,0 +1,59 @@
...
+    current.projections(defaultProjectionId).exists {
+      case _: SparqlProjection        =>
+        throw new IllegalStateException("Search views do not include any SparqlProjection")
+      case e: ElasticSearchProjection =>
+        e.query == config.query &&
+          e.mapping == config.mapping &&
+          e.settings == config.settings &&
+          e.context == config.context &&
+          e.resourceTypes == config.resourceTypes
+    } && current.rebuildStrategy == config.rebuildStrategy
```
olivergrabinski, line 55: Shouldn't this also compare the sources?

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeViewFields.scala
imsdu: Yes because `NonEmptyList` is a lot easier to use.
`NonEmptySet` is based on a sorted set, requires an order and did not bring value here
as there is also a check for duplicates on the id of sources/projections when evaluating as sources and projections must all have different ids.

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchViewFactory.scala
olivergrabinski: Never mind

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/circe/Decoders.scala
```diff
@@ -0,0 +1,3 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.circe
+
+class Decoders {}
```
olivergrabinski, line 3: Not used?

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/Transactors.scala
```diff
@@ -37,6 +37,18 @@ final case class Transactors(
   def execDDLs(ddls: List[String])(implicit cl: ClassLoader): Task[Unit] =
     ddls.traverse(execDDL).void
 
+  /**
+    * Init the partition in the events and states table for the given projects
+    */
+  def initPartitions(projects: ProjectRef*): Task[Unit] =
```
olivergrabinski, line 43: If it is only use in tests I would have not put it here

delta/plugins/search/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchViewFactory.scala
olivergrabinski: Still I think from the perspective of someone looking a this for the first time, it might not be obvious why the source is not being checked. Might be worth editing the comment?

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeViewState.scala
```diff
@@ -35,6 +35,8 @@ import scala.annotation.nowarn
   *   the collection of tags
   * @param rev
   *   the current revision of the view
+  * @param indexingRev
```
olivergrabinski, line 38: There is no `indexingRev` in this class

olivergrabinski (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeView.scala
```diff
@@ -74,17 +76,55 @@ final case class CompositeView(
...
+    * Looks for Elasticsearch projections
+    */
+  def elasticSearchProjections: Set[ElasticSearchProjection] = projectionsOf[ElasticSearchProjection]
+
+  /**
+    * Looks for Sparql projections
+    */
+  def sparqlProjections: Set[SparqlProjection] = projectionsOf[SparqlProjection]
+
+  private def projectionsOf[X: ClassTag] =
```
olivergrabinski, line 97: It won't looks super nice but it is possible to avoid reflection for `sparqlProjections` and `elasticSearchProjections`. Probably doesn't really matter though

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeViewProjection.scala
```diff
@@ -37,6 +35,8 @@ sealed trait CompositeViewProjection extends Product with Serializable {
     */
   def uuid: UUID
 
+  def indexingRev: Int
```
olivergrabinski, line 38: Utility of the docstring might be questionable but it is missing

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
```diff
@@ -534,14 +533,9 @@ object CompositeViews {
       case Some(s) if s.deprecated   =>
         IO.raiseError(ViewIsDeprecated(c.id))
       case Some(s)                   =>
+        val newRev = s.rev + 1
```
olivergrabinski, line 536: When `newRev` is used here or in further methods I have to say I keep losing track of what kind of revision we are dealing with

olivergrabinski (changes requested):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeView.scala
```diff
@@ -74,17 +76,55 @@ final case class CompositeView(
...
+    projections.foldLeft(Set.empty[X]) {
+      case (acc, projection: X) => acc + projection
+      case (acc, _)             => acc
+    }
```
shinyhappydan, line 101: You could just use .collect here

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViewFactorySuite.scala
```diff
@@ -0,0 +1,291 @@
...
+    Some(blazegraphProjectionId),
+    SparqlConstructQuery.unsafe("CONSTRUCT..."),
+    schemas,
+    types,
+    tag,
+    includeDeprecated,
+    includeMetadata
+  )
+
+  test("Create the matching source from the project source field with a defined id") {
```
shinyhappydan, line 96: What is a matching source?

shinyhappydan: More broadly I don't really understand what these tests are actually asserting. What is the API doing?

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/migration/MigrateCompositeViewsSuite.scala
```diff
@@ -0,0 +1,176 @@
...
+      _ <- loadState(Tag.latest, "composite-views/database/view-state.json")
+      _ <- loadState(UserTag.unsafe("v1.0"), "composite-views/database/view-state.json")
+      _ <- eventsToMigrate(xas).map(_.assertSize(2))
+      _ <- statesToMigrate(xas).map(_.assertSize(2))
+      _ <- new MigrateCompositeViews(xas).run.assert((2, 2))
+      // Making sure all has been updated
+      _ <- eventsToMigrate(xas).map(_.assertSize(0))
+      _ <- statesToMigrate(xas).map(_.assertSize(0))
+    } yield ()
+  }
```
shinyhappydan, line 70: This shouldn't be a test, it's just setup

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeView.scala
imsdu: projections is a Map, not a set

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViewFactorySuite.scala
imsdu: That the fields are correctly transformed into a value

delta/plugins/composite-views/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/migration/MigrateCompositeViewsSuite.scala
imsdu: It is not only setup, it also run the migration.

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
imsdu: Yep, I try to address that

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViewFactory.scala
```diff
@@ -0,0 +1,143 @@
...
+    */
+  def update(fields: CompositeViewFields, current: CompositeViewValue, nextRev: Int)(implicit
+      projectBase: ProjectBase,
+      uuidF: UUIDF
+  ): UIO[CompositeViewValue] = {
+    for {
+      sources             <- fields.sources.traverse { upsert(_, current.sources.lookup) }.map(_.toNem)
+      // If any source has changed, we update the indexing rev for sources
+      sourceHasChanged     = current.sources != sources
+      newSourceIndexingRev = if (sourceHasChanged) nextRev else current.sourceIndexingRev
```
olivergrabinski, line 48: If the source does not change, but the projection does, the CompositeViewValue does not have a new indexing rev. But then we are in fact reindexing but the name of the projection does change? Maybe I am just confused. But this is how I understand the failing supervision test

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/model/CompositeViewState.scala
olivergrabinski: ℹ️  This is still there

olivergrabinski (approved):

Issue #4081, opened 2023-07-20
Delta startup failure should result in integration-test workflow failure


shinyhappydan:
If a change is made that causes the integration-test delta instance to fail to start, this happens silently. The only way you can tell it has happened is that the integration tests will subsequently run and fail. 

When delta fails to start, the workflow should fail before the tests run and give a description of the problem if possible

PR #4152, opened 2023-08-09
Add health check on Delta for integration tests

olivergrabinski:
* Update compose file version
* Add healthcheck to Delta in the docker compose file
* Check health of the container before running integration tests and abort if not healthy

#4081

### DIFF
--- a/.github/workflows/ci-integration-tests.yml
+++ b/.github/workflows/ci-integration-tests.yml
@@ -29,12 +29,11 @@ jobs:
         run: docker-compose -f tests/docker/docker-compose.yml up -d && sleep 60
       - name: Test
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-integration-tests.yml
```diff
@@ -28,7 +28,15 @@ jobs:
       - name: Start services
         run: docker-compose -f tests/docker/docker-compose.yml up -d && sleep 60
       - name: Test
-        run: sbt -Dsbt.color=always -Dsbt.supershell=false "project tests" test
+        run: |
```
imsdu, line 31: It may be simpler to do the check on curl here ?

olivergrabinski: I suppose it would be more self-contained if it were here 🤔  I will have a look at it later

PR #4192, opened 2023-08-23
Search: new resource types & `curated` field

olivergrabinski:
* Add new resource types
* Add new `curated` field
    * Empty if no curated/unassessed annotation is present
    * `true` if curated annotation present
    * `false` if unassessed annotation present


tests/docker/config/construct-query.sparql
```diff
@@ -267,6 +268,17 @@ CONSTRUCT {
       ) AS ?sTypeIdLabel) .
   } .
 
+  OPTIONAL {
+    ?id nsg:annotation / nsg:hasBody ?curatedAnnotation .
+    BIND(
+      IF(
+         ?curatedAnnotation = <https://neuroshapes.org/Curated>,
+         true,
+         IF(?curatedAnnotation = <https://neuroshapes.org/Unassessed>, false, ?undefined)
```
crisely09, line 277: how is this undefined value used? is it like `None`/ `Null`?

olivergrabinski: This should just be a blank node that is never used, this way if the resource doesn't have the curated/unassessed annotation then there will be no `curated` field at all in the ES document; there is also a test below to ensure this is true. Let me know if you know a nicer way to deal with this!

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigSpec.scala
```diff
@@ -335,6 +339,29 @@ class SearchConfigSpec extends BaseSpec {
...
+      val query = queryField(traceId, "curated")
+      assertEmpty(query)
+    }
+
+    "have curated field true if curated" in {
+      val query    = queryField(curatedTraceId, "curated")
+      val expected = json"""{ "curated": true }"""
+
+      assertOneSource(query) { json =>
+        json should equalIgnoreArrayOrder(expected)
```
shinyhappydan, line 352: `json shouldBe json"""{ "curated": true }"""` ?

shinyhappydan (approved):

imsdu (approved):

PR #4200, opened 2023-08-25
Validate index with indexing rev

### LINKED ISSUE #4154: When updating an Elasticsearch view, the index should be created against the indexing rev
**Describe the bug**
It is currently created against the rev instead of the indexing rev.

imsdu:
Fixes #4154 

* Use `IndexingRev` value class for this property to make it less confusing with rev 
* remove `_indexingRev`as response metadata, it is better if it remains an implementation detail


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
```diff
@@ -480,8 +477,9 @@ object ElasticSearchViews {
       case Some(s) if c.value.tpe != s.value.tpe =>
         IO.raiseError(DifferentElasticSearchViewType(s.id.toString, c.value.tpe, s.value.tpe))
       case Some(s)                               =>
+        val newIndexingRev = nextIndexingRev(s.value, c.value, s.indexingRev, c.rev)
         for {
-          _ <- validate(s.uuid, s.rev + 1, c.value)
+          _ <- validate(s.uuid, newIndexingRev, c.value)
```
imsdu, line 482: The bug was here where rev was passed instead of indexingRev as there were both ints.
So that is why I used the value class for indexingRev so that it is not possible to pass the rev and to eliminate the need for a test there.

Issue #3690, opened 2023-02-17
Filter by type from the SQL query


imsdu:
**Motivation**
The  different views allow to filter resources by type.
At the moment, this operation is handled by Delta during the indexing process.
It should be handled directly by the SQL query to avoid to fetch unnecessary resources just to discard them afterwards.

**Acceptance criteria**
1. An index is added in the table
2. The SQL query includes a restriction of the types to retain
3. The indexing behaviour should remain the same (just faster when the types to retain only represents a fraction of all the resources in the project)

PR #4201, opened 2023-08-28
Filter by type in the SQL query for Elasticsearch views

olivergrabinski:
* Introduces `SelectFilter` which contains data (types and tag) used to filter the SQL query on states.
* Makes use of `SelectFilter` for ES views

#3690


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewValue.scala
```diff
@@ -94,6 +97,20 @@ object ElasticSearchViewValue {
...
+    def selectFilter: SelectFilter = {
+      val types = pipeline
+        .collectFirst {
+          case PipeStep(label, _, Some(config)) if label == FilterByType.ref.label =>
+            val filterByTypeConfig = JsonLdDecoder[FilterByTypeConfig].apply(config)
+            filterByTypeConfig.map(_.types).getOrElse(Set.empty)
+        }
+        .getOrElse(Set.empty)
+      SelectFilter(types, resourceTag.getOrElse(Latest))
+    }
```
olivergrabinski, line 112: Not 100% convinced by this

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/IndexingViewDef.scala
```diff
@@ -44,6 +45,7 @@ object IndexingViewDef {
       projection: String,
       resourceTag: Option[UserTag],
       pipeChain: Option[PipeChain],
+      selectFilter: SelectFilter,
```
imsdu, line 48: I think the resourceTag (line #46) is not used anymore now

imsdu (changes requested):

imsdu (approved):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewValue.scala
shinyhappydan: The only issue I can see is the it's only looking at one step, I don't know if types could be defined in multiple steps

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/query/SelectFilter.scala
```diff
@@ -0,0 +1,34 @@
...
+object SelectFilter {
+
+  /** All types with specified tag */
+  val tag: Tag => SelectFilter = SelectFilter(Set.empty, _)
+
+  /** All types with latest tag */
+  val latest: SelectFilter = SelectFilter(Set.empty, Tag.Latest)
+
+  /** All types with specified tag if it exists, otherwise latest */
+  val tagOrLatest: Option[Tag] => SelectFilter =
```
shinyhappydan, line 31: Personally I find this less clear than if you just use the `tag` method and deal with the option in the caller

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ElasticSearchViewsSpec.scala
```diff
@@ -291,10 +291,10 @@ class ElasticSearchViewsSpec extends BaseSpec with EitherValuable with CirceEq {
         response.status shouldEqual StatusCodes.OK
         val expected = jsonContentOf(
           "/kg/views/statistics.json",
-          "total"     -> "14",
-          "processed" -> "14",
+          "total"     -> "5",
+          "processed" -> "5",
```
shinyhappydan, line 295: I find it really weird that these changed. If this value could be anything, is it really useful?

shinyhappydan (approved):

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/query/SelectFilter.scala
```diff
@@ -0,0 +1,34 @@
...
+  * Contains the information that can be used for filtering when streaming states
+  * @param types
+  *   types that a view is indexing
+  * @param tag
+  *   tag that a view is indexing
+  */
+case class SelectFilter(types: Set[Iri], tag: Tag) {
+
+  /** The types as a postgres array */
+  def typeSqlArray: Fragment =
```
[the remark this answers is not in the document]
olivergrabinski: I can take a look into it when working on the BG selectFilter

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/query/SelectFilter.scala
olivergrabinski: Wouldn't that make the tag optional in the case class? I don't think it should be optional there

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ElasticSearchViewsSpec.scala
olivergrabinski: I see these statistics as "what delta has to work on" as such if we are now filtering types in the db then this is not something processed by delta anymore. I suppose the question is the utility of the test?

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchViewValue.scala
olivergrabinski: I feel like it's technically possible to define twice the same pipe however 1) it's not explicitly supported/encouraged anywhere and 2) I don't think it's tested. I felt this was out of scoped so that's why I left it like this

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ElasticSearchViewsSpec.scala
shinyhappydan: I would question the utility of the statistic itself, but yes the test especially

shinyhappydan: What I mean is: what is the utility of saying that I processed 14 things, when actually I didn't need to process 9 of them? Surely the only meaningful measure is of things that were actually affected

shinyhappydan: To me when a method's efficiency is improved, if that causes a test to break then the test was not good

olivergrabinski: Discarded to me was supposed to let the caller know that X elements have been dropped due to the filters they have specified in the view value. Just saying `processed: 5` doesn't do that. I suppose though with this change it could be confusing as one filter (the type filter) will not be part of the discarded anymore.

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/query/SelectFilter.scala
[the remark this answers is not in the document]
olivergrabinski: Ideally I would have had a fragment encoder for the `types` field

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ElasticSearchViewsSpec.scala
imsdu: We can discuss the statistics endpoint because its meaning has evolved with `1.8` and again with `1.9` and see what we can do (maybe just better document it first).

PR #4229, opened 2023-08-30
all remote endpoints should use the same credentials

shinyhappydan:


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageValue.scala
```diff
@@ -246,11 +246,8 @@ object StorageValue {
       */
     def authToken(config: StorageTypeConfig): Option[AuthToken] =
       config.remoteDisk
-        .flatMap { cfg =>
-          if (endpoint == cfg.defaultEndpoint) cfg.defaultCredentials else None
-        }
+        .flatMap(_.defaultCredentials)
```
imsdu, line 249: I don't know where it is used but having this in `StorageValue` does not make sense anymore as creating the token does not depend on it

shinyhappydan: I agree but this refactoring will make more sense when I'm updating auth

imsdu (dismissed):

PR #4231, opened 2023-08-30
Filter by type in the SQL query for Composite Views

olivergrabinski:
Screenshot 2023-08-30 at 16 22 30


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ElemRoutes.scala
```diff
@@ -57,18 +57,18 @@ class ElemRoutes(
                 concat(
                   (get & pathPrefix("continuous") & parameter("tag".as[UserTag].?)) { tag =>
```
imsdu, line 58: We need to get the types as an input there to pass them down as an input.
So also a change in the `DeltaClient`

imsdu (changes requested):

imsdu (approved):

PR #4234, opened 2023-08-31
Refactor remote storage auth to be from a function

shinyhappydan:


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/remote/client/RemoteDiskStorageClient.scala
```diff
@@ -32,7 +32,11 @@ import scala.concurrent.duration._
 /**
   * The client to communicate with the remote storage service
   */
-final class RemoteDiskStorageClient(baseUri: BaseUri)(implicit client: HttpClient, as: ActorSystem) {
+final class RemoteDiskStorageClient(baseUri: BaseUri)(implicit
```
imsdu, line 35: Now that the baseUri is the only thing that is not coming from the configuration, it feels like we could instantiate the remote client only once so as to pass it directly instead of its dependencies and building it each time.

The `baseUri` could become an implicit parameter for the different method

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/remote/RemoteStorageAuthTokenProvider.scala
```diff
@@ -0,0 +1,23 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.remote
+
+import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.StoragesConfig.StorageTypeConfig
+import ch.epfl.bluebrain.nexus.delta.sdk.identities.model.AuthToken
+import monix.bio.UIO
+
+trait RemoteStorageAuthTokenProvider {
```
imsdu, line 7: To rename and to move to the SDK so as to reuse it for composite views

imsdu (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -388,7 +395,9 @@ final class Files(
       (storageRef, storage) <- fetchActiveStorage(storageId, ref, pc)
       resolvedFilename      <- IO.fromOption(filename.orElse(path.lastSegment), InvalidFileLink(iri))
       description           <- FileDescription(resolvedFilename, mediaType)
-      attributes            <- LinkFile(storage).apply(path, description).mapError(LinkRejection(iri, storage.id, _))
+      attributes            <- LinkFile(storage, config, remoteDiskStorageClient)
```
imsdu, line 398: It would be nice to pass the `remoteDiskStorageClient` and the `config ` in a consistent order for the different operations

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/remote/AuthTokenProvider.scala
```diff
@@ -0,0 +1,23 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.remote
+
+import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.StoragesConfig.StorageTypeConfig
+import ch.epfl.bluebrain.nexus.delta.sdk.identities.model.AuthToken
+import monix.bio.UIO
+
+trait AuthTokenProvider {
```
imsdu, line 7: Docs

imsdu (changes requested):

imsdu (approved):

imsdu (approved):

Issue #1751, opened 2020-11-04
Find a way to tell which resources should be updated after a context update


umbreak:
Somewhat related to: https://github.com/BlueBrain/nexus/issues/1038

We have a way to guarantee resource immutability, but we don't have a way to tell suggest the clients which resources should be updated once a context changes.

Remember that changes can be in a secondary context, as for example:

to find which context is used in which. resource is not trivial, since you can have

- Resource
```
{
   "@context": "http://example.com/a",
    ...
}
```

- Context a
```
{
   "@context": [{...}, "http://example.com/a"],
   "@id": "http://example.com/a"
}
```

- Context b
```
{
   "@context": {...},
   "@id": "http://example.com/b"
}
```

If either context a or context b changes, the resource should be suggested to the client as potential change.

PR #4235, opened 2023-08-31
Keep track of the remote contexts after resolution

imsdu:
Related to #1751

The changes to save them in events/states will come in a next PR


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImpl.scala
```diff
@@ -44,7 +44,9 @@ final class ResourcesImpl private (
     for {
       projectContext             <- fetchContext.onCreate(projectRef)
       schemeRef                  <- expandResourceRef(schema, projectContext)
-      (iri, compacted, expanded) <- sourceParser(projectRef, projectContext, source)
+      (iri, compacted, expanded) <- sourceParser(projectRef, projectContext, source).map { j =>
+                                      (j.iri, j.compacted, j.expanded)
+                                    }
```
olivergrabinski, line 49: Using the JsonLdResult unapply could be an alternative though it can be looked at later

olivergrabinski (approved):

Issue #3994, opened 2023-06-19
Complete and enhance search config integration tests


olivergrabinski:
- [ ] Complete pending tests in `SearchConfigSpec`
- [ ] Assert that basic search queries are ok (ex: aggregations)

PR #4236, opened 2023-08-31
Complete search query integration tests

olivergrabinski:
#3994


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigSpec.scala
```diff
@@ -87,6 +91,77 @@ class SearchConfigSpec extends BaseSpec {
...
+      val query = queryField(neuronMorphologyId, "description")
+      assertOneSource(query) { json =>
+        json shouldEqual json"""{ "description": "This is a resource description." }"""
+      }
+    }
+
+    "have the correct createdAt property" in {
+      val query = queryField(neuronMorphologyId, "createdAt")
+      assertOneSource(query) { json =>
+        assert(isInstant(json, "createdAt"))
```
olivergrabinski, line 125: Just checking that the value can be parsed as an `Instant` for `createdAt` and `updatedAt`

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigSpec.scala
```diff
@@ -567,11 +642,34 @@ class SearchConfigSpec extends BaseSpec {
       }
     }
 
-    "have the correct metadata property" in { pending }
-    "have the correct detailed circuit property" in { pending }
-    "have the correct simulation campaign config property" in { pending }
```
olivergrabinski, line 572: This has been taken care of already

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigSpec.scala
```diff
@@ -87,6 +91,77 @@ class SearchConfigSpec extends BaseSpec {
...
+        json shouldEqual json"""{ "name": "Neuron density: CA1" }"""
+      }
+    }
+
+    "have the correct name property from skos:prefLabel" in {
+      val query = queryField(traceId, "name")
+      assertOneSource(query) { json =>
+        json shouldEqual json"""{ "name": "S1J_L6_IPC_cADpyr_2" }"""
+      }
+    }
```
shinyhappydan, line 113: I think this is a general problem with these tests, but I'll write my example here

I would look at this and wonder what it's asserting. Perhaps another way that this could be written
```
"have a name field when the resource only has skos:prefLabel" in {
  assertOneSource(queryField(traceId, "name")) {
    json should have field("name", traceSkosPrefLabel)
  }
}
```
I think the test name is the only place where you can explain why the name property should have a certain value.
"S1J_L6_IPC_cADpyr_2" is a magic value, but by labelling it you can explain what it really is

olivergrabinski: I won't make changes now but we should soon enough have updates to do there so and we can rework tests then

PR #4241, opened 2023-09-03
Update the authentication process for remote storages

### LINKED ISSUE #4063: Update the authentication process for remote storages
**Motivation**
Delta communicates with the remote storage via HTTP and this is secured with a JWT token.
The issue is that Delta needs a long-term token to handle this as it is currently not able to renew it.
This is also a bad practice to have a token with a long lifespan.

**Acceptance criteria**
1. The remote storage configuration now asks for a client id and a secret
2. The remote storage client is able to retrieve a token and to ask for a new one when the current one is expired

shinyhappydan:
Fixes #4063 

Remaining tasks:

- [x] Config option for the existing approach (token)
- [ ] Possible change to construction of KeyValueStore
- [ ] Possible move of construction into sdk module


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePluginModule.scala
```diff
@@ -146,8 +147,8 @@ class StoragePluginModule(priority: Int) extends ModuleDef {
 
   many[ResourceShift[_, _, _]].ref[Storage.Shift]
 
-  make[AuthTokenProvider].from { (cfg: StorageTypeConfig) =>
-    AuthTokenProvider(cfg)
+  make[AuthTokenProvider].from { (cfg: StorageTypeConfig, httpClient: HttpClient @Id("storage"), realms: Realms) =>
```
shinyhappydan, line 150: Is this the right http client to use?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,78 @@
...
+          )
+          .toEntity
+      )
+    ).hideErrors
+      .flatMap { json =>
+        json.hcursor.get[String]("access_token") match {
+          case Left(failure) => IO.terminate(new RuntimeException("no access_token in response: " + failure))
+          case Right(value) => UIO.pure(value)
+        }
+      }
```
shinyhappydan, line 74: both .hideErrors and the RuntimeException need to be something better

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthenticateAs.scala
```diff
@@ -0,0 +1,12 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.auth
+
+import ch.epfl.bluebrain.nexus.delta.kernel.Secret
+import pureconfig.ConfigReader
+import pureconfig.generic.semiauto.deriveReader
+
+
+case class AuthenticateAs(user: String, password: Secret[String], realm: String)
```
shinyhappydan, line 8: This needs to be an ADT: Anonymous is the other alternative

tests/docker/config/delta-postgres.conf
```diff
@@ -107,6 +107,11 @@ plugins {
 
       remote-disk {
         enabled = true
+        authentication: {
+          user: "delta"
+          password: "shhh"
+          realm: "internal"
+        }
```
shinyhappydan, line 114: This matches what is in the tests. I don't see a way to define the values in a single place, so this will have to do

tests/src/test/resources/kg/storages/remote-disk-response.json
```diff
@@ -15,9 +15,9 @@
   "readPermission": "{{read}}",
   "writePermission": "{{write}}",
   "_algorithm": "SHA-256",
-  "_incoming": "{{deltaUri}}/storages/{{project}}/nxv:{{id}}/incoming",
-  "_outgoing": "{{deltaUri}}/storages/{{project}}/nxv:{{id}}/outgoing",
-  "_self": "{{deltaUri}}/storages/{{project}}/nxv:{{id}}",
+  "_incoming": "{{self}}/incoming",
+  "_outgoing": "{{self}}/outgoing",
+  "_self": "{{self}}",
```
shinyhappydan, line 20: This is just from the self changes, missed because this test was ignored

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,78 @@
...
+  override def apply(): UIO[Option[AuthToken]] = UIO.pure(None)
+}
+
+private class KeycloakAuthTokenProvider(auth: AuthenticateAs, httpClient: HttpClient, realms: Realms) extends AuthTokenProvider with MigrateEffectSyntax {
+  override def apply(): UIO[Option[AuthToken]] = {
+    for {
+      realm <- realms.fetch(Label.unsafe(auth.realm)).toUIO
+      accessToken <- doAuth(realm, auth.user, auth.password)
+    } yield Some(AuthToken(accessToken))
+  }
```
shinyhappydan, line 51: We need to cache and invalidate that cache here

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/cache/KeyValueStore.scala
```diff
@@ -217,6 +217,14 @@ trait KeyValueStore[K, V] {
...
 object KeyValueStore {
 
+  def create[K, V](): KeyValueStore[K, V] = {
+    val cache: Cache[K, V] =
+      Caffeine
+        .newBuilder()
+        .build[K, V]()
+    new LocalCache(cache)
+  }
+
```
shinyhappydan, line 227: I need another way to create a cache where it's build directly rather than deferred

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,61 @@
...
+}
+
+private class AnonymousAuthTokenProvider extends AuthTokenProvider {
+  override def apply(): UIO[Option[AuthToken]] = UIO.pure(None)
+}
+
+private class CachingKeycloakAuthTokenProvider(identity: AuthenticateAs, service: KeycloakAuthService)(implicit
+    clock: Clock[UIO]
+) extends AuthTokenProvider {
+  private val cache = KeyValueStore.create[Unit, AccessTokenWithMetadata]()
```
shinyhappydan, line 35: I wanted to use existing code, but I don't think this class is a great fit. We really want something where it deals with multiple threads accessing the value at once and not fetching X times

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/KeycloakAuthService.scala
```diff
@@ -0,0 +1,69 @@
...
+import io.circe.Json
+import monix.bio.{IO, UIO}
+
+import java.time.{Duration, Instant}
+
+class KeycloakAuthService(httpClient: HttpClient, realms: Realms)(implicit clock: Clock[UIO]) {
+
+  def auth(auth: AuthenticateAs): UIO[AccessTokenWithMetadata] = {
+    for {
+      realm                  <- realms.fetch(Label.unsafe(auth.realm)).toUIO
```
shinyhappydan, line 24: Perhaps it's better if realm resolution is outside of this class? Not sure. It's probably clearer the correct design when it's tested

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/error/TokenError.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  implicit val identityErrorEncoder: Encoder.AsObject[TokenError] = {
+    Encoder.AsObject.instance[TokenError] {
+      case TokenHttpError(r)           =>
+        JsonObject(keywords.tpe := "TokenHttpError", "reason" := r.reason)
+      case TokenNotFoundInResponse(r)  =>
+        JsonObject(keywords.tpe -> "TokenNotFoundInResponse".asJson, "reason" := r.message)
+      case ExpiryNotFoundInResponse(r) =>
+        JsonObject(keywords.tpe -> "ExpiryNotFoundInResponse".asJson, "reason" := r.message)
+    }
```
shinyhappydan, line 43: I tried semi-automatic derivation, but the error messages are bad and I couldn't see the problem

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
shinyhappydan: It feels like there should be a scala library that can do this

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,61 @@
...
+}
+
+object AuthTokenProvider {
+  def apply(auth: Option[AuthenticateAs], keycloakAuthService: KeycloakAuthService): AuthTokenProvider = {
+    auth match {
+      case Some(authAs) => new CachingKeycloakAuthTokenProvider(authAs, keycloakAuthService)
+      case None         => new AnonymousAuthTokenProvider
+    }
+  }
+  def anonymousForTest: AuthTokenProvider = new AnonymousAuthTokenProvider
```
shinyhappydan, line 25: Perhaps we don't need to say test here, I was just trying to communicate that this wasn't really for prod code

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePluginModule.scala
```diff
@@ -146,8 +147,12 @@ class StoragePluginModule(priority: Int) extends ModuleDef {
 
   many[ResourceShift[_, _, _]].ref[Storage.Shift]
 
-  make[AuthTokenProvider].from { (cfg: StorageTypeConfig) =>
-    AuthTokenProvider(cfg)
+  make[KeycloakAuthService].from { (httpClient: HttpClient @Id("storage"), realms: Realms, clock: Clock[UIO]) =>
```
imsdu, line 150: This needs to go in `sdk` and use the same http client as `Identities`

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/cache/KeyValueStore.scala
imsdu: You should not need to introduce this method

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthenticateAs.scala
imsdu: Rename to `Credentials` ?

We also need one with Token at least for a while.
For retrocompatibility reasons but also because it will be a safety net if we detect a bug in the new implementation.
The remote storage is a sensitive thing :)

imsdu: Also the realm should be a label

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/KeycloakAuthService.scala
```diff
@@ -0,0 +1,69 @@
...
+                  case Left(failure) => IO.terminate(ExpiryNotFoundInResponse(failure))
+                  case Right(value)  => UIO.pure(value)
+                }
+    } yield {
+      (token, Duration.ofSeconds(expiry))
+    }
+  }
+}
+
+case class AccessTokenWithMetadata(token: String, expiresAt: Instant)
```
imsdu, line 69: Why not `ParsedToken` ?

imsdu (changes requested):

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/cache/KeyValueStore.scala
shinyhappydan: The other methods return an IO[KeyValueStore], how should I get just a KeyValueStore?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,61 @@
...
+}
+
+private class AnonymousAuthTokenProvider extends AuthTokenProvider {
+  override def apply(): UIO[Option[AuthToken]] = UIO.pure(None)
+}
+
+private class CachingKeycloakAuthTokenProvider(identity: AuthenticateAs, service: KeycloakAuthService)(implicit
```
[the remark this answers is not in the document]
shinyhappydan: I think given that our Keycloak setup is not necessarily OpenId (we had issues with this recently when trying to switch to cognito) perhaps it's better to say Keycloak

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,61 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.identities.model.AuthToken
+import monix.bio.UIO
+
+import java.time.{Duration, Instant}
+
+/**
+  * Provides an auth token for the service account, for use when comunicating with remote storage
+  */
+trait AuthTokenProvider {
+  def apply(): UIO[Option[AuthToken]]
```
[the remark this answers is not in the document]
shinyhappydan: There is only one set of credentials, so this isn't necessary

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/KeycloakAuthService.scala
shinyhappydan: It's just not really helpful. We would also have to write parsedExpiresAt

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthenticateAs.scala
shinyhappydan: Can you give an example of where we have an ADT in the config? Is it correct that we need to have a type field which states which class a user is specifying?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/KeycloakAuthService.scala
shinyhappydan: Sorry, I see now you meant the class name. I suppose the fact it was parsed isn't really what is important. It's the access token alongside the expiry date. I don't think saying that it was parsed helps understand what it is

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePluginModule.scala
shinyhappydan: I don't see any modules inside of sdk? Also it's using the config from the storage part

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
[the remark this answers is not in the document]
imsdu: Don't forget we will have to do the same thing for composite views

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,74 @@
...
+) extends AuthTokenProvider {
+  private val cache = KeyValueStore.create[Unit, AccessTokenWithMetadata]()
+
+  override def apply(): UIO[Option[AuthToken]] = {
+    for {
+      existingValue <- cache.get(())
+      now           <- IOUtils.instant
+      finalValue    <- existingValue match {
+                         case None                                 => fetchValue
+                         case Some(value) if isExpired(value, now) => fetchValue
```
imsdu, line 56: It would be nice to have some log whenever we fetch a new token

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/Credentials.scala
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  case class AuthToken(token: String) extends AuthMethod
+  case object AuthToken {
+    implicit val configReader: ConfigReader[AuthToken] = deriveReader[AuthToken]
+  }
+  case class Credentials(user: String, password: Secret[String], realm: Label) extends AuthMethod
+  object Credentials    {
+    @nowarn("cat=unused")
+    implicit private val labelConfigReader: ConfigReader[Label] = ConfigReader.fromString(str =>
```
imsdu, line 25: We don't already have one for Label ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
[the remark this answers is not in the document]
shinyhappydan: When we have two then it's an easy refactor, I don't want to design for something that doesn't exist yet

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/Credentials.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import ch.epfl.bluebrain.nexus.delta.kernel.Secret
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.Label
+import pureconfig.ConfigReader
+import pureconfig.error.CannotConvert
+import pureconfig.generic.semiauto.deriveReader
+
+import scala.annotation.nowarn
+
+sealed trait AuthMethod
```
imsdu, line 11: `Credentials`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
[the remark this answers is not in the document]
shinyhappydan: For example, one design might be that you have two auth token providers, which wouldn't need that change. I think it's better to change the design when it's needed. It would be a very simple change to make later

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/Credentials.scala
```diff
@@ -0,0 +1,32 @@
...
+object AuthMethod {
+  case object Anonymous extends AuthMethod {
+    implicit val configReader: ConfigReader[Anonymous.type] = deriveReader[Anonymous.type]
+  }
+
+  case class AuthToken(token: String) extends AuthMethod
+  case object AuthToken {
+    implicit val configReader: ConfigReader[AuthToken] = deriveReader[AuthToken]
+  }
+  case class Credentials(user: String, password: Secret[String], realm: Label) extends AuthMethod
```
imsdu, line 22: `ClientCredentials`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/Credentials.scala
```diff
@@ -0,0 +1,32 @@
...
+import scala.annotation.nowarn
+
+sealed trait AuthMethod
+
+object AuthMethod {
+  case object Anonymous extends AuthMethod {
+    implicit val configReader: ConfigReader[Anonymous.type] = deriveReader[Anonymous.type]
+  }
+
+  case class AuthToken(token: String) extends AuthMethod
```
imsdu, line 18: JWTToken

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/error/TokenError.scala
imsdu: What do you mean ?
Also if you only use the message/reason from `DecodingFailure`/  `HttpClientError` because it is what we are interested in here, you should not run into this

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/error/TokenError.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.contexts
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.ContextValue
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.JsonLdContext.keywords
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.encoder.JsonLdEncoder
+import ch.epfl.bluebrain.nexus.delta.sdk.http.HttpClientError
+import io.circe.syntax.{EncoderOps, KeyOps}
+import io.circe.{DecodingFailure, Encoder, JsonObject}
+
+sealed abstract class TokenError(reason: String) extends SDKError {
```
imsdu, line 11: `AuthTokenError` or whatever makes it easier to distinguish from the user tokens who can also raise errors

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/cache/KeyValueStore.scala
imsdu: Well, you just get a `IO[AuthTokenProvider]`, distage can handle that

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,74 @@
...
+trait AuthTokenProvider {
+  def apply(): UIO[Option[AuthToken]]
+}
+
+object AuthTokenProvider {
+  def apply(auth: Option[AuthMethod], keycloakAuthService: KeycloakAuthService): AuthTokenProvider = {
+    auth match {
+      case Some(credentials: Credentials)    => new CachingKeycloakAuthTokenProvider(credentials, keycloakAuthService)
+      case Some(AuthMethod.AuthToken(token)) => new FixedAuthTokenProvider(AuthToken(token))
+      case Some(Anonymous) | None            => new AnonymousAuthTokenProvider
```
imsdu, line 24: Now that there is an ADT, no need for an option

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesConfig.scala
```diff
@@ -198,7 +199,7 @@ object StoragesConfig {
   final case class RemoteDiskStorageConfig(
       digestAlgorithm: DigestAlgorithm,
       defaultEndpoint: BaseUri,
-      defaultCredentials: Option[Secret[String]],
+      authentication: Option[AuthMethod],
```
imsdu, line 202: `credentials` without an option

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+  def anonymousForTest: AuthTokenProvider = new AnonymousAuthTokenProvider
+}
+
+private class AnonymousAuthTokenProvider extends AuthTokenProvider {
```
imsdu, line 30: Passing the `credentials` as a parameter also remove the need of subclassing here

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/KeycloakAuthService.scala
```diff
@@ -0,0 +1,70 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.auth.AuthMethod.Credentials
+import ch.epfl.bluebrain.nexus.delta.sdk.error.TokenError.{ExpiryNotFoundInResponse, TokenHttpError, TokenNotFoundInResponse}
+import ch.epfl.bluebrain.nexus.delta.sdk.http.HttpClient
+import ch.epfl.bluebrain.nexus.delta.sdk.realms.Realms
+import io.circe.Json
+import monix.bio.{IO, UIO}
+
+import java.time.{Duration, Instant}
+
+class KeycloakAuthService(httpClient: HttpClient, realms: Realms)(implicit clock: Clock[UIO])
```
imsdu, line 20: Drop the keycloak term to be consistent with the other parts of the codebase and the documentation.
We also realms depending on github for example

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/KeycloakAuthService.scala
```diff
@@ -0,0 +1,70 @@
...
+    extends MigrateEffectSyntax {
+
+  def auth(auth: Credentials): UIO[AccessTokenWithMetadata] = {
+    for {
+      realm                  <- realms.fetch(auth.realm).toUIO
```
imsdu: The realm should not be deprecated

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/Credentials.scala
shinyhappydan: If we do I don't know where it is

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
shinyhappydan: We can debug log, do you want the value of the token logged?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
shinyhappydan: I don't understand, anonymous doesn't need credentials?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/KeycloakAuthService.scala
shinyhappydan: I don't understand what you mean here

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/cache/KeyValueStore.scala
shinyhappydan: Yes but it's confusing that a constructor returns an IO when no IO is necessary. I propose that either we keep the new method, or we change the existing one to return normally rather than an IO

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/error/TokenError.scala
shinyhappydan: When using semi-automatic derivation for an Encoder[TokenError], there were errors which were unclear. I tried changing some obvious things, but could not fix the issue. I noticed that we had used this method elsewhere in the codebase, so I did that.

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/cache/KeyValueStore.scala
imsdu: Creating a cache is a side-effect, that is why it is wrapped in an IO

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePluginModule.scala
```diff
@@ -146,8 +147,12 @@ class StoragePluginModule(priority: Int) extends ModuleDef {
 
   many[ResourceShift[_, _, _]].ref[Storage.Shift]
 
-  make[AuthTokenProvider].from { (cfg: StorageTypeConfig) =>
-    AuthTokenProvider(cfg)
+  make[OpenIdAuthService].from { (httpClient: HttpClient @Id("realm"), realms: Realms, clock: Clock[UIO]) =>
+    new OpenIdAuthService(httpClient, realms)(clock)
+  }
+
+  make[AuthTokenProvider].from { (cfg: StorageTypeConfig, keycloakAuthService: OpenIdAuthService) =>
```
imsdu, line 154: Rename `keycloakAuthService`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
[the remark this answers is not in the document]
imsdu: We don't need two instances, we should avoid that when we can.
The design would look like:
```
trait AuthTokenProvider {
  def apply(credentials: Credentials): UIO[Option[AuthToken]]
}

object AuthTokenProvider {

  def apply(...) = new AuthTokenProvider {
    override def apply(credentials: Credentials): UIO[Option[AuthToken]] =
      credentials match {
        case Anonymous => UIO.none
        case _ => ???
      }
  }

}
```

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
imsdu: See previous comment

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,84 @@
...
+  private val logger = Logger.cats[CachingOpenIdAuthTokenProvider]
+  private val cache  = KeyValueStore.create[Unit, AccessTokenWithMetadata]()
+
+  override def apply(): UIO[Option[AuthToken]] = {
+    for {
+      existingValue <- cache.get(())
+      now           <- IOUtils.instant
+      finalValue    <- existingValue match {
+                         case None                                 =>
+                           logger.debug("fetching auth token, no initial value").toUIO >>
```
imsdu, line 62: Capitalize

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+  * Provides an auth token for the service account, for use when comunicating with remote storage
+  */
+trait AuthTokenProvider {
+  def apply(): UIO[Option[AuthToken]]
+}
+
+object AuthTokenProvider {
+  def apply(credentials: Credentials, keycloakAuthService: OpenIdAuthService): AuthTokenProvider = {
```
imsdu, line 22: Rename `keycloakAuthService`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -0,0 +1,84 @@
...
+  override def apply(): UIO[Option[AuthToken]] = {
+    for {
+      existingValue <- cache.get(())
+      now           <- IOUtils.instant
+      finalValue    <- existingValue match {
+                         case None                                 =>
+                           logger.debug("fetching auth token, no initial value").toUIO >>
+                             fetchValue
+                         case Some(value) if isExpired(value, now) =>
+                           logger.debug("fetching new auth token, current value near expiry").toUIO >>
```
imsdu, line 65: We can put info for now while, the token life is hours or days so it should not pollute two much

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/OpenIdAuthService.scala
```diff
@@ -0,0 +1,70 @@
...
+                  case Left(failure) => IO.terminate(ExpiryNotFoundInResponse(failure))
+                  case Right(value)  => UIO.pure(value)
+                }
+    } yield {
+      (token, Duration.ofSeconds(expiry))
+    }
+  }
+}
+
+case class AccessTokenWithMetadata(token: String, expiresAt: Instant)
```
imsdu, line 70: My previous comment has been discarded it seems but the idea was:
* Rename to `TokenResponse`
* Match response fields names as they are part of the OpenId spec
* Use derivation with snake_case

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/error/AuthTokenError.scala
```diff
@@ -0,0 +1,48 @@
...
+  /**
+    * Signals that an HTTP error occurred when fetching the token
+    */
+  final case class AuthTokenHttpError(cause: HttpClientError)
+      extends AuthTokenError(s"HTTP error when requesting auth token: ${cause.reason}")
+
+  /**
+    * Signals that the token was missing from the authentication response
+    */
+  final case class AuthTokenNotFoundInResponse(failure: DecodingFailure)
```
imsdu, line 26: If you use derivation, you would not need those.
Plus the token/expiry fields are in the openid spec so they will be there

docs/src/main/paradox/docs/getting-started/running-nexus/configuration/index.md
```diff
@@ -129,7 +129,7 @@ Nexus Delta supports 3 types of storages: 'disk', 'amazon' (s3 compatible) and '
 
 - For disk storages the most relevant configuration flag is `plugins.storage.storages.disk.default-volume`, which defines the default location in the Nexus Delta filesystem where the files using that storage are going to be saved.
 - For S3 compatible storages the most relevant configuration flags are the ones related to the S3 settings: `plugins.storage.storages.amazon.default-endpoint`, `plugins.storage.storages.amazon.default-access-key` and `plugins.storage.storages.amazon.default-secret-key`.
-- For remote disk storages the most relevant configuration flags are `plugins.storage.storages.remote-disk.default-endpoint` (the endpoint where the remote storage service is running) and `plugins.storage.storages.remote-disk.default-credentials` (the Bearer token to authenticate to the remote storage service).
+- For remote disk storages the most relevant configuration flags are `plugins.storage.storages.remote-disk.default-endpoint` (the endpoint where the remote storage service is running) and `plugins.storage.storages.remote-disk.authentiction` (the method to authenticate to the remote storage service).
```
imsdu, line 132: `plugins.storage.storages.remote-disk.credentials

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/error/AuthTokenError.scala
shinyhappydan: What do you mean by derivation? I realise that the fields should be there, but this is just me dealing with the errors and giving them good names

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/OpenIdAuthService.scala
shinyhappydan: I'm not sure what the point of that would be? It's clearer with the interface how it is. We shouldn't need to care about the field names outside of integration with OpenId

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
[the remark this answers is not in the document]
shinyhappydan: What is the problem with two instances? 

I think a class that does two auth methods is less simple than two classes that are responsible for a single auth method each

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/OpenIdAuthService.scala
imsdu: Let's forget about my previous comment.
I think I said it before then the comment got discarded and I forgot about it but we should use `ParsedToken` because we need the same validation than the one that is made by the identity module (with more checks checking that the issuer is present in the token and other things).
Otherwise we may end up with a token that is validated here, passed on to the storage and when the token gets back, it gets rejected by the identities endpoint

docs/src/main/paradox/docs/releases/v1.9-release-notes.md
```diff
@@ -114,6 +114,41 @@ Storages can no longer be created with credentials that would get stored:
 
 These should instead be defined in the Delta configuration.
 
+#### Remote storages
```
imsdu, line 117: To put in the storages page and a link here to the chapter in the storage page

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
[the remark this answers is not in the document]
imsdu: * You need the authService to instantiate each of them anyway
* When we implement the thing for composite views, it will be easier to integrate in Distage as there will be only one instance of `AuthTokenProvider`

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/cache/KeyValueStore.scala
imsdu: Use the method which returns an IO

imsdu (changes requested):

shinyhappydan:
image

image

image

imsdu (dismissed):

imsdu (approved):

PR #4244, opened 2023-09-04
Add endpoint for remote contexts + documentation

### LINKED ISSUE #1751: Find a way to tell which resources should be updated after a context update

imsdu:
Fixes #1751 

image


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ResourcesSpec.scala
```diff
@@ -512,7 +510,6 @@ class ResourcesSpec extends BaseSpec with EitherValuable with CirceEq {
                  (_, response) =>
                    response.status shouldEqual StatusCodes.Created
                }
-        _ <- Task.sleep(100.millis)
```
imsdu, line 515: Was slowing down the test for nothing

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/model/jsonld/RemoteContextRef.scala
```diff
@@ -41,15 +44,15 @@ object RemoteContextRef {
   final case class StaticContextRef(iri: Iri) extends RemoteContextRef
 
   /**
-    * A reference to a context registered in Nexus
+    * A reference to a context registered in a Nexus project
     * @param id
```
olivergrabinski, line 48: Iri

docs/src/main/paradox/docs/releases/v1.9-release-notes.md
```diff
@@ -42,6 +42,13 @@ It is now possible to aggregate resources by `@type` or `project`.
 
 @ref:[More information](../delta/api/resources-api.md#aggregations)
 
+#### Remote contexts
+
+When creating/updating a resource, Nexus Delta now keeps track of the remote contexts that have been resolved during the operation,
+allowing to record for user-defined contexts in which project and at which revision this context is pointing too.
```
olivergrabinski, line 48: the part after `allowing to...` is a bit convoluted;
too -> to

olivergrabinski (approved):

PR #4245, opened 2023-09-04
Correct typeOperator behavior when using negations

### LINKED ISSUE #4100: Listing API doesn't respect type operator for multiple excluded types
In the resource list API, if you specify multiple excluded types with the `and` type operator, this does not work. It will give the same result as if you had used `or` instead

For example, if you exclude the types `mammal` and `cat` with the `and` type operator, you would expect other mammals such as `dog` to appear, but these are excluded also

- [ ] Fusion
- [x] Delta
- [ ] CLI
- [ ] Documentation

**Expected behavior**
If I have excluded the types "mammal" and "cat" with an `and` type operator, then resources with the `dog` type should appear

olivergrabinski:
Fixes #4100


delta/plugins/elasticsearch/src/test/resources/query/query-schema-deprecation-types.json
```diff
@@ -2,14 +2,26 @@
   "query": {
     "bool": {
       "must_not": {
-        "term": {
-          "@type": "https://bluebrain.github.io/nexus/vocabulary/CrossProject"
+        "bool": {
```
imsdu, line 5: I think it better to delete the related test, the important is to test the output, not to test the query

imsdu (changes requested):

imsdu (approved):

PR #4246, opened 2023-09-04
Add annotated source as a format for multi-fetch and archives

### LINKED ISSUE #4203: Add the ability to return anotated source for archives and multi-fetch
**Motivation**
Introduced in 1.8, annotated source allows to return the original payload with the Nexus metadata:
https://bluebrainnexus.io/docs/delta/api/resources-api.html#fetch-original-payload
The option to retrieve it for archives and the multi-fetch operation is currently missing.

**Acceptance criteria**
1. When the user selects the annotated source format when creating an archive / performing a multi-fetch operation 
2. The resource is returned in the same way as the fetch operation (original payload without dropping the null values + metadata)

imsdu:
Fixes #4203


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -292,30 +291,9 @@ object ResourcesRoutes {
       decodingOption: DecodingOption
   ): Route = new ResourcesRoutes(identities, aclCheck, resources, projectsDirectives, index).routes
 
-  implicit private val api: JsonLdApi = JsonLdJavaApi.lenient
-
   def asSourceWithMetadata(
       resource: ResourceF[Resource]
-  )(implicit baseUri: BaseUri, cr: RemoteContextResolution): IO[ResourceRejection, Json] = {
```
imsdu, line 299: Extracted to `AnnotatedSource` as it is now reused for archives and multi-fetch

olivergrabinski (approved):

Issue #4102, opened 2023-07-24
Allow users to perform dry-runs on resources (with optional validation with a schema)


imsdu:
**Motivation**
Delta does not provide any endpoint that would allow users to perform dry-runs on resources when this can be useful to test and debug without performing any writes to the primary store.

As the lower layers of Delta already allow to perform dry-runs, we want to make this available with a new endpoint which would allow users to test resources with and without a schema validation

**Acceptance criteria**
1. When a user calls this endpoint, no write operation to the primary store is done
2. The user must provide a valid resource payload following the same rules as the create and update operations of a resource
3. Optionally the user can also provide a schema to validate this resource with:
  * A reference to an existing schema (with an optional revision/tag)
  * The payload of a new schema
4. When the resource is valid, then it is returned to the user (the same as the result of the fetch operation)
6. When the resource can't be validated with the provided schema then an error containing the resulting resource and the validation report is returned to the user
7. When the resource is not valid for any other reason (the resource payload is invalid, the provided schema reference does not exist), an appropriate message is returned to the user

**API changes**
An example of requests would be:
***With no validation***
```
POST /v1/dry-run/resources/{org}/{project}
 {
   "resource": {
    "@context": [
      "https://bbp.neuroshapes.org"
    ],
    "@id": "https://bbp.epfl.ch/nexus/v1/realms/bbp/users/bob",
    "@type": [
      "Person",
      "Agent"
    ],
    "email": "bob@epfl.ch",
    "familyName": "Doe",
    "givenName": "Bob",
  }
}
```
***With validation with a new schema***
```
POST /v1/dry-run/resources/{org}/{project}
  "schema": {
    "@type": "ExistingSchema",
    "@id": "nxv:mySchema"
  },
  "resource": {
    "@context": [
      "https://bbp.neuroshapes.org"
    ],
    "@id": "https://bbp.epfl.ch/nexus/v1/realms/bbp/users/bob",
    "@type": [
      "Person",
      "Agent"
    ],
    "email": "bob@epfl.ch",
    "familyName": "Doe",
    "givenName": "Bob",
  }
}
```
***With validation with a new schema***
```
POST /v1/dry-run/resources/{org}/{project}
  "schema": {
    "@type": "NewSchema",
    "value": { ... } # SHACL schema
  },
  "resource": {
    "@context": [
      "https://bbp.neuroshapes.org"
    ],
    "@id": "https://bbp.epfl.ch/nexus/v1/realms/bbp/users/bob",
[...]

PR #4259, opened 2023-09-06
Introducing ResourcesPractice to generate resources without persisting them

imsdu:
Related to #4102


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -106,23 +105,6 @@ trait Resources {
...
-    *
-    * @param id
-    *   the identifier that will be expanded to the Iri of the resource
-    * @param projectRef
-    *   the project reference where the resource belongs
-    * @param schemaOpt
-    *   the optional identifier that will be expanded to the schema reference to validate the resource. A None value
-    *   uses the currently available resource schema reference.
-    */
-  def validate(
```
imsdu, line 120: Moved to `ResourcesPractice`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ValidateResource.scala
```diff
@@ -69,38 +62,6 @@ trait ValidateResource {
 }
 
 object ValidateResource {
-  sealed trait ValidationResult {
```
imsdu, line 72: Extracted to its own file

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesPractice.scala
```diff
@@ -0,0 +1,151 @@
...
+
+    override def generate(project: ProjectRef, schema: ResourceF[Schema], source: NexusSource)(implicit
+        caller: Caller
+    ): IO[ResourceRejection, DataResource] =
+      for {
+        projectContext <- fetchContext.onRead(project)
+        jsonld         <- sourceParser(project, projectContext, source.value)
+        validation     <- validateResource(jsonld.iri, jsonld.expanded, schema)
+        res            <- toResourceF(project, jsonld, source, validation)
+      } yield res
```
olivergrabinski, line 101: It feels like this could be simplified

imsdu: I agree, we have to refactor this along with `ResourcesImpl` but I would postpone that after the migration to Cats-Effects as it will likely affect how we fetch project context for example.
Some of it may also be included in the evaluate method in the case of `ResourcesImpl`, it was done this way because with `akka-cluster`/`akka-persistence`, it was spawning an actor and we needed to define its id before the evaluation

olivergrabinski (approved):

PR #4263, opened 2023-09-11
Reduce log level when querying aggregate views

imsdu:
Also:
- deleting some dead code
- use log4cats


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/views/ViewsStore.scala
```diff
@@ -26,14 +22,6 @@ trait ViewsStore[Rejection] {
     */
   def fetch(id: IdSegmentRef, project: ProjectRef): IO[Rejection, View]
 
-  /**
-    * Fetch default views and combine them in an aggregate view
-    * @param scope
-    *   to get all default view from the system / a given organization / a given project
-    * @return
-    */
-  def fetchDefaultViews(scope: Scope): UIO[AggregateView]
```
imsdu, line 35: Dead code is dead

shinyhappydan (approved):

PR #4274, opened 2023-09-14
Add resource trial routes

### LINKED ISSUE #4102: Allow users to perform dry-runs on resources (with optional validation with a schema)

imsdu:
Fixes #4102 
image


docs/src/main/paradox/docs/delta/api/practice.md
```diff
@@ -0,0 +1,51 @@
...
+
+* `{schema}`: String/Json: The schema to validate the provided resource. If a string is provided, it will attempt to resolve it as an existing schema.
+If a json payload is provided, it will attempt to generate the schema and then use the result to validate the resource.
+This field is optional and defaults to no SHACL validation.
+* `{resource}`: Json: The resource payload to test and validate
+
+The Json response will contain:
+* The generated resource in the compacted JSON-LD format if the generation and the validation was successful
+* The generated schema if a new schema payload was provided
+* The error if the one of the steps fails (invalid resource/invalid new schema/existing schema not found/...)
```
olivergrabinski, line 40: This is not rendered correctly (just the empty line missing)

docs/src/main/paradox/docs/delta/api/practice.md
```diff
@@ -0,0 +1,51 @@
+# Practice
```
olivergrabinski, line 1: Is it worth linking to the validation endpoint? Or it's better not to mix them here yet?

docs/src/main/paradox/docs/delta/api/practice.md
```diff
@@ -0,0 +1,51 @@
...
+Practice operations contains read-only operations designed to help users compose and validate their
+resources before effectively save them in Nexus.
```
olivergrabinski, line 3: `contain`, `saving`

olivergrabinski (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesPracticeRoutes.scala
```diff
@@ -0,0 +1,162 @@
...
+      baseUri: BaseUri,
+      s: Scheduler,
+      cr: RemoteContextResolution,
+      ordering: JsonKeyOrdering,
+      decodingOption: DecodingOption
+  ): ResourcesPracticeRoutes =
+    new ResourcesPracticeRoutes(
+      identities,
+      aclCheck,
+      (project, source, caller) => schemas.createDryRun(project, source)(caller).toBIO[SchemaRejection],
```
olivergrabinski, line 158: Having this as a private method in `ResourcesPracticeRoutes` would allow getting rid of the constructor all together? Not super important

shinyhappydan (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesPracticeRoutes.scala
```diff
@@ -0,0 +1,162 @@
...
+import io.circe.generic.extras.semiauto.deriveConfiguredDecoder
+import io.circe.{Decoder, Json}
+import monix.bio.IO
+import monix.execution.Scheduler
+
+import scala.annotation.nowarn
+
+/**
+  * The resource practice routes allowing to do read-only operations on resource
+  */
```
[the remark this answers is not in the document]
imsdu: They are not, the two operations do not go through the state machine because they don't result in an event and a new state to be persisted in the primary store

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesPracticeRoutes.scala
```diff
@@ -0,0 +1,162 @@
...
+      concat(validateRoute, practiceRoute)
+    }
+
+  private def validateRoute: Route =
+    pathPrefix("resources") {
+      extractCaller { implicit caller =>
+        resolveProjectRef.apply { ref =>
+          authorizeFor(ref, Write).apply {
+            (get & idSegment & idSegmentRef & pathPrefix("validate") & pathEndOrSingleSlash) { (schema, id) =>
+              val schemaOpt = underscoreToOption(schema)
```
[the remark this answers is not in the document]
imsdu: Ok to create a ticket and address it in another PR ?
This happens in a couple of places and this one is already complex enough

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesPracticeRoutes.scala
```diff
@@ -0,0 +1,162 @@
...
+    input.schema match {
+      case ExistingSchema(schemaId) =>
+        emit(resourcesPractice.generate(project, schemaId, input.resource).flatMap(_.asJson))
+      case NewSchema(schemaSource)  =>
+        emit(
+          generateSchema(project, schemaSource, caller).flatMap { schema =>
+            resourcesPractice.generate(project, schema, input.resource).flatMap(_.asJson)
+          }
+        )
+    }
```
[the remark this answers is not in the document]
imsdu: Those are 10 lines of code here without complexity where:
* Most of it is related to how we handle the input
* ResourcesPractis(c)e is about generating resources, not schemas

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesPracticeRoutes.scala
imsdu: The idea is to make unit-testing easier because `Schemas` comes with a lot of dependencies (doobie, SHACL engine, ...)

shinyhappydan: Using functions like this makes general navigation of the code harder, as well as this constructor being unnecessary. It should already be easy to use mock implementations in a unit test

Interface segregation is a good way of handling this sort of situation and will mean you won't need the dependencies you mention

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesPracticeRoutes.scala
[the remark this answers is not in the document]
shinyhappydan: Maybe there is another way to phrase it, but I think this comment is misleading

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesPracticeRoutes.scala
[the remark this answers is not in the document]
shinyhappydan: Sure

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesPracticeRoutes.scala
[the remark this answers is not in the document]
shinyhappydan: I think this function is unrelated to parsing or response-writing so it should be in a service rather than routes, but that's just my opinion I guess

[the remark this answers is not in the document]
imsdu: > I think this function is unrelated to parsing or response-writing so it should be in a service rather than routes, but that's just my opinion I guess

It is related to parsing as `GenerationInput` only lives in the scope of the routes and reflects the payload submitted to the user.

> Any time the user calls an API they are ostensibly performing a single action. The user is saying they want to generate using a generated schema. Perhaps that affects they way you design the ResourcesPractise, but it shouldn't really be a concern for the routes
We are talking here about a pattern matching and a flatMap operation.
I prefer this solution than changing `ResourcePractise` or including yet another layer

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesPracticeRoutes.scala
imsdu: I am not sure we want to do a remake between OOP and FP in the comments here.
As long as we have the `apply` method here, I don't think it hurts navigability.
The FP way allows better expressivity and better flexibility than introducing a multitude of interfaces.
On the flexibility side, what happens if in another case, you want to generate a schema but only care about one of the fields or just if it succeeds ? Or something else ?
You would have to introduce new interfaces or have your mocks to know about the result before dropping the info.
(Those are just simple examples, there could be others)

shinyhappydan: Every time I want to see what code actually run when we are passing one of these functions, I have to navigate multiple steps and potentially multiple branches. That is definitely harder than just clicking on show implementations

In this instance, using a function instead of an interface gives us nothing and costs a lot, which is why I'm saying we should change it here

shinyhappydan: Also I don't believe there is a conflict between OOP and FP here, I'm not sure how using an interface makes something not functional?

dantb: Interesting discussion! In these kinds of situations we've found single-method traits to be quite a good compromise in the past. They essentially act as just named functions, sort of like a type alias but first-class. Navigation is still easy and Scala's de-sugaring makes them simple to stub. 

For example if a dependent expects a `trait MyTrait { def foo(i: Int): String }`, you can pass it as a stub like `(i: Int) -> "hello"` in tests and the compiler will understand it needs to do the whole `new MyTrait { ... }` bit for you, which is nice. In Scala 3 there's even less boilerplate needed (better type inference).

Not sure I understand this bit `you want to generate a schema but only care about one of the fields or just if it succeeds ?` - what would that look like? (Just trying figure out what I might do in that situation)

As an aside, in Haskell there are a number of approaches to this type of dependency injection. You might do a type alias, newtype wrapper round the function, or a type class (if doing tagless final). Otherwise type signatures become very hard to read when containing lots of functions passed as arguments. The single-method trait isn't that different from the type class pattern as I see it. (Well really all of these approaches are pretty similar 😛 )

The "method dictionary" pattern can be quite neat too - you basically have a bunch of newtype wrapped functions stored in some `Env` data type, and access them by the record accessor (just the thing inside the newtype).  For convenience you might also have syntax to pull out the functions and run them in one fell swoop. In this case that may require quite a big refactor though so I'm not sure.

shinyhappydan: The only downside with SAM (the example `(i: Int) -> "hello"` from above) is that Intellij doesn't really understand that it's an implementation of the interface from a code-navigation perspective. I suppose mostly we would only use SAM in tests, where it isn't so necessary to have the navigation functionality

In production code it's just nice to easily see what the dependency actually does

imsdu: I would love to be able to use a SAM here and we should use them where we can.
The Schemas module is complex and contains a collection of operations though (with like 3 operations to create a schema) and the part which is mentioned is only used in this context (so far).
As it is a matter that goes further than the context of this PR, we should rather discuss that when we are all in the office

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -183,13 +178,20 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap {
     "fail to create a resource that does not validate against a schema" in {
       Put(
         "/v1/resources/myorg/myproject/nxv:myschema/wrong",
-        payload.removeKeys(keywords.id).replaceKeyWithValue("number", "wrong").toEntity
+        payloadWithoutId.replaceKeyWithValue("number", "wrong").toEntity
```
shinyhappydan, line 181: After looking at this for a while I can see now that if you set the number field to be a string, presumably that breaks validation. It's not very obvious though, because there is nothing to suggest that is what the schema checks. With this and other tests, it's possible to make it clearer

imsdu: I tried to make it a bit clearer, note that this was some existing test code, that I only refactored a bit

crisely09:
does this mean that we can perform dry-runs now?

PR #4279, opened 2023-09-18
Enable compression for outgoing requests

### LINKED ISSUE #4249: Allow to compress payloads from outgoing requests from the Http client
**Motivation**
Delta interacts via other components via http and so far, it only set the `Accept-Encoding` header so that those components return compressed responses.

Compressing requests would also help especially during indexing processes where a lot of data is pushed to the secondary stores.

**Acceptance criteria**
1. With configuration it is now possible to compress requests to those components
2. This configuration option has to be different from the one for responses as not all components accept compressed requests (looking at you Blazegraph)
3. The http clients where it can be enabled are: `ElasticsearchClient`, `DeltaClient`

imsdu:
Fixes #4249 

I reused the compression flag instead of introducing a new one as:
* It was always enabled
* The `Accept-Encoding` header was correctly set to `gzip` and `deflate`

Allows to enable compression for outgoing requests:
* Only enabled for Elasticsearch by default (Blazegraph does not support it and there is an error with Keycloak where the payloads are always minimal anyway)

### DIFF
--- a/delta/app/src/main/resources/app.conf
+++ b/delta/app/src/main/resources/app.conf
@@ -310,7 +310,7 @@ app {
   }
 
   defaults {
-    http-client {
+    http-client-compression {
       # the retry strategy for the http client
       retry = ${app.defaults.constant-retry-strategy}
       # the strategy to decide if it is worth retrying when an Http error occurs.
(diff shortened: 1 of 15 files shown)


olivergrabinski (changes requested):
Does this warrant a note in the release notes?

olivergrabinski (approved):

PR #4285, opened 2023-09-19
Replace uses of 'archives/write' permission with 'resources/read'

dantb:
Think this change depends on changing the permissions in forge ([these places](https://github.com/search?q=repo%3ABlueBrain%2Fnexus-web+archives%2Fwrite&type=code)) and I guess potentially other clients? Unless we know for sure that every user with `archives/write` also has `resources/read`.

Also not sure what to unit test here - is this alright? The archive `model.permissions` could be deleted and the route could point directly to the resource permissions.

### DIFF
--- a/delta/app/src/main/resources/app.conf
+++ b/delta/app/src/main/resources/app.conf
@@ -112,7 +112,6 @@ app {
       "schemas/write",
       "files/write",
       "storages/write",
-      "archives/write",
       "version/read",
       "quotas/read",
       "supervision/read"
@@ -139,7 +138,6 @@ app {
       "schemas/write",
       "files/write",
       "storages/write",
-      "archives/write",
       "version/read",
       "quotas/read"
     ]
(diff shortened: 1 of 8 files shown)


imsdu:
> Think this change depends on changing the permissions in forge ([these places](https://github.com/search?q=repo%3ABlueBrain%2Fnexus-web+archives%2Fwrite&type=code)) and I guess potentially other clients? Unless we know for sure that every user with archives/write also has resources/read.
Forge does not rely on the archives api as far as I know, Fusion and eventually SBO does.

> Also not sure what to unit test here - is this alright? The archive model.permissions could be deleted and the route could point directly to the resource permissions.
I think it is ok like this, there is a similar approach in other modules (`schemas/read`, is actually `resources/read` and so on)

imsdu:
It looks good to me, you just need to add a section about it in the release notes for 1.9.
(The v1.9-release-notes.md file)

dantb:
> It looks good to me, you just need to add a section about it in the release notes for 1.9. (The v1.9-release-notes.md file)

Done 👍

imsdu (approved):

dantb:
Sorry @imsdu, need another ✅  after rebase

Issue #4050, opened 2023-07-07
Remove support for tars in archives


imsdu:
**Motivation**
Tar support has been deprecated in 1.8 with the introduction of zip as an option to download them

**Acceptance criteria**
1. The main/test code is removed
2. The documentation is also updated

PR #4286, opened 2023-09-20
Remove Tar archiving

dantb:
#4050 

- Removed `ArchiveFormat` abstraction
- Instead use `Zip` directly. The zip-specific stuff is still grouped together in an object - this might be more readable and I suppose it saves some duplication at callsites
- Removed metadata type parameter and hardcode to Akka's `ArchiveMetadata` for zips
- Remove tar-ey tests and make `ArchiveDownloadSpec` specific to zips (no longer an abstract class)


delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/model/Zip.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  def fileExtension: String = "zip"
+
+  def metadata(filename: String): ArchiveMetadata =
+    ArchiveMetadata.create(filename)
+
+  def writeFlow: WriteFlow[ArchiveMetadata] = Archive.zip()
+
+  def ordering: Ordering[ArchiveMetadata] = Ordering.by(md => md.filePath)
+}
```
shinyhappydan, line 33: I would probably go a step further and inline these, I think they're only used in one place?

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/routes/ArchiveRoutes.scala
```diff
@@ -96,9 +97,9 @@ class ArchiveRoutes(
...
 
-  private def sourceToFileResponse(source: AkkaSource, format: ArchiveFormat[_]): FileResponse =
-    FileResponse(s"archive.${format.fileExtension}", format.contentType, 0L, source)
+  private def sourceToFileResponse(source: AkkaSource): FileResponse =
+    FileResponse(s"archive.${Zip.fileExtension}", Zip.contentType, 0L, source)
 
-  private def archiveResponse: Directive1[Option[ArchiveFormat[_]]] =
-    extractRequest.map(ArchiveFormat(_))
+  private def archiveResponse: Directive1[Option[Zip.type]] =
+    extractRequest.map(Zip(_))
```
shinyhappydan, line 104: I think this should be a boolean if we're not using the `Some` ?

delta/plugins/archive/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/ArchiveDownloadSpec.scala
```diff
@@ -147,20 +146,20 @@ abstract class ArchiveDownloadSpec
...
 
     def rejectedAccess(value: ArchiveValue) = {
       archiveDownload
-        .apply(value, project.ref, format, ignoreNotFound = true)(Caller.Anonymous, global)
+        .apply(value, project.ref, ignoreNotFound = true)(Caller.Anonymous, global)
         .rejectedWith[AuthorizationFailed]
     }
 
-    s"provide a ${format.fileExtension} for both resources and files" in {
+    s"provide a ${Zip.fileExtension} for both resources and files" in {
```
shinyhappydan, line 162: It might be nice to not use a variable here any more, I think it can cause trouble with tooling

shinyhappydan (approved):

delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/routes/ArchiveRoutes.scala
dantb: Ah yeah definitely

dantb:
@shinyhappydan I'm getting a failing test which doesn't fail locally 🤔  it fails on a future which populates a file from a stream

imsdu (approved):

imsdu (approved):

PR #4290, opened 2023-09-22
Also allows post for the multi-fetch operation

imsdu:


delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/MultiFetchRoutesSpec.scala
```diff
@@ -89,9 +89,9 @@ class MultiFetchRoutesSpec extends BaseRouteSpec {
       }
     }
 
-    "return expected results as annotated source for a user with limited access" in {
+    "return expected results as annotated source for a user with limited access using post method" in {
       val entity = request(ResourceRepresentation.AnnotatedSourceJson).toEntity
-      Get(endpoint, entity) ~> asAlice ~> routes ~> check {
+      Post(endpoint, entity) ~> asAlice ~> routes ~> check {
```
olivergrabinski, line 94: If we allow both get and post here shouldn't both be tested?

imsdu: I planned to push this "hotfix" first so as not to unblock Dinika and then complete the tests.
But I can add them now

olivergrabinski (approved):

PR #4298, opened 2023-09-26
Add organization deletion route

dantb:
Key implementation points:
- Added `prune` parameter to `DELETE /orgs/{label}` defaulting to false and requiring a new `organizations/delete` permission. Previously only deprecation was allowed for a specific `rev`. Not sure how great it is to ignore `rev` even if specified, maybe slightly confusing for the client? If the org is empty, return 409 Conflict.
- `OrganizationDeleter` has all the logic. If the org is non-empty, we do these steps
	1. Drop the org partitions for scoped tables
	2. Delete the org records from the global tables
	3. Delete the ACL records from the global tables

I used this ticket to try to understand some of the patterns in the codebase so I have a few questions / comments, some specific to this and some more general:
1. For the ACLs I manually wrote the deletion queries rather than reusing `Acls`, is this ok? The main reason was so it could all be in one transaction, since this feels like a bad thing to get stuck in a half-way state. It would also add a bigger dependency.
2. What places do I need to update with the new permission? Is it in infra somewhere?
3. Spent far too long trying to make the integration tests sane 😅  I wasn't sure the best approach and ended up using the old school arrange / act / assert. Preferably the tests should be independent so the tables get truncated before each. I don't really like the dodgy Monix type signatures though...  Guess it comes down to trying to wrangle errors into a single type - I just didn't bother and opted for `Any` grimness. Is there a better way?
4. The unit tests do a lot of global state mutating and depend on each other a lot (as we've discussed before). It's a bit like a mishmash of pure FP and side effects which makes it quite difficult to reason about. I just threw more crap on the pile for this change. It would be nice to try to improve them at some point though - are there any plans for that? I could imagine tests like these being quite nice using CE, http4s, munit and scalacheck. Maybe that's just a personal preference though 🙂


dantb:
Failing test is just to do with how we respond when `rev` is missing, will fix tomorrow

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleter.scala
```diff
@@ -0,0 +1,65 @@
...
+    private val logger: Logger = Logger[OrganizationDeleter]
+
+    def delete(id: Label): IO[OrganizationNonEmpty, Unit] =
+      for {
+        orgIsEmpty <- orgIsEmpty(id)
+        _          <- if (orgIsEmpty) log(s"Deleting empty organization $id") *> deleteAll(id)
+                      else IO.raiseError(OrganizationNonEmpty(id))
+      } yield ()
+
+    private def log(msg: String): UIO[Unit] = UIO.delay(logger.info(msg))
```
imsdu, line 36: There is a logger relying on Log4Cats in the kernel module

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleter.scala
```diff
@@ -0,0 +1,65 @@
...
+
+    private def deleteAll(id: Label): UIO[Unit] =
+      (for {
+        _ <- List("scoped_events", "scoped_states").traverse(dropPartition(id, _)).void
+        _ <- List("global_events", "global_states").traverse(deleteGlobal(id, _))
+      } yield ()).transact(xas.write).void.hideErrors
+
+    private def dropPartition(id: Label, table: String): ConnectionIO[Unit] =
+      Update0(s"DROP TABLE IF EXISTS ${table}_${MD5.hash(id.value)}", None).run.void
```
imsdu, line 45: In `PartitionInit`, you should have the way to encode the org id into a partition name so as all this logic remains in one place

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleter.scala
```diff
@@ -0,0 +1,65 @@
...
+
+    private def deleteGlobal(id: Label, table: String): ConnectionIO[Unit] =
+      for {
+        _ <- deleteGlobalQuery(Organizations.encodeId(id), Organizations.entityType, table)
+        _ <- deleteGlobalQuery(Acls.encodeId(AclAddress.fromOrg(id)), Acls.entityType, table)
+      } yield ()
+
+    private def deleteGlobalQuery(id: IriOrBNode.Iri, tpe: EntityType, table: String): ConnectionIO[Unit] =
+      Update[(EntityType, IriOrBNode.Iri)](s"DELETE FROM $table WHERE" ++ " type = ? AND id = ?").run((tpe, id)).void
```
imsdu, line 54: You could use the sql interpolator here

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleterSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+import ch.epfl.bluebrain.nexus.testkit.IOFixedClock
+import ch.epfl.bluebrain.nexus.testkit.bio.BioSuite
+import doobie.implicits._
+import doobie.util.update.Update0
+import monix.bio.{IO, Task, UIO}
+import munit.AnyFixture
+
+import java.util.UUID
+
+class OrganizationDeleterSpec extends BioSuite with IOFixedClock with ConfigFixtures {
```
imsdu, line 26: We tend to call them *Suite

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleterSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+  private val config         = ProjectsConfig(eventLogConfig, pagination, cacheConfig, deletionConfig)
+  private val projectFixture = ProjectsFixture.init(fetchOrg, defaultApiMappings, config)
+
+  override def munitFixtures: Seq[AnyFixture[_]] = List(projectFixture)
+
+  private lazy val (xas, projects) = projectFixture()
+  private lazy val orgDeleter      = OrganizationDeleter(xas)
+  private val projRef              = ProjectRef.unsafe("org", "myproj")
+  private val fields               = ProjectFields(None, ApiMappings.empty, None, None)
+  private lazy val orgs            = GlobalEventLog(Organizations.definition, logConfig, xas)
```
imsdu, line 45: It feels like initializing an `OrganizationImpl` would not need much more work and would avoid to know about the underlying command and other things

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleterSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  implicit val subject: Subject = Identity.User("Bob", Label.unsafe("realm"))
+  implicit val uuidF: UUIDF     = UUIDF.fixed(orgUuid)
+
+  test("Fail when trying to delete a non-empty organization") {
+    val arrange: IO[Any, Unit] = for {
+      _ <- truncateTables
```
imsdu, line 52: Using two different orgs for the two tests could avoid to truncate the tables and may make things easier

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleterSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+  test("Successfully delete an empty organization") {
+    val arrange: IO[Any, Unit] = for {
+      _ <- truncateTables
+      _ <- orgs.evaluate(org, OrganizationCommand.CreateOrganization(org, None, Identity.Anonymous))
+    } yield ()
+
+    val act: UIO[Either[OrganizationNonEmpty, Unit]] =
+      orgDeleter.delete(org).attempt
+
+    def assert(result: Either[OrganizationNonEmpty, Unit]): IO[Any, Unit] = for {
```
imsdu, line 88: You should test for the acl somewhere

imsdu (changes requested):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/acls/Acls.scala
```diff
@@ -25,7 +25,7 @@ import java.time.Instant
 /**
   * Operations pertaining to managing Access Control Lists.
   */
-trait Acls {
+trait Acls extends PurgeAcl {
```
imsdu, line 28: It seems that this change is not used in the end

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleterSpec.scala
shinyhappydan: MUnit tests are *Suite, ScalaTests are *Spec

imsdu:
> For the ACLs I manually wrote the deletion queries rather than reusing Acls, is this ok? The main reason was so it could all be in one transaction, since this feels like a bad thing to get stuck in a half-way state. It would also add a bigger dependency.

I am ok with that

> What places do I need to update with the new permission? Is it in infra somewhere?

Nothing to do with the permission itself but we will need to assign it to a user to test it

> Spent far too long trying to make the integration tests sane 😅 I wasn't sure the best approach and ended up using the old school arrange / act / assert. Preferably the tests should be independent so the tables get truncated before each. I don't really like the dodgy Monix type signatures though... Guess it comes down to trying to wrangle errors into a single type - I just didn't bother and opted for Any grimness. Is there a better way?

Not sure we can do much better with this kind of PR but as this part is mostly independent from the rest, you may have opted for Cats-Effect in the end :) .

> The unit tests do a lot of global state mutating and depend on each other a lot (as we've discussed before). It's a bit like a mishmash of pure FP and side effects which makes it quite difficult to reason about. I just threw more crap on the pile for this change. It would be nice to try to improve them at some point though - are there any plans for that? I could imagine tests like these being quite nice using CE, http4s, munit and scalacheck. Maybe that's just a personal preference though

There is definitely room to improve things by making the code more modular, by improving the way we use MUnit (like setting up CE Resources and transform to MUnit fixtures the latest possible as those don't compose), by designing how to test the lifecycle of a resource without too much repetition and slowing them down too much...

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleter.scala
dantb: It's not possible when parameterising over the table name since substitutions only work for values. This is maybe a bit overkill but I was curious to see if it was possible 😅

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleterSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+      statePartitionDeleted <- queryPartitions("scoped_states").map(_.isEmpty)
+    } yield {
+      assertEquals(result, Right(()))
+      assertEquals(eventPartitionDeleted, true)
+      assertEquals(statePartitionDeleted, true)
+      assertEquals(globalStateDeleted, true)
+      assertEquals(globalEventsDeleted, true)
+    }
+
+    arrange >> act >>= assert
```
shinyhappydan, line 101: Maybe it's possible to rewrite this to be a little more descriptive? I think this test is saying

Given there is an organisation
When I delete that organisation
All data from that organisation is deleted

But's a little abstracted

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleterSpec.scala
dantb: 👍

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleterSpec.scala
dantb: ah yeah that was unnecessary

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/model/OrganizationRejection.scala
```diff
@@ -18,7 +18,7 @@ import io.circe.{Encoder, JsonObject}
   * @param reason
   *   a descriptive message as to why the rejection occurred
   */
-sealed abstract class OrganizationRejection(val reason: String) extends Product with Serializable
+sealed abstract class OrganizationRejection(val reason: String) extends Throwable with Product with Serializable
```
shinyhappydan, line 21: I think you need to pass the message to throwable?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/permissions/Permissions.scala
```diff
@@ -157,6 +157,8 @@ object Permissions {
     final val read: Permission   = Permission.unsafe("organizations/read")
     final val write: Permission  = Permission.unsafe("organizations/write")
     final val create: Permission = Permission.unsafe("organizations/create")
+    // TODO setup this permission
```
shinyhappydan, line 160: Is this to do?

shinyhappydan (approved):

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/OrganizationsRoutesSpec.scala
```diff
@@ -38,7 +47,11 @@ class OrganizationsRoutesSpec extends BaseRouteSpec with IOFromMap {
     aclChecker.append,
     Set(orgsPermissions.write, orgsPermissions.read)
   )
-  private lazy val orgs  = OrganizationsImpl(Set(aopd), config, xas)
+
+  private lazy val orgs                            = OrganizationsImpl(Set(aopd), config, xas)
+  private lazy val orgDeleter: OrganizationDeleter = id =>
+    if (id == org1.label) IO.raiseError(OrganizationNonEmpty(id))
```
imsdu, line 53: `IO.raiseWhen`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/OrganizationDeleter.scala
imsdu: `Fragment.const` could also help here ?

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/Transactors.scala
```diff
@@ -28,7 +29,10 @@ final case class Transactors(
     write: Transactor[Task],
     streaming: Transactor[Task],
     cache: PartitionsCache
-) {
+)(implicit s: Scheduler) {
+
+  def readCE: Transactor[IO]  = read.mapK(BIO.liftTo)
```
imsdu, line 34: Nice that something like this is possible

docs/src/main/paradox/docs/releases/v1.9-release-notes.md
```diff
@@ -123,6 +123,12 @@ Creating an archive now requires only the `resources/read` permission instead of
 
 Tarball archives are no longer supported due to unnecessary restrictions. ZIP is now the only allowed format and clients should send `application/zip` in the `Accept` header when creating archives.
 
+### Organizations
+
+#### Support deletion of empty organizations
```
imsdu, line 128: Link to the operation in the API ?

imsdu (approved):

imsdu (approved):

PR #4307, opened 2023-09-26
Update to Java 17

### LINKED ISSUE #3645: Update to JRE 17
Distage is relying on cglib to resolve cyclic dependencies and cglib does not work with JRE 16+

cglib is only imported by Distage

The idea is to test 17 adding this options (just an example, can be cleaned up):
`- JVM_OPTS= --add-exports=java.naming/com.sun.jndi.ldap=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.lang.invoke=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.security=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.management/javax.management=ALL-UNNAMED --add-opens=java.naming/javax.naming=ALL-UNNAMED`

cf. https://github.com/cglib/cglib/issues/191#issuecomment-1181253120

NB: 
* The latest versions of Distage rely on Byte Buddy but they require Cats-Effect 3
* Additional tasks about breaking those cyclic dependencies can be done too

olivergrabinski:
Closes #3645

### DIFF
--- a/.github/workflows/ci-delta-app.yml
+++ b/.github/workflows/ci-delta-app.yml
@@ -14,6 +14,13 @@ jobs:
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
+      - name: Setup JDK
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+          check-latest: true
+          cache: 'sbt'
       - name: Checkout
         uses: actions/checkout@v3
         with:
--- a/.github/workflows/ci-delta-app.yml
+++ b/.github/workflows/ci-delta-app.yml
@@ -20,7 +20,6 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
           check-latest: true
-          cache: 'sbt'
       - name: Checkout
         uses: actions/checkout@v3
         with:
--- a/delta/plugins/blazegraph/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryStoreSuite.scala
+++ b/delta/plugins/blazegraph/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryStoreSuite.scala
@@ -70,7 +70,7 @@ object BlazegraphSlowQueryStoreSuite {
     )
   }
 
-  private val Now          = Instant.now()
+  private val Now          = Instant.now().truncatedTo(ChronoUnit.MILLIS)
   private val OneWeekAgo   = Now.minus(Duration.ofDays(7))
   private val EightDaysAgo = Now.minus(Duration.ofDays(8))
   private val RecentQuery  = queryAtTime(Now)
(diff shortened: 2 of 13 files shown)


delta/plugins/blazegraph/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/slowqueries/BlazegraphSlowQueryStoreSuite.scala
```diff
@@ -31,7 +32,7 @@ class BlazegraphSlowQueryStoreSuite
       SparqlQuery(""),
       failed = true,
       1.second,
-      Instant.now(),
+      Instant.now().truncatedTo(ChronoUnit.MILLIS),
```
olivergrabinski, line 35: > [The clock used by `Instant.now()`] is based on the best available system clock. This may use [System.currentTimeMillis()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#currentTimeMillis--), or a higher resolution clock if one is available.

shinyhappydan: I think in production code we are safe, as we should never be calling `Instant.now()` directly. In tests it's more annoying to have to use a clock.

One possible solution is to be a bit more relaxed in tests. I know that we use a fixed clock and check exact timestamps, but an alternative approach is to use a normal clock be a little more flexible. That is probably too big a change though

Perhaps there is some sort of typeclass equality that could rescue us..?

.github/workflows/ci-delta-app.yml
```diff
@@ -14,6 +14,12 @@ jobs:
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
+      - name: Setup JDK
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+          check-latest: true
```
shinyhappydan, line 22: I don't think you necessarily have to do it in this PR, but here is something interesting: https://github.com/marketplace/actions/setup-java-jdk#caching-sbt-dependencies

shinyhappydan (approved):

.github/workflows/ci-delta-app.yml
olivergrabinski: I have tried it but for some reason it takes ages to recover the cache. Example: https://github.com/BlueBrain/nexus/actions/runs/6318630985/job/17158027837 It takes 2m45 to fetch the cache but doesn't speed up the actual build/tests 🤔

olivergrabinski: Btw there is caching added to the release and snapshot pipelines to try it out. My suspicion is that it should work ok on the cloud runners. If it is also slow I will remove it.

imsdu (approved):

PR #4310, opened 2023-09-27
Fix fetching resources after trial routes introduction

### LINKED ISSUE #4309: Fetching resources fail for users with only resources/read permission
**Describe the bug**
Fetching resources fail for users with only resources/read permission and get a forbidden response

- [ ] Fusion
- [X] Delta
- [ ] Documentation

imsdu:
Fixes #4309


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesTrialRoutes.scala
```diff
@@ -63,8 +63,8 @@ final class ResourcesTrialRoutes(
     pathPrefix("resources") {
       extractCaller { implicit caller =>
         resolveProjectRef.apply { project =>
-          authorizeFor(project, Write).apply {
```
imsdu, line 66: Where it was placed made that in the end only users with `resources/write` could access the resources

shinyhappydan (approved):

PR #4318, opened 2023-09-30
Migrate auth token code to cats effect

shinyhappydan:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
```diff
@@ -1,36 +1,38 @@
...
-import ch.epfl.bluebrain.nexus.delta.kernel.cache.KeyValueStore
+import ch.epfl.bluebrain.nexus.delta.kernel.cache.LocalCache
 import ch.epfl.bluebrain.nexus.delta.kernel.effect.migration.MigrateEffectSyntax
-import ch.epfl.bluebrain.nexus.delta.kernel.utils.IOUtils
+import ch.epfl.bluebrain.nexus.delta.kernel.utils.IOInstant
 import ch.epfl.bluebrain.nexus.delta.sdk.auth.Credentials.ClientCredentials
 import ch.epfl.bluebrain.nexus.delta.sdk.identities.ParsedToken
 import ch.epfl.bluebrain.nexus.delta.sdk.identities.model.AuthToken
-import monix.bio.UIO
+import monix.bio
```
olivergrabinski, line 11: Can we get rid of monix here?

dantb (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/auth/AuthTokenProvider.scala
shinyhappydan: Ah, I should have commented this: distage `fromEffect` does not seem to work with cats effect right now. I didn't want to investigate, since this was just something I was doing for myself and I didn't want to waste too much time on it

imsdu: It is configured to expected a Monix IO for now and I don't think we can mix both

PR #4320, opened 2023-10-02
Add tags to indices for resources, files, schemas

### LINKED ISSUE #943: Add tag metadata to the indices
The tag metadata is not added to `elasticsearch` and `sparql` indices.

We should only introduce it for resources/schemas and files

dantb:
Fixes #943 

I checked that the mapping is as expected my local ES.

Unsure if / where I need to add tests? `ListingsSpec` hits the metadata field (hence the change)

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/File.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/File.scala
@@ -1,9 +1,12 @@
 package ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model
 
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.files.Files
+import ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model.File.Metadata
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.StoragesConfig.StorageTypeConfig
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.StorageType
 import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.contexts
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.ContextValue
 import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.JsonLdContext.keywords
 import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.encoder.JsonLdEncoder
 import ch.epfl.bluebrain.nexus.delta.sdk.ResourceShift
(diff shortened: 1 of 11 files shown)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/File.scala
```diff
@@ -37,32 +41,44 @@ final case class File(
     storageType: StorageType,
     attributes: FileAttributes,
     tags: Tags
-)
+) {
+  def metadata(rev: Int): Metadata = Metadata(tags.byRevision(rev))
```
imsdu, line 45: We want all tags

imsdu (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/File.scala
```diff
@@ -37,32 +41,42 @@ final case class File(
...
-      val storageJson                       = Json.obj(
-        keywords.id  -> file.storage.iri.asJson,
-        keywords.tpe -> storageType.iri.asJson,
-        "_rev"       -> file.storage.rev.asJson
-      )
-      file.attributes.asJsonObject.add("_storage", storageJson)
-    }
+  final case class Metadata(tags: List[UserTag])
+
+  implicit def fileEncoder(implicit config: StorageTypeConfig): Encoder[File] = { file =>
```
dantb, line 52: I noticed we're doing `Encoder.AsObject` all over the place - I'm not sure this is necessary? The `Json` ADT provides functions to work with objects (`mapObject` etc). It's designed so we operate on the abstraction - I think this adds unnecessary boilerplate?

imsdu: It does look nicer this way

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/schemas/model/Schema.scala
```diff
@@ -88,14 +94,21 @@ object Schema {
         value.source.topContextValueOrEmpty.merge(ContextValue(contexts.shacl))
     }
 
-  type Shift = ResourceShift[SchemaState, Schema, Nothing]
+  implicit private val fileMetadataEncoder: Encoder[Metadata] = { m =>
+    Json.obj("_tags" -> m.tags.asJson)
+  }
+
+  implicit val fileMetadataJsonLdEncoder: JsonLdEncoder[Metadata] =
+    JsonLdEncoder.computeFromCirce(ContextValue(contexts.schemasMetadata))
```
imsdu, line 102: `contexts.metadata`

imsdu (approved):

PR #4337, opened 2023-10-05
Add `Vary` header for resource/files fetch operations

### LINKED ISSUE #4299: Introduce the Vary header when fetching a resource by rev/tag
**Motivation**
See #4295
```
Complex applications (e.g. SBO) in some cases require fetching a lot of interconnected data from KG
either by using SPARQL queries or sets of subsequent calls to ES/resource/file endpoints. 
Such operations might be slow which leads to degraded user experience.

Describe the solution you'd like.
Web clients these days support Cache API which, in the context of SBO, can be efficiently used 
to cache resources/files with fixed revisions, but requires VARY header (the spec, some pretty 
good article) to be present on the HTTP responses in order for 
key matching algorithm to function properly.

I think that a minimal solution would be to have a Vary: Accept header for Resources/Files 
fetch endpoints, so that clients can discriminate between requests for a file or its metadata, 
a resource or it's Web representation/redirect (when using Accept: text/html request header).

This will not only enable efficient caching on the client side, improving the UX, 
but as a side-effect will lower the amount of requests to KG.
```

**Acceptance criteria**
1. When a user fetches a resource by rev or by tag then a vary header is introduced with the response with the `Accept`, `Accept-Encoding` and `Cache-Control` headers provided by the client
2. This list of header can be configurable so that other headers may be easily included in the future
3. The fetch original payload follow the same rules

### LINKED ISSUE #4300: Introduce the Vary header when fetching a file by rev/tag
**Motivation**
See #4295

**Acceptance criteria**
1. When a user fetches a file by rev or by tag then a vary header is introduced with the response with the Accept, Accept-Encoding and Cache-Control headers provided by the client
2. The Vary header is returned when return file metadata and the file itself

olivergrabinski:
Closes #4300
Closes #4299

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutes.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutes.scala
@@ -184,12 +184,16 @@ final class FilesRoutes(
     }
 
   def fetch(id: IdSegmentRef, ref: ProjectRef)(implicit caller: Caller): Route =
-    headerValueByType(Accept) {
+    (headerValueByType(Accept) & varyAcceptHeaders) {
       case accept if accept.mediaRanges.exists(metadataMediaRanges.contains) =>
-        emit(fetchMetadata(id, ref).rejectOn[FileNotFound])
+        emit(fetchMetadata(id, ref).rejectWhen(unauthorizedOrNotFound))
       case _                                                                 =>
-        emit(files.fetchContent(id, ref).rejectOn[FileNotFound])
+        emit(files.fetchContent(id, ref).rejectWhen(unauthorizedOrNotFound))
     }
+
+  private val unauthorizedOrNotFound: PartialFunction[FileRejection, Boolean] = {
+    case _: AuthorizationFailed | _: FileNotFound => true
+  }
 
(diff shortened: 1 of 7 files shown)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutes.scala
```diff
@@ -184,13 +184,17 @@ final class FilesRoutes(
...
-        emit(fetchMetadata(id, ref).rejectOn[FileNotFound])
+        emit(fetchMetadata(id, ref).rejectWhen(unauthorizedOrNotFound))
       case _                                                                 =>
-        emit(files.fetchContent(id, ref).rejectOn[FileNotFound])
+        emit(files.fetchContent(id, ref).rejectWhen(unauthorizedOrNotFound))
     }
 
+  private val unauthorizedOrNotFound: PartialFunction[FileRejection, Boolean] = {
+    case _: AuthorizationFailed | _: FileNotFound => true
+  }
```
olivergrabinski, line 196: It's necessary to do this as in the ResourcesRoutes. Without the rejection on AuthorizationFailed it will inject the Vary header into the errors.

olivergrabinski: That's better yes

imsdu (approved):

PR #4339, opened 2023-10-06
Allow creation of a tagged resource

### LINKED ISSUE #4301: Allow to tag a resource at creation
**Motivation**
In SBO, there is a need to tag a resource directly at creation so the ability to provide an optional tag would allow to reduce the number of calls

**Acceptance criteria**
1. When a user provides an optional tag during resource creation
2. The resource is both created and tagged.

dantb:
Fixes #4301 

Few questions
1. Normally tagging bumps the revision but I assume it shouldn't here?
2. Since the tag is metadata like rev I figured it could also be a query param? 
3. Regarding the route response, I noticed tags don't appear there? So there's not much extra to test

### DIFF
--- a/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
+++ b/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
@@ -181,7 +181,6 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap {
           response.asJson shouldEqual
             resourceMetadata(projectRef, id, schema, (nxv + "Custom").toString, createdBy = alice, updatedBy = alice)
         }
-
       }
     }
 
@@ -197,7 +196,6 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap {
           response.asJson shouldEqual
             resourceMetadata(projectRef, id, schema, (nxv + "Custom").toString, createdBy = alice, updatedBy = alice)
         }
-
       }
     }
 
--- a/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
+++ b/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
@@ -17,7 +17,7 @@ import ch.epfl.bluebrain.nexus.delta.sdk.generators.{ProjectGen, ResourceResolut
 import ch.epfl.bluebrain.nexus.delta.sdk.identities.IdentitiesDummy
 import ch.epfl.bluebrain.nexus.delta.sdk.identities.model.Caller
 import ch.epfl.bluebrain.nexus.delta.sdk.implicits._
-import ch.epfl.bluebrain.nexus.delta.sdk.model.{IdSegmentRef, ResourceUris, Tags}
+import ch.epfl.bluebrain.nexus.delta.sdk.model.{IdSegmentRef, ResourceUris}
 import ch.epfl.bluebrain.nexus.delta.sdk.permissions.Permissions.{events, resources}
 import ch.epfl.bluebrain.nexus.delta.sdk.projects.FetchContextDummy
 import ch.epfl.bluebrain.nexus.delta.sdk.projects.model.ApiMappings
@@ -175,7 +175,7 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap {
         Post(endpoint, payload.toEntity) ~> routes ~> check {
           status shouldEqual StatusCodes.Created
           response.asJson shouldEqual resourceMetadata(projectRef, id, schema, (nxv + "Custom").toString)
-          resourceTags(resources, id).contains(tag) shouldBe true
+          lookupResourceByTag(resources, id) should contain(tag)
         }
       }
     }
@@ -207,13 +207,13 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap {
           status shouldEqual StatusCodes.Created
           response.asJson shouldEqual
             resourceMetadata(projectRef, id, schema, (nxv + "Custom").toString, createdBy = alice, updatedBy = alice)
-          resourceTags(resources, id).contains(tag) shouldBe true
+          lookupResourceByTag(resources, id) should contain(tag)
         }
       }
     }
 
-    def resourceTags(resources: Resources, id: Iri): Tags =
-      resources.fetch(IdSegmentRef(id, tag), projectRef, None).accepted.value.tags
+    def lookupResourceByTag(resources: Resources, id: Iri): List[UserTag] =
+      resources.fetch(IdSegmentRef(id, tag), projectRef, None).accepted.value.tags.value.keys.toList
 
     "reject the creation of a resource which already exists" in {
       Put("/v1/resources/myorg/myproject/_/myid", payload.toEntity) ~> routes ~> check {
--- a/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
+++ b/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
@@ -67,6 +67,8 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap {
   private val myId3                       = nxv + "myid3" // Resource created against no schema with id passed and present on the payload
   private val myId4                       = nxv + "myid4" // Resource created against schema1 with id passed and present on the payload
   private val myId5                       = nxv + "myid5" // Resource created against schema1 with id passed and present on the payload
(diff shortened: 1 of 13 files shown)


delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -63,11 +63,17 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap {
...
+  private val myId2 = nxv + "myid2" // Resource created against schema1 with id present on the payload
+  private val myId3 = nxv + "myid3" // Resource created against no schema with id passed and present on the payload
+  private val myId4 = nxv + "myid4" // Resource created against schema1 with id passed and present on the payload
+  private val myId5 = nxv + "myid5" // Resource created against schema1 with id passed and present on the payload
+  private val myId6 = nxv + "myid6" // Resource created and tagged against no schema with id present on the payload
+  private val myId7 = nxv + "myid7" // Resource created and tagged against no schema with id present on the payload
+  private val myId8 =
+    nxv + "myid8" // Resource created and tagged against no schema with id passed and present on the payload
+  private val myId9 =
+    nxv + "myid9" // Resource created and tagged against schema1 with id passed and present on the payload
```
dantb, line 76: This is kinda gross, I'd like to refactor it at some point to use generated data

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/model/ResourceEvent.scala
```diff
@@ -336,13 +339,15 @@ object ResourceEvent {
...
+        dropNullValues(obj)
           .remove("compacted")
           .remove("expanded")
           .remove("remoteContexts")
           .add(keywords.context, context.value)
       }
     }
   }
+
+  private def dropNullValues(j: JsonObject): JsonObject = j.filter { case (_, v) => !v.isNull }
```
dantb, line 352: To drop `"tag": null`. It's already provided on `Json` but not `JsonObject`

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -151,6 +157,20 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap {
...
+        ("/v1/resources/myorg/myproject/myschema?tag=mytag", myId7, schema1.id)
+      )
+      forAll(endpoints) { case (endpoint, id, schema) =>
+        val payload = jsonContentOf("resources/resource.json", "id" -> id)
+        Post(endpoint, payload.toEntity) ~> routes ~> check {
+          status shouldEqual StatusCodes.Created
+          response.asJson shouldEqual resourceMetadata(projectRef, id, schema, (nxv + "Custom").toString)
+        }
+      }
+    }
```
shinyhappydan, line 172: I think you need to look up the resource by the tag after creation. Otherwise the operation could do nothing and this test would pass

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImplSpec.scala
```diff
@@ -143,7 +161,26 @@ class ResourcesImplSpec
...
+            subject = subject
+          )
+        }
+      }
+
+      "succeed and tag with the id present on the payload and passed" in {
+        val list =
+          List(
+            (myId12, "_", resourceSchema),
+            (myId13, "myschema", Latest(schema1.id))
```
shinyhappydan, line 177: Similar with this test: we probably need to resolve by tag afterwards

shinyhappydan (changes requested):

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
imsdu: Yep or break it down in some way, this does too many things

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
dantb: @shinyhappydan Ah yeah, I checked that it was as expected in the route manually - but really a fake `Resources` should be injected in check that it's called with the right tag? But this is essentially an integration test so guess that isn't possible 😞 

Also could it make sense to have the tag on the creation response?

Anyway - updated with the check!

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImplSpec.scala
dantb: Why's that, is there something asynchronous happening / other state mutated? The return value from `resources.create` isn't enough to check against?

shinyhappydan: If you think about the functionality you are adding, this test doesn't assert that it has occurred. 

The outcome of creating with a tag is really that the resource is resolvable by that tag. The fact that the data returned from the `create` contains tag data does not indicate that the tag has been added correctly. The way to think about it is that I could comment out the code that persists the tag, and this test would not fail

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -163,10 +194,29 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap {
...
+          status shouldEqual StatusCodes.Created
+          response.asJson shouldEqual
+            resourceMetadata(projectRef, id, schema, (nxv + "Custom").toString, createdBy = alice, updatedBy = alice)
+          resourceTags(resources, id).contains(tag) shouldBe true
+        }
       }
     }
 
+    def resourceTags(resources: Resources, id: Iri): Tags =
+      resources.fetch(ResourceRef(id), projectRef).accepted.value.tags
```
shinyhappydan, line 218: I think even better than this would be to look up by tag. I know this is similar, but lookup is the reason we have this functionality

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
shinyhappydan: I'm not for/against the tag being on the creation response, but it wouldn't help us determine that tag creation happened

dantb:
@shinyhappydan should be good to go I think

shinyhappydan (approved):

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -163,10 +194,29 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap {
...
+        ("/v1/resources/myorg/myproject/myschema/myid9?tag=mytag", myId9, schema1.id)
+      )
+      val (routes, resources) = routesWithDecodingOption(DecodingOption.Strict)
+      forAll(endpoints) { case (endpoint, id, schema) =>
+        val payload = jsonContentOf("resources/resource.json", "id" -> id)
+        Put(endpoint, payload.toEntity) ~> asAlice ~> routes ~> check {
+          status shouldEqual StatusCodes.Created
+          response.asJson shouldEqual
+            resourceMetadata(projectRef, id, schema, (nxv + "Custom").toString, createdBy = alice, updatedBy = alice)
+          resourceTags(resources, id).contains(tag) shouldBe true
```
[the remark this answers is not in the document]
dantb: Done (as an aside, if it's a lookup we could call it `lookup` instead of `fetch`?)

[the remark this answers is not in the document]
shinyhappydan: I don't really have an opinion on that, they're both "correct" to me

shinyhappydan (dismissed):

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/Vocabulary.scala
```diff
@@ -173,6 +173,9 @@ object Vocabulary {
     val originalSource        = Metadata("original_source")
     val effectiveApiMappings  = Metadata("effectiveApiMappings")
 
+    // TODO already added in another ticket, remove here after rebase
```
imsdu, line 176: Remove the TODO ?

imsdu (approved):

imsdu (approved):

PR #4340, opened 2023-10-06
Remove callback to Delta to check the token, validate it locally

### LINKED ISSUE #4275: Remove/Cache the callback from storage to Delta to get the subject/realm information
**Motivation**
Before each operation, the storage app calls back Nexus with the token to get information for this caller.
This adds latency and can lead to some timeouts errors.

As the caller information (subject + issuer/realm) are already present in the token, we can just parse the token and get rid of this call.

**Acceptance criteria**
1. The token is parsed by the storage app to extract the required user information
2. The callback to Delta is suppressed

imsdu:
Fixes #4275


tests/docker/config/storage.conf
```diff
@@ -8,21 +8,14 @@ app {
...
   storage {
     root-volume = "/tmp"
     protected-directory = "protected"
     fixer-enabled = false
   }
 
-  delta {
-    public-iri = "https://test.nexus.bbp.epfl.ch"
-    internal-iri = "http://delta:8080"
+  authorization {
```
imsdu, line 17: We can't really test this part in the integration tests anymore as we need the public keys  for the realm.
In the different envs, we will get notified if Delta can't interact with the storage though

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/routes/AuthDirectivesSpec.scala
```diff
@@ -4,98 +4,111 @@ import akka.http.scaladsl.model.StatusCodes
...
+          Instant.now().plusSeconds(100L),
+          Instant.now().minusSeconds(100L),
+          None,
+          None,
+          false,
+          Some(subject)
+        )
+        .value
+
+    implicit val anonymousMethod: AuthorizationMethod =
```
olivergrabinski, line 76: Not the anonymous method ☝️

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/auth/AuthorizationError.scala
```diff
@@ -0,0 +1,21 @@
+package ch.epfl.bluebrain.nexus.storage.auth
+
+import ch.epfl.bluebrain.nexus.delta.kernel.jwt.TokenRejection
+
+sealed abstract class AuthorizationError(message: String) extends Exception with Product with Serializable {
+  override def fillInStackTrace(): AuthorizationError = this
+  override def getMessage: String                     = message
+}
```
shinyhappydan, line 8: I think it makes more sense to say `extends Exception(message)` rather than override? 

Actually, there are more things usable from the Exception constructor:
```
Exception(String message, Throwable cause, boolean enableSuppression,boolean writableStackTrace)
```

Setting `writableStackTrace` to false does the equivalent of what you're writing here

Given that this is likely a pattern we are going to repeat a lot, I would suggest making some sort of `DeltaError` that you could extend instead, which could capture this boilerplate and hide it

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/routes/AuthDirectivesSpec.scala
```diff
@@ -4,98 +4,111 @@ import akka.http.scaladsl.model.StatusCodes
...
-        deltaIdentities()(any[Option[AccessToken]]) shouldReturn
-          Task.raiseError(DeltaIdentitiesClientError.IdentitiesClientStatusError(StatusCodes.Unauthorized, ""))
-        val route                               = Routes.wrap(extractCaller.apply(_ => complete("")))
-        Get("/") ~> route ~> check {
-          status shouldEqual StatusCodes.Unauthorized
-        }
+    }
+
+    "Fail with a token signed with another key" in {
+      val anotherKey: RSAKey = new RSAKeyGenerator(2048).keyID(genString()).generate()
```
olivergrabinski, line 101: It's possible to just use `generateKey` here

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/auth/AuthorizationError.scala
shinyhappydan: Of course it's also fine to do this another day, it's not a big deal

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/auth/AuthorizationMethod.scala
```diff
@@ -0,0 +1,90 @@
...
+                                     "string",
+                                     "'method' value must be one of ('anonymous', 'verify-token')"
+                                   ),
+                                   obj
+                                 )
+                               )
+                             )
+                         }
+      } yield method
+    }
```
shinyhappydan, line 86: I'm curious what the difference is between this and the method from `Credentials`, I would have thought that the result was the same?

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/routes/Routes.scala
```diff
@@ -85,16 +85,13 @@ object Routes {
...
     //TODO: Fetch Bearer token and verify identity
     wrap {
       concat(
         AppInfoRoutes(config.description).routes,
-        (pathPrefix(config.http.prefix) & extractToken) { implicit token =>
-          extractCaller.apply {
-            case Caller(config.subject.subjectValue, _) => StorageRoutes(storages).routes
-            case _                                      => failWith(AuthenticationFailed)
-          }
+        (pathPrefix(config.http.prefix) & validateUser) {
```
shinyhappydan, line 93: This is just a thought I had, but I wonder if this makes more sense that you say `validUser` here

Now I've learned a bit about akka I wonder if this is suppose to read as:

- The request has a path prefix of `config.http.prefix`
- The request has a valid user

Not an issue at all just something I was thinking about

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/jwt/ParsedToken.scala
```diff
@@ -18,7 +25,21 @@ final case class ParsedToken private (
...
+    val proc        = new DefaultJWTProcessor[SecurityContext]
+    val keySelector = new JWSVerificationKeySelector(JWSAlgorithm.RS256, new ImmutableJWKSet[SecurityContext](keySet))
+    proc.setJWSKeySelector(keySelector)
+    audiences.foreach { aud =>
+      proc.setJWTClaimsSetVerifier(new DefaultJWTClaimsVerifier(aud.toSet.asJava, null, null, null))
+    }
+    Either
+      .catchNonFatal(proc.process(jwtToken, null))
+      .leftMap(err => InvalidAccessToken(subject, issuer, err.getMessage))
+  }
```
shinyhappydan, line 40: I'm expecting there might be complication with this, but it might be better if we has a constructor that returned an already validated token? There is something about validate being a method you can optionally call which feels a bit weird. Not a big deal though since you've got testing to cover

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/jwt/ParsedToken.scala
```diff
@@ -33,13 +54,13 @@ object ParsedToken {
...
-        .leftMap(_ => InvalidAccessTokenFormat)
+        .leftMap { e => InvalidAccessTokenFormat(e.getMessage) }
 
     def claims(jwt: SignedJWT): Either[TokenRejection, JWTClaimsSet] =
       Either
         .catchNonFatal(jwt.getJWTClaimsSet)
-        .filterOrElse(_ != null, InvalidAccessTokenFormat)
-        .leftMap(_ => InvalidAccessTokenFormat)
+        .leftMap { e => InvalidAccessTokenFormat(e.getMessage) }
+        .filterOrElse(_ != null, InvalidAccessTokenFormat("No claim is defined."))
```
shinyhappydan, line 63: If we are expecting `null` then we could instead say `Either.catchNonFatal(Option(jwt.getJWTClaimsSet))` so the null never escapes

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/routes/AuthDirectivesSpec.scala
```diff
@@ -4,98 +4,111 @@ import akka.http.scaladsl.model.StatusCodes
...
-        implicit val token: Option[AccessToken] = None
-        deltaIdentities()(any[Option[AccessToken]]) shouldReturn
-          Task.raiseError(DeltaIdentitiesClientError.IdentitiesClientStatusError(StatusCodes.Forbidden, ""))
-        val route                               = Routes.wrap(extractCaller.apply(_ => complete("")))
-        Get("/") ~> route ~> check {
-          status shouldEqual StatusCodes.Forbidden
-        }
+    }
+
+    "Fail with an invalid token" in {
```
olivergrabinski, line 108: Worth checking expired and not yet active tokens?

shinyhappydan (approved):

storage/src/test/resources/app.conf
```diff
@@ -1,8 +1,7 @@
 # All application specific configuration should reside here
 app {
-  # Allowed subject to perform calls
-  subject {
-    # flag to decide whether or not the allowed subject is Anonymous or a User
-    anonymous = true
+  # Authorization method
+  authorization {
+    method = anonymous
```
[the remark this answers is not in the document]
olivergrabinski: This is just the test config 😃

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/routes/AuthDirectivesSpec.scala
imsdu: This is already tested elsewhere and handled when parsing the token, I did not want to cover all the cases again here

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/jwt/ParsedToken.scala
imsdu: It comes from the way we do things in `IdentitiesImpl`, we do some surface checks first and extract the issuer to:
- Get the matching realm and the public keys associated to it
- Validate the token with the keys
I will check if there is a way to improve that a bit

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/auth/AuthorizationError.scala
imsdu: Done for the first part, for the second, I don't remember if there was a small difference between the two approachs.

But we can have a dedicated task to review this

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/jwt/ParsedToken.scala
shinyhappydan: One other thing: if we don't use the `JWTClaimsSet` then it's probably clearer to return `Unit`

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/jwt/ParsedToken.scala
```diff
@@ -18,7 +25,24 @@ final case class ParsedToken private (
...
+    audiences.foreach { aud =>
+      proc.setJWTClaimsSetVerifier(new DefaultJWTClaimsVerifier(aud.toSet.asJava, null, null, null))
+    }
+    Either
+      .catchNonFatal(proc.process(jwtToken, null))
+      .bimap(
+        err => InvalidAccessToken(subject, issuer, err.getMessage),
+        _ => ()
+      )
+  }
```
shinyhappydan, line 43: As someone unfamiliar with this process, it's not clear what an audience or keySet is. And it's not clear what about them is being validated

I think with some name changes and perhaps pulling some code into private methods this could be made really clear.

shinyhappydan (approved):

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/jwt/ParsedToken.scala
imsdu: This part is already documented in the realm documentation:
https://bluebrainnexus.io/snapshot/docs/delta/api/realms-api.html#payload
And in the configuration file

PR #4346, opened 2023-10-09
Ensure file errors are written correctly

### LINKED ISSUE #4308: A failure in authentication when downloading a file from remote storage does not result in an error
**Describe the bug**
When storage authentication fails, a file download will "succeed" but the contents of the file will be an error response

Delta

**To Reproduce**
1. Set up the wrong password in your storage auth configuration
2. Download a file from a remote storage using fusion
3. Open the file and see that the contents are a Delta error

**Expected behavior**
Delta should return an error response

shinyhappydan:
Fixes #4308


delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutesSpec.scala
```diff
@@ -473,6 +550,17 @@ class FilesRoutesSpec
...
+    "fail when auth fails" in {
+      val mockRoutes = createRoutes(filesWithErrorInFileContents[ResourceRejection](ResourceRejection.BlankResourceId))
+      Get("/v1/files/org/proj/file1") ~> Accept(`*/*`) ~> mockRoutes ~> check {
+        contentType.value shouldEqual `application/ld+json`.value
+        response.asJson shouldEqual jsonContentOf(
+          "/files/errors/blank-id.json"
+        )
+        status shouldEqual StatusCodes.BadRequest
+      }
+    }
```
shinyhappydan, line 563: I don't see an alternative to creating a "mock" files which I can inject the error into. I realise that the `Files` interface is really too big, but there's not a lot I can do about it in the scope of this ticket

I'm using `ResourceRejection.BlankResourceId` only because it's easy to create

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/JsonLdValue.scala
```diff
@@ -24,4 +29,18 @@ object JsonLdValue {
...
+      override def expand(
+          value: JsonLdValue
+      )(implicit opts: JsonLdOptions, api: JsonLdApi, rcr: RemoteContextResolution): IO[RdfError, ExpandedJsonLd]  =
+        value.encoder.expand(value.value)
+      override def compact(
+          value: JsonLdValue
+      )(implicit opts: JsonLdOptions, api: JsonLdApi, rcr: RemoteContextResolution): IO[RdfError, CompactedJsonLd] =
+        value.encoder.compact(value.value)
+    }
+  }
```
shinyhappydan, line 45: `JsonLdValue` is a slightly strange class. I can see why we have it, but I think also this typeclass makes sense for it. The alternative is that we do some wrangling in ResponseToJsonLd, but I think this makes more sense

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/error/AuthTokenError.scala
```diff
@@ -45,8 +39,6 @@ object AuthTokenError {
         JsonObject(keywords.tpe := "AuthTokenHttpError", "reason" := r.reason)
       case AuthTokenNotFoundInResponse(r) =>
         JsonObject(keywords.tpe -> "AuthTokenNotFoundInResponse".asJson, "reason" := r.message)
-      case ExpiryNotFoundInResponse(r)    =>
-        JsonObject(keywords.tpe -> "ExpiryNotFoundInResponse".asJson, "reason" := r.message)
```
shinyhappydan, line 49: This was unused

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutesSpec.scala
```diff
@@ -107,21 +105,100 @@ class FilesRoutesSpec
...
+          "file.name",
+          ContentTypes.`text/plain(UTF-8)`,
+          1024,
+          IO.raiseError(error)
+        )
+      )
+      override def fetch(id: IdSegmentRef, project: ProjectRef): IO[FileRejection, FileResource] = ???
+      override private[files] def attributesUpdateStream(offset: Offset)                         = ???
+      override private[files] def updateAttributes(iri: Iri, project: ProjectRef)                = ???
+    }
```
shinyhappydan, line 189: This is what I wanted to turn scalafmt off for. An alternative might be to put it in another file, it's pretty horrible boilerplate

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutesSpec.scala
imsdu: The bug is related to `FileResponse` and not files right ?
`FileResponse` is also used in the archives plugin for example ?
What about setting a route that would emit a successful/failing `FileResponse` ? It should be a lot easier ?

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutesSpec.scala
olivergrabinski: Strange. I checked out this branch to try it out and it works fine if I just add `// format: off` before line 128. It doesn't reformat any changes whether I use cmd+opt+L in IntelliJ or if I use `scalafmtAll` directly in sbt.

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutesSpec.scala
shinyhappydan: I would prefer to test the actual functionality where the bug was occurring, rather than make another situation where the bug would occur. I think the problem here is that our current design makes mocking this really hard. I think if we don't like the outcome of this, the solution is to fix our design rather than work around it.

I suppose it's possible that the archive functionality would have a problem if authentication failed, but I don't think it would manifest the same way. We would need a different test to catch that.

I realise this boilerplate isn't ideal, but other than that I think this test makes the most sense. If the boilerplate is annoying we can hide it

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutesSpec.scala
shinyhappydan: Probable error 40

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJsonLd.scala
```diff
@@ -106,12 +106,9 @@ object ResponseToJsonLd extends FileBytesInstances {
       override def apply(statusOverride: Option[StatusCode]): Route = {
         val flattened = io.flatMap { fr => fr.content.attempt.map(_.map { s => fr.metadata -> s }) }.attempt
         onSuccess(flattened.runToFuture) {
-          case Left(complete: Complete[E]) => emit(complete)
-          case Left(reject: Reject[E])     => emit(reject)
-          case Right(Left(c))              =>
-            implicit val valueEncoder = c.value.encoder
-            emit(c.value.value)
-
+          case Left(complete: Complete[E])       => emit(complete)
```
imsdu, line 109: This file is currently duplicated for the Cats migration, we need to keep the two of them in sync

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutesSpec.scala
shinyhappydan: We have already said we want to refactor the way that files are handled. When that happens, it would probably be very easy to improve this test. I think a test with a mock makes sense in principle, the only problem is that the design of `Files`

I'm not sure what `FilesRoutesSpec` is supposed to be testing, but given that we have `FilesSpec` as a real integration test, it seems like `FileRoutesSpec` being a unit test that handles this sort of thing makes a lot of sense

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJsonLd.scala
shinyhappydan: Can I port the FileRoutes across to the cats? I don't feel great about this being untested

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutesSpec.scala
imsdu: I am not arguing that `Files` would need to be refactored but more about the fact that this mock on `Files` does not bring value when what you want to test concerns `FileResponse` and how it is sent to the user.
Besides having a dedicated test class for this in the sdk can be interesting for the following reasons:
* The test would live in the same module as `FileResponse` in the sdk and not in a related plugin, it is always better to have it closer when possible
* It won't be mixed in the folklore of Files and its numerous tests
* Until the refactoring is done, we will have to deal with this mock everytime we modify `Files`

shinyhappydan: Those are good points, I think your original suggestion of a new routes returning a `FileResponse` makes the most sense. This would essentially be a unit test of `ResponseToJsonLd` (albeit only for the file response portion), which is basically the closest it can be

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJsonLd.scala
shinyhappydan: Perhaps now I'm unit test this class, I can test both

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutesSpec.scala
shinyhappydan: My only explanation is that I was doing it inside of the files block, and perhaps that's not legitimate

imsdu (dismissed):

imsdu (approved):

PR #4352, opened 2023-10-10
Migrate archives to Cats Effect

### LINKED ISSUE #4328: Migrate archives to CE

imsdu:
Fixes #4328

### DIFF
--- a/delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
+++ b/delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
@@ -1,15 +1,16 @@
 package ch.epfl.bluebrain.nexus.delta.plugins.archive
 
-import cats.effect.Clock
+import cats.effect.{Clock, IO}
 import ch.epfl.bluebrain.nexus.delta.kernel.kamon.KamonMetricComponent
 import ch.epfl.bluebrain.nexus.delta.kernel.syntax._
-import ch.epfl.bluebrain.nexus.delta.kernel.utils.{IOUtils, UUIDF}
+import ch.epfl.bluebrain.nexus.delta.kernel.utils.{IOInstant, UUIDF}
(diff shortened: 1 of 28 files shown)


delta/plugins/archive/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/Archives.scala
```diff
@@ -50,30 +50,14 @@ class Archives(
...
-    * Creates an archive with a system generated id.
-    *
-    * @param project
-    *   the archive parent project
-    * @param value
-    *   the archive value
-    * @param subject
-    *   the subject that initiated the action
-    */
-  def create(
```
imsdu, line 67: Dead code

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/stream/CatsStreamConverter.scala
```diff
@@ -0,0 +1,89 @@
...
+import cats.effect._
+import cats.syntax.all._
+import fs2._
+
+/**
+  * Converts a fs2 stream to an Akka source Original code from the streamz library from Martin Krasser (published under
+  * Apache License 2.0):
+  * https://github.com/krasserm/streamz/blob/master/streamz-converter/src/main/scala/streamz/converter/Converter.scala
+  */
+object CatsStreamConverter {
```
imsdu, line 15: Yet another copy of a class existing for Monix for Cats Effects.

olivergrabinski: Why not migrate it? Seems like it would have only a small impact outside of archives?

olivergrabinski (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/stream/CatsStreamConverter.scala
imsdu: It is used by SSEs and composite views too for the moment

PR #4353, opened 2023-10-11
Migrate permissions to CE

### LINKED ISSUE #4330: Migrate the permissions module to CE

shinyhappydan:
Fixes #4330

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/marshalling/RdfExceptionHandler.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/marshalling/RdfExceptionHandler.scala
@@ -13,6 +13,7 @@ import ch.epfl.bluebrain.nexus.delta.sdk.directives.DeltaDirectives._
 import ch.epfl.bluebrain.nexus.delta.sdk.error.ServiceError.AuthorizationFailed
 import ch.epfl.bluebrain.nexus.delta.sdk.error.{AuthTokenError, IdentityError, ServiceError}
 import ch.epfl.bluebrain.nexus.delta.sdk.model.BaseUri
+import ch.epfl.bluebrain.nexus.delta.sdk.permissions.model.PermissionsRejection
 import com.typesafe.scalalogging.Logger
 import io.circe.syntax._
 import io.circe.{Encoder, JsonObject}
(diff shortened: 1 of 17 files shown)


delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/ce/CatsIOValues.scala
```diff
@@ -23,6 +23,8 @@ trait CatsIOValues extends CatsIOValuesLowPrio {
   implicit final class CatsIOValuesOps[A](private val io: IO[A]) {
     def accepted: A = io.unsafeRunSync()
 
+    def rejected2(implicit pos: source.Position): Throwable = rejectedWith[Throwable]
+
```
shinyhappydan, line 27: `rejected2` should be renamed `rejected`

`rejected` should be renamed to something like `rejectedAssert`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/marshalling/RdfExceptionHandler.scala
```diff
@@ -32,11 +33,12 @@ object RdfExceptionHandler {
       base: BaseUri
   ): ExceptionHandler =
     ExceptionHandler {
-      case err: IdentityError  => discardEntityAndForceEmit(err)
-      case err: AuthTokenError => discardEntityAndForceEmit(err)
-      case AuthorizationFailed => discardEntityAndForceEmit(AuthorizationFailed: ServiceError)
-      case err: RdfError       => discardEntityAndForceEmit(err)
-      case err: Throwable      =>
+      case err: IdentityError        => discardEntityAndForceEmit(err)
+      case err: PermissionsRejection => discardEntityAndForceEmit(err)
```
imsdu, line 37: Neither solution is great but so far we tend to do a `attemptNarrow` to have an `IO[Either[_, _]]` to handle with errors (see the routes of the already migrated parts)

A better solution long term (so not in this PR and probably in a while) would be to have the different modules declare their exception handler and to use.

shinyhappydan: Yeah I was thinking about this. It's one of those situations where typeclasses don't work so well. I wonder if what we need is something more like a trait on `Rejection` where we "force" user-facing errors to provide the necessary typeclasses. That way it's forced by the compiler and is a bit more streamlined

imsdu (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/marshalling/RdfExceptionHandler.scala
shinyhappydan: Sorry, I misread this. I'll merge this but I'll do the fix you mentioned in the next PR I'm working on now

PR #4356, opened 2023-10-11
Add proxy pass route for id resolution

### LINKED ISSUE #4347: Resolution of resources coming from proxy pass
**Motivation**
In #4228, we introduced the endpoint to resolve a resource simply by providing an ID.

In the case the client is a browser, the redirection that happens in the existing endpoint is not enough. Indeed, a user will paste the `@id` in the browser, in which case the call (redirected, going to `/v1/resolve`) will be unauthenticated and end up not being resolved for the user. The user token needs to be "injected".

`@id` pasted in browser -> redirection to `/v1/resolve-proxy-pass` -> fusion page to inject user token -> `/v1/resolve`

The following proxy pass exists:
```
https://bbp.epfl.ch/neurosciencegraph/data/* to https://nise.bbp.epfl.ch/nexus/v1/resolve-proxy-pass/*
```

It is required to implement `/v1/resolve-proxy-pass/{uuid}` so that:
* If the `Accept: text/html` header is present
    * Redirect to Fusion page that authenticates the user and sends a request to `https://nise.bbp.epfl.ch/nexus/v1/resolve/{id}` where `{id} = https://bbp.epfl.ch/neurosciencegraph/data/{uuid}` (url encoded).
* Otherwise, call the `/v1/resolve` endpoint on the id directly (the user needs to provide authenticated).

**Acceptance criteria**
1. If `Accept: text/html` is present. A call to `/v1/resolve-proxy-pass/{uuid}` results in a redirection to a "injection" fusion page with the id = `https://bbp.epfl.ch/neurosciencegraph/data/{uuid}` to be resolved.
2. Otherwise, a call to `/v1/resolve-proxy-pass/{uuid}` results in `/v1/resolve/{id}` where `{id} = https://bbp.epfl.ch/neurosciencegraph/data/{uuid}`.

**Open questions**
1. Address of the Fusion page?
2. Are we sure `{uuid}` will always be a UUID? How to make this modular enough as to enable other potential resolutions that don't start with `https://bbp.epfl.ch/neurosciencegraph/data` in the future.

olivergrabinski:
Closes #4347 

Screenshot 2023-10-13 at 09-41-56 Id Resolution · Blue Brain Nexus

### DIFF
--- a/delta/plugins/elasticsearch/src/main/resources/elasticsearch.conf
+++ b/delta/plugins/elasticsearch/src/main/resources/elasticsearch.conf
@@ -50,7 +50,4 @@ plugins.elasticsearch {
   listing-bucket-size = 500
   # set to false to disable Elasticsearch indexing
   indexing-enabled = ${app.defaults.indexing.enable}
-
-  # base to use to reconstruct resource identifiers in the proxy pass
-  proxy-id-base = "https://bbp.epfl.ch"
 }
--- a/delta/plugins/elasticsearch/src/main/resources/elasticsearch.conf
+++ b/delta/plugins/elasticsearch/src/main/resources/elasticsearch.conf
@@ -50,4 +50,7 @@ plugins.elasticsearch {
   listing-bucket-size = 500
   # set to false to disable Elasticsearch indexing
   indexing-enabled = ${app.defaults.indexing.enable}
+
+  # base to use to reconstruct resource identifiers in the proxy pass
+  proxy-id-base = "https://bbp.epfl.ch"
 }
--- a/delta/plugins/elasticsearch/src/main/resources/elasticsearch.conf
+++ b/delta/plugins/elasticsearch/src/main/resources/elasticsearch.conf
@@ -10,8 +10,6 @@ plugins.elasticsearch {
   #   username= "elastic_user"
   #   password= "password"
   # }
-  # configuration of the Elasticsearch client
-  client = ${app.defaults.http-client-compression}
   # the elasticsearch event log configuration
   event-log = ${app.defaults.event-log}
   # the elasticsearch pagination config
@@ -20,8 +18,6 @@ plugins.elasticsearch {
   prefix = ${app.defaults.indexing.prefix}
   # configuration of the maximum number of view references allowed on an aggregated view
   max-view-refs = 20
-  # the maximum idle duration in between events on the indexing stream after which the stream will be stopped (min. 10 minutes).
-  idle-timeout = 30 minutes
   # In order to disable this feature, set an infinite time ('Inf')
   # idle-timeout = Inf
   #the maximum duration allowed so that synchronous indexing can complete
--- a/tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/IdResolutionSpec.scala
+++ b/tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/IdResolutionSpec.scala
@@ -10,6 +10,7 @@ import ch.epfl.bluebrain.nexus.tests.BaseSpec
 import ch.epfl.bluebrain.nexus.tests.Identity.listings.{Alice, Bob}
 import ch.epfl.bluebrain.nexus.tests.iam.types.Permission.Organizations
 import io.circe.Json
+import org.scalatest.Assertion
 
 class IdResolutionSpec extends BaseSpec {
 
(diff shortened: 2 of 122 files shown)


delta/plugins/elasticsearch/src/main/resources/elasticsearch.conf
```diff
@@ -50,4 +50,7 @@ plugins.elasticsearch {
   listing-bucket-size = 500
   # set to false to disable Elasticsearch indexing
   indexing-enabled = ${app.defaults.indexing.enable}
+
+  # base to use to reconstruct resource identifiers in the proxy pass
+  proxy-id-base = "https://bbp.epfl.ch"
```
imsdu, line 55: What about putting this one in the fusion config ?

(Which can be renamed)

Also localhost could be the default value

olivergrabinski: Made the changes. Also renamed `proxy-id-base` to `resolve-base`. If not suitable we can still change it later

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/IdResolutionSpec.scala
```diff
@@ -118,5 +137,9 @@ class IdResolutionSpec extends BaseSpec {
     List(Accept(MediaRange.One(`text/html`, 1f)))
   private def fusionResourcePageFor(encodedId: String) =
     s"https://bbp.epfl.ch/nexus/web/$ref11/resources/$encodedId".replace("%3A", ":")
+  private def fusionResolveEndpoint(encodedId: String) =
+    s"https://bbp.epfl.ch/nexus/web/resolve/$encodedId".replace("%3A", ":")
+  private def deltaResolveEndpoint(encodedId: String)  =
+    s"http://delta:8080/v1/resolve/$encodedId".replace("%3A", ":")
```
shinyhappydan, line 143: Shouldn't this be URL encoded?

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/IdResolutionSpec.scala
```diff
@@ -110,6 +115,20 @@ class IdResolutionSpec extends BaseSpec {
...
+        locationHeaderOf(response) shouldEqual deltaResolveEndpoint(encodedNeurosciencegraphId)
+      }(PredefinedFromEntityUnmarshallers.stringUnmarshaller)
+    }
+
+    "redirect to fusion resolve if the request comes to the proxy endpoint with text/html accept header is present" in {
+      deltaClient.get[String](s"/resolve-proxy-pass/$neurosciencegraphSegment", Bob, acceptTextHtml) { (_, response) =>
+        response.status shouldEqual StatusCodes.SeeOther
+        locationHeaderOf(response) shouldEqual fusionResolveEndpoint(encodedNeurosciencegraphId)
+      }(PredefinedFromEntityUnmarshallers.stringUnmarshaller)
+    }
```
shinyhappydan, line 130: Dream assertion would be

`response shouldBe redirectTo(fusionResolveEndpoint(encodedNeurosciencegraphId))`

It would be really nice to get rid of this boilerplate too but I don't know how: `(PredefinedFromEntityUnmarshallers.stringUnmarshaller)`

shinyhappydan (dismissed):

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/IdResolutionSpec.scala
olivergrabinski: I'm not sure I follow why it should be encoded? I don't think the location in a redirect needs to be encoded?

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/IdResolutionSpec.scala
olivergrabinski: Yeah but with this option we'd be matching on the response body which is not really relevant

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/IdResolutionSpec.scala
olivergrabinski: Only the last segment is encoded because it needs to be a single segment representing the resource ID, and that can contain `/`. Similar to how the id needs to be URL encoded when fetching resources, etc.

shinyhappydan: I wonder if other unpredictable things could happen if url characters are in the ID

shinyhappydan (approved):

PR #4357, opened 2023-10-11
Filter by type before fetching elems in composite projections

### LINKED ISSUE #4233: Filter by type before fetching elems in composite projections
**Motivation**
Complementary to #4160.
Improve the performance by not querying for elems that will be discarded in the composite projections.

olivergrabinski:
Closes #4233

### DIFF
--- a/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewDef.scala
+++ b/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewDef.scala
@@ -189,7 +189,13 @@ object CompositeViewDef {
     val fetchProgress: UIO[CompositeProgress] = compositeProjections.progress(view.indexingRef)
 
     def compileSource =
-      CompositeViewDef.compileSource(view.ref.project, compilePipeChain, graphStream, sinks.commonSink(view))(_, _)
+      CompositeViewDef.compileSource(
+        view.ref.project,
+        compilePipeChain,
+        graphStream,
+        sinks.commonSink(view),
+        projectionTypes(view)
+      )(_)
 
     def compileTarget = CompositeViewDef.compileTarget(compilePipeChain, sinks.projectionSink(view, _))(_)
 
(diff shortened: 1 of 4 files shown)


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/stream/CompositeGraphStream.scala
```diff
@@ -62,13 +65,20 @@ object CompositeGraphStream {
...
-    override def rebuild(source: CompositeViewSource, project: ProjectRef): Source = {
+    override def rebuild(
+        source: CompositeViewSource,
+        project: ProjectRef,
+        projectionTypes: Set[Iri]
+    ): Source = {
       source match {
         case p: ProjectSource       =>
-          Source(local.currents(project, p.selectFilter, _).through(drainSource))
+          val filter = p.selectFilter.copy(types = p.selectFilter.types ++ projectionTypes)
```
olivergrabinski, line 75: Do we need the source types here at all? Can we just use the projection types?

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewDef.scala
```diff
@@ -554,4 +564,11 @@ object CompositeViewDef {
     } yield target.id -> result
   }
 
+  /** Union of all resourceTypes specified in the view's projections */
+  private def projectionTypes(view: ActiveViewDef): Set[Iri] = {
+    val targets = view.value.projections
+    if (targets.exists(_.resourceTypes.isEmpty)) Set.empty[Iri]
+    else targets.foldLeft(Set.empty[Iri])(_ ++ _.resourceTypes)
+  }
```
shinyhappydan, line 572: The model here where empty set means there is no restriction is quite confusing. Consider changing it to be an ADT such as `NoRestriction` and `RestrictedTo(types: Set[Iri])`

shinyhappydan (approved):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/stream/CompositeGraphStream.scala
imsdu: I think it is slightly better to include them in the case where there is a limitation on the source but not in the projection.
In that case, you would have resources that won't be in the common space anyway.
It was also the original behaviour.

imsdu (approved):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/indexing/CompositeViewDef.scala
olivergrabinski: => #4374

PR #4360, opened 2023-10-12
Migrate Organizations to Cats Effect

### LINKED ISSUE #4332: Migrate organizations to CE

shinyhappydan:
Fixes #4332

### DIFF
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/OrganizationsModule.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/OrganizationsModule.scala
@@ -1,6 +1,6 @@
 package ch.epfl.bluebrain.nexus.delta.wiring
 
-import cats.effect.Clock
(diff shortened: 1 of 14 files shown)


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/OrganizationsModule.scala
```diff
@@ -33,15 +30,16 @@ object OrganizationsModule extends ModuleDef {
...
-        xas: Transactors
+        xas: Transactors,
+        contextShift: ContextShift[IO]
     ) =>
       OrganizationsImpl(
         scopeInitializations,
         config.organizations,
         xas
-      )(clock, uuidF)
+      )(clock, uuidF, contextShift)
```
imsdu, line 42: Not related to this PR but it feels that we should rework how we run those scope initialization.
It is not up to orgs/projects to know how they should be run.
One other thing is that it introduces circular dependencies between orgs and the modules with the scope initialization implementations

olivergrabinski (approved):

Issue #4313, opened 2023-09-28
Invalid ES settings can be saved and cause further errors


imsdu:
**Describe the bug**
According to logs in production, it seems possible to save invalid settings for an Elasticsearch view, causing errors down the line during indexing:

```
2023-09-28 09:32:33 WARN  c.e.b.n.d.sdk.http.HttpClientConfig - Error ch.epfl.bluebrain.nexus.delta.sdk.http.HttpClientError$HttpServerStatusError: an HTTP response to endpoint 'https://elasticsearch.bbp-ou-nise-production.svc:9200/nexus_2f921a01-2554-4cf3-86da-bc1fb0dd9ac7_14' with method 'HttpMethod(PUT)' that should have been successful, returned the HTTP status code '500 Internal Server Error': the request failed with body '{"error":{"root_cause":[{"type":"settings_exception","reason":"Failed to load settings from [{\"analysis\":{\"analyzer\":{\"autocomplete\":{\"filter\":[\"lowercase\"],\"tokenizer\":\"autocomplete\"},\"autocomplete_search\":{\"tokenizer\":\"whitespace\"},\"username\":{\"filter\":[{\"pattern\":\"https://bbp.epfl.ch/nexus/v1/realms/bbp/users/\",\"type\":\"pattern_replace\",\"replacement\":\"\"}]}},\"tokenizer\":{\"autocomplete\":{\"token_chars\":[\"letter\",\"digit\",\"punctuation\"],\"min_gram\":1,\"type\":\"edge_ngram\",\"max_gram\":6}}}}]"}],"type":"settings_exception","reason":"Failed to load settings from [{\"analysis\":{\"analyzer\":{\"autocomplete\":{\"filter\":[\"lowercase\"],\"tokenizer\":\"autocomplete\"},\"autocomplete_search\":{\"tokenizer\":\"whitespace\"},\"username\":{\"filter\":[{\"pattern\":\"https://bbp.epfl.ch/nexus/v1/realms/bbp/users/\",\"type\":\"pattern_replace\",\"replacement\":\"\"}]}},\"tokenizer\":{\"autocomplete\":{\"token_chars\":[\"letter\",\"digit\",\"punctuation\"],\"min_gram\":1,\"type\":\"edge_ngram\",\"max_gram\":6}}}}]","caused_by":{"type":"illegal_state_exception","reason":"only value lists are allowed in serialized settings"}},"status":500}'
Response body '{
  "error" : {
    "root_cause" : [
      {
[...]

PR #4365, opened 2023-10-12
Fix error when creating a view with invalid mappings or settings

imsdu:
Related to #4313

### DIFF
--- a/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchPluginModule.scala
+++ b/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchPluginModule.scala
@@ -23,7 +23,7 @@ import ch.epfl.bluebrain.nexus.delta.sdk.acls.AclCheck
 import ch.epfl.bluebrain.nexus.delta.sdk.deletion.ProjectDeletionTask
 import ch.epfl.bluebrain.nexus.delta.sdk.directives.DeltaSchemeDirectives
 import ch.epfl.bluebrain.nexus.delta.sdk.fusion.FusionConfig
-import ch.epfl.bluebrain.nexus.delta.sdk.http.HttpClient
+import ch.epfl.bluebrain.nexus.delta.sdk.http.{HttpClient, HttpClientConfig}
 import ch.epfl.bluebrain.nexus.delta.sdk.identities.Identities
 import ch.epfl.bluebrain.nexus.delta.sdk.identities.model.ServiceAccount
 import ch.epfl.bluebrain.nexus.delta.sdk.model._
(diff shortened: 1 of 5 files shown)


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchPluginModule.scala
```diff
@@ -53,8 +53,8 @@ class ElasticSearchPluginModule(priority: Int) extends ModuleDef {
   make[ElasticSearchViewsConfig].from { ElasticSearchViewsConfig.load(_) }
 
   make[HttpClient].named("elasticsearch-client").from {
-    (cfg: ElasticSearchViewsConfig, as: ActorSystem[Nothing], sc: Scheduler) =>
-      HttpClient()(cfg.client, as.classicSystem, sc)
+    val httpConfig = HttpClientConfig.noRetry(true)
```
imsdu, line 56: The retry strategy was responsible for timeouts.
We don't need retries when interacting with Elasticsearch so it is better to let things fail when it returns an error

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ElasticSearchViewsSpec.scala
```diff
@@ -84,6 +84,20 @@ class ElasticSearchViewsSpec extends BaseSpec with EitherValuable with CirceEq {
         }
     }
 
+    "fail to create a view with an invalid mapping" in {
+      val invalidMapping =
+        json"""{"mapping": "fail"}"""
+      val payload        = json"""{ "@type": "ElasticSearchView", "mapping": $invalidMapping }"""
+      deltaClient.put[Json](s"/views/$fullId/invalid", payload, ScoobyDoo) { expectBadRequest }
```
olivergrabinski, line 91: The `invalid` segment reads as if the view id is invalid, when the issue is the payload. The `payload` could be called `payloadWithInvalidMapping`, and the view id something else than `invalid`. Similar comment on the other test

olivergrabinski (approved):

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ElasticSearchViewsSpec.scala
```diff
@@ -84,6 +84,20 @@ class ElasticSearchViewsSpec extends BaseSpec with EitherValuable with CirceEq {
...
+    }
+
+    "fail to create a view with invalid settings" in {
+      val invalidSettings =
+        json"""{"analysis": "fail"}"""
+      val payload         = json"""{ "@type": "ElasticSearchView", "mapping": { }, "settings": $invalidSettings }"""
+      deltaClient.put[Json](s"/views/$fullId/invalid", payload, ScoobyDoo) { expectBadRequest }
+    }
```
olivergrabinski, line 99:
```suggestion
    "fail to create a view with an invalid mapping" in {
      val invalidMapping            =
        json"""{"mapping": "fail"}"""
      val payloadWithInvalidMapping = json"""{ "@type": "ElasticSearchView", "mapping": $invalidMapping }"""
      deltaClient.put[Json](s"/views/$fullId/invalid", payloadWithInvalidMapping, ScoobyDoo) { expectBadRequest }
    }

    "fail to create a view with invalid settings" in {
      val invalidSettings            =
        json"""{"analysis": "fail"}"""
      val payloadWithInvalidSettings =
        json"""{ "@type": "ElasticSearchView", "mapping": { }, "settings": $invalidSettings }"""
      deltaClient.put[Json](s"/views/$fullId/invalid", payloadWithInvalidSettings, ScoobyDoo) { expectBadRequest }
    }
```

olivergrabinski (approved):

PR #4367, opened 2023-10-16
Migrate SSE to cats effect

### LINKED ISSUE #4331: Migrate Sse module to CE

shinyhappydan:
Fixes #4331


olivergrabinski (dismissed):
* One thing that could be done is not to use the cats.IO => CIO alias. Instead rename the monix.IO => BIO as we did so far. Then we only have to delete things, and not rename things again. But it's easy enough to do and can be done later I guess.
* There's at least one place where `EitherT` could be used without much impact, which I think makes the code easier to read. But it's not critical

olivergrabinski (approved):

PR #4368, opened 2023-10-16
Migrate global event log and store to cats effect

### LINKED ISSUE #4363: Migrate Global event log to cats effects

shinyhappydan:
Fixes #4363

### DIFF
--- a/delta/plugins/archive/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/ArchiveDownloadSpec.scala
+++ b/delta/plugins/archive/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/ArchiveDownloadSpec.scala
@@ -39,7 +39,7 @@ import ch.epfl.bluebrain.nexus.delta.sourcing.model.Identity.Subject
 import ch.epfl.bluebrain.nexus.delta.sourcing.model.ResourceRef.Latest
 import ch.epfl.bluebrain.nexus.delta.sourcing.model.{Identity, Label, ProjectRef, ResourceRef}
 import ch.epfl.bluebrain.nexus.testkit.archive.ArchiveHelpers
-import ch.epfl.bluebrain.nexus.testkit.ce.CatsIOValues
+import ch.epfl.bluebrain.nexus.testkit.ce.{CatsIOValues, CatsRunContext}
 import ch.epfl.bluebrain.nexus.testkit.{EitherValuable, TestHelpers}
 import io.circe.syntax.EncoderOps
 import org.scalatest.matchers.should.Matchers
@@ -55,6 +55,7 @@ class ArchiveDownloadSpec
     with AnyWordSpecLike
     with Inspectors
     with EitherValuable
+    with CatsRunContext
     with CatsIOValues
     with OptionValues
     with TestHelpers
--- a/delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/GlobalEventLog.scala
+++ b/delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/GlobalEventLog.scala
@@ -1,6 +1,8 @@
 package ch.epfl.bluebrain.nexus.delta.sourcing
 
+import cats.effect.IO
 import cats.syntax.all._
+import ch.epfl.bluebrain.nexus.delta.kernel.effect.migration._
 import ch.epfl.bluebrain.nexus.delta.sourcing.EvaluationError.{EvaluationFailure, EvaluationTimeout}
 import ch.epfl.bluebrain.nexus.delta.sourcing.config.EventLogConfig
 import ch.epfl.bluebrain.nexus.delta.sourcing.event.Event.GlobalEvent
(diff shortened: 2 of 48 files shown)


dantb (approved):
Bit of a moot point since you've already done the work, but would it be easier to first move the sdk, since it relies quite heavily on the log? I guess as it is it will just be a few translations to remove later, so not too bad - as long as the compiler tells us what to do!

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/GlobalEventLog.scala
```diff
@@ -151,55 +150,54 @@ object GlobalEventLog {
...
+    override def stateOr[R <: Rejection](id: Id, rev: Int, notFound: => R, invalidRevision: (Int, Int) => R): IO[S] =
+      stateMachine.computeState(eventStore.history(id, rev).translate(ioToTaskK)).toCatsIO.flatMap {
         case Some(s) if s.rev == rev => IO.pure(s)
         case Some(s)                 => IO.raiseError(invalidRevision(rev, s.rev))
         case None                    => IO.raiseError(notFound)
       }
 
-    override def evaluate(id: Id, command: Command): IO[Rejection, (E, S)] =
-      stateStore.get(id).toUIO.flatMap { current =>
+    override def evaluate(id: Id, command: Command): IO[(E, S)] =
```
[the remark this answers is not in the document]
dantb: Actually that second statement probably won't work without [this](https://github.com/oleg-py/better-monadic-for) 😓 but in Scala 3 it is supported out of the box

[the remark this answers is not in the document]
shinyhappydan: I definitely like the way the function is split up in your suggestion. I'd like to experiment with `EitherT` but I won't do it in this PR, as there is a lot of code I would like to refactor and I'll never finish if I start :)

shinyhappydan:
> Bit of a moot point since you've already done the work, but would it be easier to first move the sdk, since it relies quite heavily on the log? I guess as it is it will just be a few translations to remove later, so not too bad - as long as the compiler tells us what to do!

I think you mean doing a more outwards-in approach? I think in general that probably would be better, leading to less rabbit-holes

shinyhappydan:
> Bit of a moot point since you've already done the work, but would it be easier to first move the sdk, since it relies quite heavily on the log? I guess as it is it will just be a few translations to remove later, so not too bad - as long as the compiler tells us what to do!

I think you mean doing a more outwards-in approach? I think in general that probably would be better, leading to less rabbit-holes

shinyhappydan:
> Bit of a moot point since you've already done the work, but would it be easier to first move the sdk, since it relies quite heavily on the log? I guess as it is it will just be a few translations to remove later, so not too bad - as long as the compiler tells us what to do!

I think you mean doing a more outwards-in approach? I think in general that probably would be better, leading to less rabbit-holes

shinyhappydan:
Oh good, something in here is making the unit tests time out 🤦

dantb (dismissed):

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/GlobalEventLog.scala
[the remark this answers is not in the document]
imsdu: Note that in `EphemeralLog` that has already been migrated, there is no such thing so it would be better if we could have some consistency between the different logs (`EphemeralLog` is way simpler though)

delta/plugins/archive/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/ArchiveDownloadSpec.scala
```diff
@@ -55,6 +55,7 @@ class ArchiveDownloadSpec
     with AnyWordSpecLike
     with Inspectors
     with EitherValuable
+    with CatsRunContext
```
imsdu, line 58: It would be better if we would not need two traits `CatsRunContext` and `CatsIOValues` here

delta/sourcing-psql/src/test/scala/ch/epfl/bluebrain/nexus/delta/sourcing/Arithmetic.scala
```diff
@@ -87,7 +88,9 @@ object Arithmetic {
     }
   }
 
-  sealed trait ArithmeticRejection extends Product with Serializable
+  sealed trait ArithmeticRejection extends Rejection with Product with Serializable {
```
imsdu, line 91: You do not need `Product` and `Serializable` here anymore

delta/plugins/archive/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/ArchiveDownloadSpec.scala
shinyhappydan: I don't disagree as such, but I think there are times where you need the context but not the values. I think it would be best to address this by looking at some of our base test classes and putting this in there so it's more hidden. Do you think we can do that just so we can put off dealing with it now?

imsdu: It can be done in further PR, can you create an issue for it ?

shinyhappydan: https://github.com/BlueBrain/nexus/issues/4371

PR #4369, opened 2023-10-16
Allow creating/linking file with tag

dantb:


delta/plugins/storage/src/test/resources/files/database/file-created.json
```diff
@@ -3,6 +3,7 @@
   "project" : "myorg/myproj",
   "storage" : "https://bluebrain.github.io/nexus/vocabulary/disk-storage?rev=1",
   "storageType" : "DiskStorage",
+  "tag" : null,
```
shinyhappydan, line 6: For back-compat clarity it might be better to write nothing than to write `null`?

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FileFixtures.scala
```diff
@@ -11,46 +13,60 @@ import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.nxv
...
-    size,
-    digest,
-    Client
-  )
+  def withUUIDF[T](id: UUID)(test: => T): T = (for {
+    old <- ref.getAndSet(id)
+    t   <- Task.delay(test).onError(_ => ref.set(old))
+    _   <- ref.set(old)
+  } yield t).accepted
+
```
shinyhappydan, line 56: I'm guessing this is because of the exact-matching and lack of fake storage? Maybe there's another way? Perhaps not asserting on the UUID is a possible

shinyhappydan (approved):

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FileFixtures.scala
dantb: Yeah and wanted to avoid having to rewrite all the other tests. It is possible to just use the returned one, reckon that would be preferable? Currently it does verify at least that the UUID generator is called, and that it's passed out on the response

shinyhappydan: Let's maybe have a chat IRL because I think I might be misunderstanding what's going on here

imsdu (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -142,18 +148,21 @@ final class Files(
...
+    * @param tag
+    *   the optional tag this file link is being created with, attached to the current revision
     */
   def createLink(
       storageId: Option[IdSegment],
       projectRef: ProjectRef,
       filename: Option[String],
       mediaType: Option[ContentType],
-      path: Uri.Path
+      path: Uri.Path,
```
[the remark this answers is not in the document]
dantb: Not sure what you had in mind but done something like this in https://github.com/BlueBrain/nexus/pull/4386

PR #4373, opened 2023-10-17
Support searching resources in ES by tag

### LINKED ISSUE #4335: Allow to filter resources by tag in the listing operation
**Motivation**
Now that tags are indexed in the default Elasticsearch views, we need to be able to filter resources/schemas/files with a given tag through the listing API

**Acceptance criteria**
1. It is only possible to filter by one tag at a time
2. It will return the latest version of the resource with this tag not the tagged version

dantb:
Fixes #4335


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ListingsSpec.scala
```diff
@@ -258,6 +272,31 @@ final class ListingsSpec extends BaseSpec with Inspectors with EitherValuable wi
       }
     }
 
+    "get the latest revision of previously tagged resources when queried by tag" in {
```
dantb, line 275: This searches on tags used on resources in different projects, is searching across different orgs also worth testing?

olivergrabinski: IMO should be ok since the scope of the search is already tested

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ListingsSpec.scala
```diff
@@ -256,6 +270,31 @@ final class ListingsSpec extends BaseIntegrationSpec {
...
+          "proj2"        -> proj13,
+          "id2"          -> resource13Id,
+          "self2"        -> resource13Self
+        ): _*
+      )
+
+      deltaClient.get[Json](s"/resources?tag=$tag1", Bob) { (json, response) =>
+        response.status shouldEqual StatusCodes.OK
+        filterSearchMetadata(json) should equalIgnoreArrayOrder(expected)
+      }
```
olivergrabinski, line 295: Here you are doing a search across all resources, that's why there is a test error. You should do /resources/{org}/{project}: https://bluebrainnexus.io/docs/delta/api/resources-api.html#within-a-project

olivergrabinski (changes requested):

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ListingsSpec.scala
dantb: Turns out this intermittently passed locally since I forgot the `eventually` 😓 probably since it's beefier than the GHA worker. Fixed now

olivergrabinski (approved):

PR #4376, opened 2023-10-18
Migrate Delta app to Cats effect

### LINKED ISSUE #4375: Migrate Delta Main to Cats Effect

imsdu:
Fixes #4375


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/IndexingAction.scala
```diff
@@ -48,8 +48,7 @@ trait IndexingAction {
       // We build and start the projections where the resource will apply
       _         <- projections(project, elem)
                      .translate(taskToIoK)
-                     // TODO make this configurable
-                     .parEvalMap(5) {
+                     .evalMap {
```
imsdu, line 51: I could not fix it for now, it was blocking the integration tests for ES views locally, I will create an issue for it but running them like this should be fine as there are not so many views per project

shinyhappydan (approved):

PR #4381, opened 2023-10-18
Migrate Expanded and CompactedJsonLd to Cats Effect

shinyhappydan:
This is an attempt to fix the test failure situation


delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/ce/CatsEffectSuite.scala
```diff
@@ -17,6 +17,7 @@ abstract class CatsEffectSuite
     with CatsRunContext
     with CatsEffectAssertions
     with CatsStreamAssertions
+    with CatsIOValues
```
imsdu, line 20: We should not have to import something related to ScalaTest here

Issue #4364, opened 2023-10-12
Migrate Files to Cats Effect


imsdu:
* Storages will be done in a second pass
* Bonus: split the Distage module for the storage plugin in 2

PR #4386, opened 2023-10-18
Refactor optional params in file operations

dantb:


shinyhappydan (commented):
I like the idea here but I have a couple of thoughts

Firstly, I think the `FileOptions` abstraction isn't quite right. For example, when calling create you could specify media type and filename, but they wouldn't be used. For that reason I'm not sure I would look at this as an improvement.

Another possibility is to use a case class as the sole argument to each method. It would mean more case classes but I think it's a better fit overall

I think another possibility for improvement might be to combine the ID and project?

I think also leaving it how it is also could be fine!

dantb:
> Firstly, I think the `FileOptions` abstraction isn't quite right. For example, when calling create you could specify media type and filename, but they wouldn't be used. For that reason I'm not sure I would look at this as an improvement.

Yeah I had separate datatypes initially, it just seemed like a lot more code. I agree it isn't nice to be able to represent things that won't be used

> Another possibility is to use a case class as the sole argument to each method. It would mean more case classes but I think it's a better fit overall

That could work actually 👍 are there any software best practices for the codebase written somewhere? E.g. whether to use overloading, default arguments, when to use type annotations, method visibility... Personally I don't really mind lots of parameters.

The most glaringly obvious problem with this class is that it has no abstraction, so the real one has to be used in all tests of dependents

dantb:
@shinyhappydan Just seen there's a ticket to migrate files to CE - #4364. I'll do that first and refactor on top, maybe more potential improvements will emerge

Issue #3401, opened 2022-07-04
Unexpected schema revision used when updating resources


bogdanromanx:
_reported internally by @annakristinkaufmann_

**Describe the bug**
I was trying to update a resource which had been validated against a schema which itself has been updated in the meantime. The resource update includes changes which are incompatible with the previous revision of that schema, which is why I want to validate the resource against the latest revision of that given schema. When specifying the schema in the `url` for the `PUT` request, no schema revision needs to be specified. The resource seems to validate against the latest revision of the schema. I.e. in the below example, the resource validates against the latest revision of the `datashapes:cellrecordseries`:

```
PUT https://nexusdomain/nexus/v1/resources/{org}/{proj}/datashapes:cellrecordseries/{uuid}?rev=1
```

However, when not specifying the schema in the `url` for the `PUT` request, the resource is being validated against the **old version** of the schema (for which validation now fails):

```
PUT https://nexusdomain/nexus/v1/resources/{org}/{proj}/_/{uuid}?rev=1
```

The behaviour I consider as non-desired is that even after updating against the latest revision of the schema, if I want to perform yet another update, I again experience the same behaviour (i.e.validation happens against a prior revision of the schema), even if I have already performed an update against the latest schema revision prior to that update-attempt. I.e. the following fails again if not providing a schema id:

```
PUT https://nexusdomain/nexus/v1/resources/{org}/{proj}/_/{uuid}?rev=2
```

**What part of Nexus is affected**
- [ ] Fusion
- [x] Delta
- [ ] CLI
- [ ] Documentation

**Expected behaviour**
1. if a resource is created against the latest revision of a schema, when performing an update, the validation should use the new latest schema revision, without the need to specify the schema again
2. if a resource is created against a specific revision of a schema (via a tag or rev qualifier) the validation should always use that very specific schema revision... either the exact revision or the revision represented by the tag at the moment when the update is performed

shinyhappydan:
This issue is also present when refreshing a resource

PR #4387, opened 2023-10-18
Allow to change the schema of a resource

### LINKED ISSUE #4343: Allow to change the schema of a resource
**Motivation**
Related to #737

In Delta, the schema of a resource is set at creation and can't be changed afterwards.
This makes things difficult as:
* People may want to experiment before setting up a schema
* Some resources are currently created without a schema and a schema can't be enforced now
* Schema migration from one to another
* ...

**Acceptance criteria**
1. When a user updates a resource
2. The user can provide a new schema by passing it in the dedicated segment when updating the resource

olivergrabinski:
Closes #4343


delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesSpec.scala
```diff
@@ -331,7 +331,8 @@ class ResourcesSpec
             source = newSource,
             updatedAt = time2,
             updatedBy = subject,
-            types = newTypes
+            types = newTypes,
+            schema = schemaRev
```
olivergrabinski, line 335: I believe this is because the `ResourceState` allows the schema to be a `ResourceRef` while in a `ResourceEvent` it has to specifically be a `ResourceRef.Revision`. This shouldn't be a big change if we want to change it 🤔

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ResourcesSpec.scala
```diff
@@ -380,6 +381,27 @@ class ResourcesSpec extends BaseSpec with EitherValuable with CirceEq {
...
+      val updateResourceAndSchema =
+        deltaClient.put[Json](s"/resources/$id1/new-schema/$resourceId?rev=1", payload, Rick)(_)
+
+      postResourceAndNewSchema >>
+        updateResourceAndSchema { (json, response) =>
+          response.status shouldEqual StatusCodes.OK
+          _rev.getOption(json) should contain(2)
+          _constrainedBy.getOption(json) should contain("http://delta:8080/v1/resources/" + id1 + "/_/new-schema")
+        }
+    }
```
shinyhappydan, line 404: It would be good to make it clear if you're updating or adding a schema to the resource. One way would be to label the payload something like `resourceWithASchemaPayload`

Or: 
```
thereIsASchema(schemaId) >>
  thereIsAResourceWithSchema(otherSchemaId) >>
  updateResourceSchema(resourceId, schemaId)
```

There's also a question of whether the functionality is actually tested here: when updating the schema the desired effect for the user is that now the resource is constrained by the schema. The fact that it's returning that it was constrained in the response doesn't actually show that it's constrained

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -355,22 +353,29 @@ object Resources {
...
-        IO.raiseWhen(targetRev <= 0 || targetRev > s.rev)(RevisionNotFound(targetRev, s.rev)).as(s)
+        raiseWhenDifferentSchema(c, s) >>
+          IO.raiseWhen(targetRev <= 0 || targetRev > s.rev)(RevisionNotFound(targetRev, s.rev)).as(s)
       }
     }
 
+    def raiseWhenDifferentSchema(c: ModifyCommand, s: ResourceState) =
+      IO.raiseWhen(c.schemaOpt.exists(cur => cur.iri != s.schema.iri))(
+        UnexpectedResourceSchema(s.id, c.schemaOpt.get, s.schema)
+      )
```
shinyhappydan, line 371: This is nice

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImplSpec.scala
```diff
@@ -346,6 +348,14 @@ class ResourcesImplSpec
           mkResource(expectedData).copy(rev = 3)
       }
 
+      "succeed with a schema update" in {
+        val updated      = source.removeKeys(keywords.id) deepMerge json"""{"number": 70}"""
+        val expectedData = ResourceGen.resource(myId2, projectRef, updated, Revision(schema3.id, 1))
+        resources
+          .update(myId2, projectRef, Some(schema3.id), 3, updated)
+          .accepted shouldEqual mkResource(expectedData).copy(rev = 4)
+      }
```
shinyhappydan, line 357: Another test where the actual desired outcome is the schema constraining the resource, not sure if it's testable from here

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -355,22 +353,29 @@ object Resources {
 
     def stateWhereTagExistsOnResource(c: ModifyCommand, tag: UserTag) = {
       stateWhereResourceExists(c).flatMap { s =>
-        IO.raiseWhen(!s.tags.contains(tag))(TagNotFound(tag)).as(s)
+        raiseWhenDifferentSchema(c, s) >>
```
imsdu, line 356: Not in the scope of this PR but we could consider dropping the check on the schema for commands where the schema is not relevant (ex: tag/refresh) and even the fetch operation

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImplSpec.scala
olivergrabinski: Should be better now?

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesSpec.scala
imsdu: Note that we will likely change part of how we model this to fix: https://github.com/BlueBrain/nexus/issues/3401

imsdu (approved):

PR #4388, opened 2023-10-19
Try to ensure that MUnit tests do no include ScalaTest matchers

shinyhappydan:
The main approach is to try and force assertion traits on only work on the appropriate classes. For example, `EitherMatchers` has a `self` type of `org.scalatest.Suite`, meaning that you can't accidentally use it in an MUnit test

General naming conventions:
`*Matchers` -> ScalaTest matchers (assertions)
`*Assertions` -> MUnit assertions
`*Values` -> Methods which return a value from an IO, such as accepted. Right now, this name doesn't indicate whether it's for MUnit or ScalaTest, this is because once we do more migration these classes will probably be absorbed by the Matchers/Assertions

I have also introduced some base test classes. If you're making a new test it's likely you should use one of these as the basis
`CatsEffectSpec` -> ScalaTest with CatsEffect
`CatsEffectSuite` -> MUnit with CatsEffect
`BioSpec` -> ScalaTest with BIO 
`BioSuite` -> MUnit with BIO

If you use these it should mean having to import less, as each comes with most of the assertions and helpers you will need


delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/MainSuite.scala
```diff
@@ -42,7 +42,7 @@ class MainSuite extends CatsEffectSuite with MainSuite.Fixture {
       make[ContextShift[IO]].fromValue(contextShift)
       make[Timer[IO]].fromValue(timer)
     }
-    val (cfg, config, cl, pDefs) = Main.loadPluginsAndConfig(pluginLoaderConfig).unsafeRunSync()
+    val (cfg, config, cl, pDefs) = Main.loadPluginsAndConfig(pluginLoaderConfig).accepted
```
imsdu, line 45: Whenever we can, we should avoid to block in MUnit tests

delta/plugins/archive/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/ArchiveDownloadSpec.scala
```diff
@@ -57,6 +57,8 @@ class ArchiveDownloadSpec
     with EitherValuable
     with CatsRunContext
     with CatsIOValues
+    with IOValues
```
imsdu, line 60: This plugin as been migrated, we should not see things related to Monix

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/acls/AclFixtures.scala
```diff
@@ -6,12 +6,14 @@ import ch.epfl.bluebrain.nexus.delta.sdk.permissions.Permissions.{acls, resolver
...
 import ch.epfl.bluebrain.nexus.delta.sourcing.model.Identity.{Anonymous, Group, User}
 import ch.epfl.bluebrain.nexus.delta.sourcing.model.Label
-import ch.epfl.bluebrain.nexus.testkit.EitherValuable
-import org.scalatest.Inspectors
+import org.scalatest.Suite
 
 import java.time.Instant
 
-trait AclFixtures extends EitherValuable with Inspectors {
+trait AclFixtures {
```
imsdu, line 13: This one does not seem to be strictly related to scalatest

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/DeltaSpec.scala
```diff
@@ -0,0 +1,7 @@
+package ch.epfl.bluebrain.nexus.testkit
+
+import org.scalatest.{Inspectors, OptionValues}
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+abstract class DeltaSpec extends AnyWordSpecLike with Matchers with EitherValuable with OptionValues with Inspectors
```
imsdu, line 7: `BaseSpec` ?
We already have another one in integration tests and a `BaseRouteSpec` for routes

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/ce/CatsSpec.scala
```diff
@@ -0,0 +1,5 @@
+package ch.epfl.bluebrain.nexus.testkit.ce
+
+import ch.epfl.bluebrain.nexus.testkit.DeltaSpec
+
+abstract class CatsSpec extends DeltaSpec with CatsRunContext with CatsEffectScalaTestAssertions
```
imsdu, line 5: CatsEffectSpec ?

delta/plugins/archive/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/ArchivesSpec.scala
```diff
@@ -38,6 +38,7 @@ class ArchivesSpec
     with Matchers
     with IOFixedClock
     with CatsRunContext
+    with CatsEffectScalaTestAssertions
```
imsdu, line 41: Extends `CatsEffectSpec` to avoid to repeat these ? (maybe in other places too)

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/DeltaSpec.scala
shinyhappydan: I wanted to not call it BaseSpec just because we already have a BaseSpec and I was a little confusing. I don't mind what they are call, but I would like to make sure they have different names. Perhaps `BaseSpec` could become `BaseIntegrationSpec`?

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/MainSuite.scala
shinyhappydan: That's ok, but I will not try and fix all these instances in this PR (I switched to accepted here just so that the timeout will be effective)

delta/plugins/archive/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/ArchiveDownloadSpec.scala
shinyhappydan: Ah, this is because `AclSimpleCheck` hasn't been migrated yet

delta/plugins/archive/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/archive/ArchivesSpec.scala
shinyhappydan: Ah yes this is because of `DoobieScalaTestFixture`, this will have to be fixed in another iteration

imsdu (dismissed):

olivergrabinski (approved):

shinyhappydan:
Fixes #4395

PR #4389, opened 2023-10-19
Partially migrate graph analytics to Cats Effect

dantb:
This is pretty much the bare minimum - the routes, module and a few other bits. `GraphAnalyticsStream` and `GraphAnalyticsSink` are tightly coupled to streaming logic in the sourcing module. I tried migrating sourcing, but it became a massive and probably dangerous change!

In `GraphAnalyticsCoordinator` when trying to remove BIO everywhere in the implementation details, I hit some issues around translating streams - I think related to stream interruptions. Even using `translateInterruptible` produced test failures for the coordinator. Behavioural changes like this are kind of scary, we may not even test all the edge cases 😅 it should be safer to migrate this together with the sourcing module.


delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/syntax/IOSyntax.scala
```diff
@@ -66,3 +68,10 @@ final class IOFunctorOps[A, F[_]: Functor](private val io: IO[F[A]]) {
     */
   def mapValue[B](f: A => B): IO[F[B]] = io.map(_.map(f))
 }
+
+final class IOOps[A](private val task: IO[A]) extends AnyVal {
+  def logErrors(action: String)(implicit logger: Logger): IO[A] =
```
imsdu, line 73: If possible, can you pass a `Log4cats` instance ?

delta/plugins/graph-analytics/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/graph/analytics/indexing/package.scala
```diff
@@ -12,25 +16,28 @@ package object indexing {
 
   val updateRelationshipsScriptId = "updateRelationships"
 
-  val scriptContent =
-    ioContentOf("elasticsearch/update_relationships_script.painless")
+  val scriptContent: UIO[String] =
```
imsdu, line 19: Can be done later but we migrate those too

imsdu (approved):

PR #4390, opened 2023-10-20
Migrate (possibly) all scalatests to use BaseSpec

shinyhappydan:
Still problematic: tests that extend TestKit (using the actor system, will presumably go away soon


imsdu (approved):

shinyhappydan:
Fixes #4395

PR #4391, opened 2023-10-20
Do not allow deprecated schemas/ontologies in imports

### LINKED ISSUE #1855: Deprecation status of imports in schemas
**Motivation**
When resolving imports in a schema during creation/update of a schema, deprecated resources and schemas are currently valid imports.

This should raise an error and the resolution process should continue to the next resolver

**Acceptance criteria**

1. When an import is resolved as a deprecated schema/context, an error should occur, logged in the report and the resolution should continue

imsdu:
Fixes #1855 

It is implemented at the resource resolution level as it felt weird to do it at the schema import part.
It is also not excluded that we exclude deprecated resources for other operations.


shinyhappydan (commented):
My guess is there's a reason why it makes more sense to do the exclusion in the resolution part, but I think it's a less straightforward system that doing it when importing from the schema. Is the resolution system another one of these features that isn't really used but adding complexity, or is this a more necessary one?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resolvers/ResolverResolution.scala
```diff
@@ -143,9 +144,16 @@ final class ResolverResolution[R](
...
+    for {
+      resourceOpt <- fetch(ref, projectRef)
+      result      <- resourceOpt.traverse(checkLatestDeprecated(ref, projectRef, _))
+    } yield {
+      result match {
+        case None        => ResolverReport.failed(resolver.id, projectRef -> ResolutionFetchRejection(ref, projectRef)) -> None
+        case Some(true)  =>
+          ResolverReport.failed(resolver.id, projectRef -> ResourceIsDeprecated(ref.original, projectRef)) -> None
+        case Some(false) => ResolverReport.success(resolver.id, projectRef)                                             -> resourceOpt
+      }
```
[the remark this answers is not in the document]
imsdu: Yep, I forgot to get back to this one, I will see what I can do

shinyhappydan:
> > My guess is there's a reason why it makes more sense to do the exclusion in the resolution part, but I think it's a less straightforward system that doing it when importing from the schema. Is the resolution system another one of these features that isn't really used but adding complexity, or is this a more necessary one?
> 
> Resolution is necessary and is widely used, we rely on it to resolve remote contexts in resources, imports in schemas, there are a couple of endpoints depending on it to help people find resources, ...

If that is the case then isn't it better to have this functionality localised to where it's used, rather than in this more general system

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resolvers/ResolverResolution.scala
```diff
@@ -143,9 +144,19 @@ final class ResolverResolution[R](
...
+    } yield {
+      result match {
+        // The resource has not been found in the project
+        case None        => ResolverReport.failed(resolver.id, projectRef -> ResolutionFetchRejection(ref, projectRef)) -> None
+        // The resource exists but the deprecation check is positive so we reject it
+        case Some(true)  =>
+          ResolverReport.failed(resolver.id, projectRef -> ResourceIsDeprecated(ref.original, projectRef)) -> None
+        // The resource has been successfully resolved
+        case Some(false) => ResolverReport.success(resolver.id, projectRef)                                             -> resourceOpt
+      }
```
shinyhappydan, line 159: I think by using meaningful variable and method names you don't really need the comments here

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resolvers/ResolverResolution.scala
```diff
@@ -177,11 +188,12 @@ final class ResolverResolution[R](
...
         case (f: ResolverFailedReport, _)  =>
           val resolve = for {
-            _        <- validateIdentities(projectRef)
-            resource <- fetch(ref, projectRef).flatMap { res =>
-                          IO.fromOption(res)(ResolutionFetchRejection(ref, projectRef))
-                        }
-            _        <- validateResourceTypes(extractTypes(resource), projectRef)
+            _            <- validateIdentities(projectRef)
+            resourceOpt  <- fetch(ref, projectRef)
+            resource     <- IO.fromOption(resourceOpt)(ResolutionFetchRejection(ref, projectRef))
```
shinyhappydan, line 193: I preferred this previously. An option would be to make a method to hide the boilerplate

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resolvers/ResolverResolution.scala
```diff
@@ -177,11 +188,12 @@ final class ResolverResolution[R](
...
+            _            <- validateResourceTypes(extractTypes(resource), projectRef)
+            isDeprecated <- runDeprecationCheck(ref, projectRef, resource)
+            _            <- IO.raiseWhen(isDeprecated)(ResourceIsDeprecated(ref.original, projectRef))
```
shinyhappydan, line 196: Another instance where a method can hide the boilerplate

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SchemasSpec.scala
```diff
@@ -6,59 +6,59 @@ import ch.epfl.bluebrain.nexus.testkit.{CirceEq, EitherValuable}
...
+      val baseSchemaPayload      = withPowerLevelShape(id = baseSchemaId, maxPowerLevel = 10000)
+      val importingSchemaId      = "https://localhost/importing-schema"
+      val importingSchemaPayload = withImportOfPowerLevelShape(id = importingSchemaId, importedSchemaId = baseSchemaId)
+
+      for {
+        _ <- deltaClient.post[Json](s"/schemas/$project", baseSchemaPayload, Rick) { expectCreated }
+        _ <- deltaClient.delete[Json](s"/schemas/$project/${UrlUtils.encode(baseSchemaId)}?rev=1", Rick) { expectOk }
+        _ <- deltaClient.post[Json](s"/schemas/$project", importingSchemaPayload, Rick) { (json, response) =>
+               response.status shouldEqual StatusCodes.BadRequest
+               `@type`.getOption(json) shouldEqual Some("InvalidSchemaResolution")
```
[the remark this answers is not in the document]
shinyhappydan: This still feels like the wrong error to receive

imsdu:
> > > My guess is there's a reason why it makes more sense to do the exclusion in the resolution part, but I think it's a less straightforward system that doing it when importing from the schema. Is the resolution system another one of these features that isn't really used but adding complexity, or is this a more necessary one?
> > 
> > 
> > Resolution is necessary and is widely used, we rely on it to resolve remote contexts in resources, imports in schemas, there are a couple of endpoints depending on it to help people find resources, ...
> 
> If that is the case then isn't it better to have this functionality localised to where it's used, rather than in this more general system

* This feature can be interesting in other cases such as context resolution when creating/updating resources
* Schema imports does not know how to resolve and fetch schemas and ontologies, it would not be able to fetch the latest version to check the deprecation status
* This way, it fits well in the error report returned whenever the resolution of imports fails

shinyhappydan:
> * This feature can be interesting in other cases such as context resolution when creating/updating resources
If no other feature is using it right now I would say YAGNI
> * Schema imports does not know how to resolve and fetch schemas and ontologies, it would not be able to fetch the latest version to check the deprecation status
We could give it the ability to check the deprecation status of the latest revision

It's just an option. Because this feels like quite a complex solution for what on the face of it seems like a simple problem, so I think it's worth making sure we want to pay that price

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resolvers/ResolverResolution.scala
shinyhappydan: Something like this? 

scala
```
    fetch(ref, projectRef).flatMap {
      case None => IO.pure(ResolverReport.failed(resolver.id, projectRef -> ResolutionFetchRejection(ref, projectRef)) -> None)
      case Some(resource) => shouldExcludeDueToDeprecation(ref, projectRef, resource)).map {
        case Exclude => ResolverReport.failed(resolver.id, projectRef -> ResourceIsDeprecated(ref.original, projectRef)) -> None
        case Include => ResolverReport.success(resolver.id, projectRef) -> Some(resource)
      }
    }
```

imsdu:
> > * This feature can be interesting in other cases such as context resolution when creating/updating resources
> 
> If no other feature is using it right now I would say YAGNI

I don't see why we would exclude deprecated schemas from imports and not deprecated contexts in resources.

> 
> > * Schema imports does not know how to resolve and fetch schemas and ontologies, it would not be able to fetch
> 
> the latest version to check the deprecation status We could give it the ability to check the deprecation status of the latest revision
> 
> It's just an option. Because this feels like quite a complex solution for what on the face of it seems like a simple problem, so I think it's worth making sure we want to pay that price

It is not so easy because resolvers are involved and schemas imports role is limited to accumulate imports or errors

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SchemasSpec.scala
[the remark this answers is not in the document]
imsdu: This is because we accumulate errors here, we can have an error because an import is deprecated and another because another one is not found at all.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resolvers/ResolverResolution.scala
imsdu: It does not look really clearer and it will have an impact on the cross-project resolver side

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SchemasSpec.scala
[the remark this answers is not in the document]
imsdu: The information is there, you can see it in the diff mentioning the report that is included in the payload.
I have also added a check to look for this specific error in the report

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resolvers/ResolverResolution.scala
olivergrabinski: I don't fully understand the impact the change would have on cross-project resolver 🤔

shinyhappydan: Obviously this is all a matter of opinion, but the ways in which I would see it as objectively clearer:
- No comments needed
- Boolean matching removed for clearer case objects
- `resourceOpt` is only used when it is Some, this is fixed
- shouldExcludeDueToDeprecation is clearer than runDeprecationCheck as it is explaining the check
- Traversal is maybe not a big deal but it is a complication

I think we can move on for this PR, but the things I have mentioned are things which in general make the codebase we have difficult to understand/navigate. They are all small things by themselves, but they add up. 

I'm just posting this to attempt to express how I feel about it, I think it's fine to disagree

shinyhappydan (approved):

PR #4394, opened 2023-10-20
Log failed IO that are in a Kamon span

### LINKED ISSUE #4393: Log failed IO that are in a Kamon span
**Motivation**
In order to better understand what goes wrong and see if any improvements can be made on Delta, log the failed IOs that are in Kamon spans.

olivergrabinski:
Closes #4393


olivergrabinski:
The non-CE `KamonMonitoring` is not concerned by this. Should it be? There will be more logs as the migration goes on.

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/kamon/KamonMonitoringCats.scala
```diff
@@ -62,14 +62,15 @@ object KamonMonitoringCats {
       component: String,
       tags: Map[String, Any] = Map.empty,
       takeSamplingDecision: Boolean = true
-  )(io: IO[A]): IO[A] =
+  )(io: IO[A]): IO[A] = {
     if (enabled)
       buildSpan(name, component, tags).bracketCase(_ => io) {
         case (span, ExitCase.Completed)    => finishSpan(span, takeSamplingDecision)
         case (span, ExitCase.Error(cause)) => failSpan(span, cause, takeSamplingDecision)
```
imsdu, line 69: The log may be better here ?

olivergrabinski: I thought it would still make sense to log the errors even if Kamon is disabled 🤔

imsdu (dismissed):

imsdu (approved):

imsdu (approved):

imsdu (dismissed):

PR #4409, opened 2023-10-23
Migrate acls to Cats Effect

### LINKED ISSUE #4362: Migrate Acls to Cats Effect

imsdu:
Fixes #4362


delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/mu/bio/BioSuite.scala
```diff
@@ -14,6 +15,7 @@ abstract class BioSuite
     with BioAssertions
     with BIOValues
     with BIOStreamAssertions
+    with CatsIOValues
```
shinyhappydan, line 18: Is this definitely good? I guess I can't see a problem that would occur

shinyhappydan (approved):

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/mu/bio/BioSuite.scala
imsdu: It is also temporary, the thing is acl check now returns a CatsIO and is used by some module that are not yet migrated

PR #4418, opened 2023-10-24
Add an endpoint to change only the schema of a resource

### LINKED ISSUE #4344: Add an endpoint to change only the schema of a resource
**Motivation**
With #4343, it will be possible to change the schema of a resource.

However, it requires to also update the resource itself and we also need an endpoint to update only the schema

**Acceptance criteria**
1. When a user call, this endpoint, Delta attempts to validate the resource with this new schema
2. If the validation is successful, the resource is associated to the new schema with a dedicated event

olivergrabinski:
Closes #4344

Screenshot 2023-10-25 at 13 30 19

Screenshot 2023-10-24 at 12 20 58

### DIFF
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
@@ -13,8 +13,8 @@ import ch.epfl.bluebrain.nexus.delta.rdf.utils.JsonKeyOrdering
 import ch.epfl.bluebrain.nexus.delta.routes.ResourcesRoutes.asSourceWithMetadata
 import ch.epfl.bluebrain.nexus.delta.sdk._
 import ch.epfl.bluebrain.nexus.delta.sdk.acls.AclCheck
-import ch.epfl.bluebrain.nexus.delta.sdk.circe.CirceUnmarshalling
 import ch.epfl.bluebrain.nexus.delta.sdk.ce.DeltaDirectives._
+import ch.epfl.bluebrain.nexus.delta.sdk.circe.CirceUnmarshalling
 import ch.epfl.bluebrain.nexus.delta.sdk.directives.{AuthDirectives, DeltaSchemeDirectives}
 import ch.epfl.bluebrain.nexus.delta.sdk.fusion.FusionConfig
 import ch.epfl.bluebrain.nexus.delta.sdk.identities.Identities
@@ -24,7 +24,7 @@ import ch.epfl.bluebrain.nexus.delta.sdk.model.routes.Tag
 import ch.epfl.bluebrain.nexus.delta.sdk.model.{BaseUri, ResourceF}
 import ch.epfl.bluebrain.nexus.delta.sdk.permissions.Permissions.resources.{read => Read, write => Write}
 import ch.epfl.bluebrain.nexus.delta.sdk.resources.NexusSource.DecodingOption
-import ch.epfl.bluebrain.nexus.delta.sdk.resources.model.ResourceRejection.{InvalidJsonLdFormat, InvalidSchemaRejection, ResourceNotFound}
+import ch.epfl.bluebrain.nexus.delta.sdk.resources.model.ResourceRejection.{IdenticalSchema, InvalidJsonLdFormat, InvalidSchemaRejection, ResourceNotFound}
 import ch.epfl.bluebrain.nexus.delta.sdk.resources.model.{Resource, ResourceRejection}
 import ch.epfl.bluebrain.nexus.delta.sdk.resources.{NexusSource, Resources}
 import ch.epfl.bluebrain.nexus.delta.sourcing.model.ProjectRef
@@ -170,6 +170,22 @@ final class ResourcesRoutes(
                             )
                           }
                         )
+                      },
+                      (pathPrefix("updateSchema") & put & pathEndOrSingleSlash) {
+                        authorizeFor(ref, Write).apply {
+
+                          emit(
+                            IO.fromOption(schemaOpt)(IdenticalSchema())
+                              .flatMap { schema =>
+                                resources
+                                  .updateResourceSchema(id, ref, schema)
+                                  .flatTap(indexUIO(ref, _, mode))
+                              }
+                              .attemptNarrow[ResourceRejection]
+                              .rejectWhen(wrongJsonOrNotFound)
+                          )
+
+                        }
                       },
                       (pathPrefix("refresh") & put & pathEndOrSingleSlash) {
                         authorizeFor(ref, Write).apply {
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
@@ -178,7 +178,7 @@ final class ResourcesRoutes(
                             IO.fromOption(schemaOpt)(NoSchemaProvided)
                               .flatMap { schema =>
                                 resources
-                                  .updateResourceSchema(id, ref, schema)
+                                  .updateAttachedSchema(id, ref, schema)
                                   .flatTap(indexUIO(ref, _, mode))
                               }
                               .attemptNarrow[ResourceRejection]
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
@@ -24,7 +24,7 @@ import ch.epfl.bluebrain.nexus.delta.sdk.model.routes.Tag
 import ch.epfl.bluebrain.nexus.delta.sdk.model.{BaseUri, ResourceF}
 import ch.epfl.bluebrain.nexus.delta.sdk.permissions.Permissions.resources.{read => Read, write => Write}
 import ch.epfl.bluebrain.nexus.delta.sdk.resources.NexusSource.DecodingOption
-import ch.epfl.bluebrain.nexus.delta.sdk.resources.model.ResourceRejection.{IdenticalSchema, InvalidJsonLdFormat, InvalidSchemaRejection, ResourceNotFound}
+import ch.epfl.bluebrain.nexus.delta.sdk.resources.model.ResourceRejection.{InvalidJsonLdFormat, InvalidSchemaRejection, NoSchemaProvided, ResourceNotFound}
 import ch.epfl.bluebrain.nexus.delta.sdk.resources.model.{Resource, ResourceRejection}
 import ch.epfl.bluebrain.nexus.delta.sdk.resources.{NexusSource, Resources}
 import ch.epfl.bluebrain.nexus.delta.sourcing.model.ProjectRef
@@ -175,7 +175,7 @@ final class ResourcesRoutes(
                         authorizeFor(ref, Write).apply {
 
                           emit(
-                            IO.fromOption(schemaOpt)(IdenticalSchema())
+                            IO.fromOption(schemaOpt)(NoSchemaProvided())
                               .flatMap { schema =>
                                 resources
                                   .updateResourceSchema(id, ref, schema)
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
@@ -171,7 +171,7 @@ final class ResourcesRoutes(
                           }
                         )
                       },
-                      (pathPrefix("updateSchema") & put & pathEndOrSingleSlash) {
+                      (pathPrefix("update-schema") & put & pathEndOrSingleSlash) {
                         authorizeFor(ref, Write).apply {
 
                           emit(
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
@@ -175,7 +175,7 @@ final class ResourcesRoutes(
                         authorizeFor(ref, Write).apply {
 
                           emit(
-                            IO.fromOption(schemaOpt)(NoSchemaProvided())
+                            IO.fromOption(schemaOpt)(NoSchemaProvided)
                               .flatMap { schema =>
                                 resources
                                   .updateResourceSchema(id, ref, schema)
--- a/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
+++ b/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
@@ -351,7 +351,7 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap with CatsIOValues
 
     "fail to update a schema without resources/write permission" in {
       aclCheck.subtract(AclAddress.Root, Anonymous -> Set(resources.write)).accepted
-      Put(s"/v1/resources/$projectRef/_/someId/updateSchema") ~> routes ~> check {
+      Put(s"/v1/resources/$projectRef/_/someId/update-schema") ~> routes ~> check {
         response.status shouldEqual StatusCodes.Forbidden
         response.asJson shouldEqual jsonContentOf("errors/authorization-failed.json")
       }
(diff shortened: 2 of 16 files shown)


delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -346,6 +349,51 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap with CatsIOValues
...
+    "fail to update schema when providing _ as schema" in {
+      thereIsAResourceWithSchema { id =>
+        Put(s"/v1/resources/$projectRef/_/$id/updateSchema") ~> routes ~> check {
+          response.status shouldEqual StatusCodes.BadRequest
+          response.asJson.hcursor.get[String]("@type").toOption should contain("IdenticalSchema")
+        }
+      }
+    }
+
+    "fail to update schema when providing current schema" in {
```
olivergrabinski, line 379: The remaining question is whether we want this behavior or not

shinyhappydan: I don't think this provides any value to the user. 

A legitimate use case might be a user wanting to bulk update schemas, not knowing which ones might already be pointed at it already

If a user performed an update on a resource with an identical payload we wouldn't fail. We shouldn't fail here either. 

Not having this check is good for Delta because it's less code

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -171,6 +171,22 @@ final class ResourcesRoutes(
                           }
                         )
                       },
+                      (pathPrefix("updateSchema") & put & pathEndOrSingleSlash) {
+                        authorizeFor(ref, Write).apply {
+
+                          emit(
+                            IO.fromOption(schemaOpt)(IdenticalSchema())
```
shinyhappydan, line 178: I would say the error here is that the schema is not specified rather than it being the same

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -346,6 +349,51 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap with CatsIOValues
...
+    }
+
+    "succeed to update the schema" in {
+      thereIsAResourceWithSchema { id =>
+        Put(s"/v1/resources/$projectRef/otherSchema/$id/updateSchema") ~> routes ~> check {
+          response.status shouldEqual StatusCodes.OK
+          response.asJson.hcursor.get[String]("_constrainedBy").toOption should contain(schema3.id.toString)
+        }
+      }
+    }
```
shinyhappydan, line 395: I think we need to assert that the side-effect happened

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
```diff
@@ -346,6 +349,51 @@ class ResourcesRoutesSpec extends BaseRouteSpec with IOFromMap with CatsIOValues
...
+      thereIsAResourceWithSchema { id =>
+        Put(s"/v1/resources/$projectRef/myschema/$id/updateSchema") ~> routes ~> check {
+          response.status shouldEqual StatusCodes.BadRequest
+          response.asJson.hcursor.get[String]("@type").toOption should contain("IdenticalSchema")
+        }
+      }
+    }
+
+    "succeed to update the schema" in {
+      thereIsAResourceWithSchema { id =>
```
shinyhappydan, line 389: I think both this and the previous test would read better if you pass the schema name to `thereIsAResourceWithSchema`

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -171,6 +171,22 @@ final class ResourcesRoutes(
                           }
                         )
                       },
+                      (pathPrefix("updateSchema") & put & pathEndOrSingleSlash) {
```
imsdu, line 174: `update-schema` ?
Not sure if we have already occurrences of path where we have this case

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/model/ResourceRejection.scala
```diff
@@ -198,6 +198,12 @@ object ResourceRejection {
     */
   final case class SchemaIsDeprecated(schemaId: Iri) extends ResourceRejection(s"Schema '$schemaId' is deprecated.")
 
+  /**
+    * Rejection returned when attempting to update the resource with the schema it is already constrained with.
+    */
+  final case class IdenticalSchema()
```
shinyhappydan, line 204: can be a case object

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImplSpec.scala
```diff
@@ -443,6 +443,47 @@ class ResourcesImplSpec
...
+      "reject if the project doesn't exist" in {
+        resources
+          .updateResourceSchema(id, nonExistentProject, schema3.id)
+          .rejectedWith[ProjectContextRejection]
+      }
+
+      "succeed" in {
+        val updated = resources.updateResourceSchema(id, projectRef, schema3.id).accepted
+        updated.schema.iri shouldEqual schema3.id
+      }
```
shinyhappydan, line 483: I think again we need to assert on the side-effect taking place

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -90,6 +90,22 @@ trait Resources {
...
+    * Update the schema of an existing resource
+    *
+    * @param id
+    *   identifier that will be expanded to the iri of the resource
+    * @param projectRef
+    *   project reference where the resource belongs
+    * @param schema
+    *   identifier of the new schema that will be used to validate the resource. This identifier will be expanded
+    */
+  def updateResourceSchema(
```
imsdu, line 103: `updateAttachedSchema` ?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImpl.scala
```diff
@@ -79,6 +80,22 @@ final class ResourcesImpl private (
...
+      schema: IdSegment
+  )(implicit caller: Caller): IO[DataResource] = {
+    for {
+      projectContext <- fetchContext.onModify(projectRef).toCatsIO
+      iri            <- expandIri(id, projectContext).toCatsIO
+      schemaRef      <- expandResourceRef(schema, projectContext)
+      resource       <- log.stateOr(projectRef, iri, ResourceNotFound(iri, projectRef)).toCatsIO
+      _              <- IO.raiseWhen(schemaRef.iri == resource.schema.iri)(IdenticalSchema())
+      jsonld         <- sourceParser(projectRef, projectContext, iri, resource.source).toCatsIO
+      res            <- eval(UpdateResource(iri, projectRef, schemaRef.some, resource.source, jsonld, resource.rev, caller))
```
imsdu, line 95: It would be nice to have a dedicated command and a dedicated event.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImpl.scala
```diff
@@ -79,6 +80,22 @@ final class ResourcesImpl private (
...
+      projectRef: ProjectRef,
+      schema: IdSegment
+  )(implicit caller: Caller): IO[DataResource] = {
+    for {
+      projectContext <- fetchContext.onModify(projectRef).toCatsIO
+      iri            <- expandIri(id, projectContext).toCatsIO
+      schemaRef      <- expandResourceRef(schema, projectContext)
+      resource       <- log.stateOr(projectRef, iri, ResourceNotFound(iri, projectRef)).toCatsIO
+      _              <- IO.raiseWhen(schemaRef.iri == resource.schema.iri)(IdenticalSchema())
+      jsonld         <- sourceParser(projectRef, projectContext, iri, resource.source).toCatsIO
```
imsdu, line 94: We should not have to parse the resource again

imsdu (changes requested):

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
olivergrabinski: My thinking is that when updating resources `_` means keep the same schema. I am reworking things a bit anyway

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/model/ResourceRejection.scala
shinyhappydan: You might have to put `rejectctedWith[IdenticalSchema.type]`, this can be gotten around by making an empty trait also called IdenticalSchema. I can't say whether it's worth it but it's possible anyway

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
olivergrabinski: I am not sure if this is the right way to think about it, but I thought here we are just checking that the response to the new endpoint is as expected. The whole thing including the fetch is done in integration?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
olivergrabinski: Is this a suggestion just for this method or also for the event/command name?

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
shinyhappydan: I think it's unclear what the `*RoutesSpec` should be testing in general, but whether unit or integration, it feels like the tests need to cover
1. parsing request
2. routing of request
3. serialisation of response

I would say that 2 is where the side effect test requirement comes from

shinyhappydan: It's definitely fair to say this is likely covered in integration tests. I think our test purposes aren't clear

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/model/ResourceRejection.scala
olivergrabinski: I have removed the test since it was not needed so now it's just a case object :macgyver:

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
olivergrabinski: I changed the rejection name

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
imsdu: Just for this method

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
olivergrabinski: I'm resolving this thread because I have removed this test and the case where the schema is identical is handled by going to a fetch operation which is already testes.

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
olivergrabinski: If the routes were dummy routes, this test should still pass. But whether the second part (fetching the resource after the update) is working depends on the `resources` implementation that's passed to the routes in the test, and that is something that is tested elsewhere

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/model/ResourceRejection.scala
olivergrabinski: Resolving this since it's no longer relevant

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
olivergrabinski: Resolving this since the name changed

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
olivergrabinski: I was thinking about what I did in another test class when I read your comment. Pushing changes now

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
shinyhappydan: The issue is that you're not testing that the resources implementation was actually called. But given the state of our routes tests I think maybe we can just move on

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/model/ResourceEvent.scala
```diff
@@ -294,12 +305,13 @@ object ResourceEvent {
...
-            case _: ResourceTagAdded   => Tagged
-            case _: ResourceTagDeleted => TagDeleted
-            case _: ResourceDeprecated => Deprecated
+            case _: ResourceCreated       => Created
+            case _: ResourceUpdated       => Updated
+            case _: ResourceRefreshed     => Refreshed
+            case _: ResourceTagAdded      => Tagged
+            case _: ResourceTagDeleted    => TagDeleted
+            case _: ResourceDeprecated    => Deprecated
+            case _: ResourceSchemaUpdated => Updated
```
olivergrabinski, line 314: This mapping felt correct to me

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutesSpec.scala
olivergrabinski: Maybe we can discuss about some guidelines to clarify these situations

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -382,6 +403,16 @@ object Resources {
       // format: on
     }
 
+    def updateResourceSchema(u: UpdateResourceSchema) = {
+      for {
+        s                          <- stateWhereResourceIsEditable(u)
+        schemaRef                  <- IO.fromOption(u.schemaOpt)(???)
```
imsdu, line 409: ???

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/model/ResourceEvent.scala
imsdu: Not sure but with the last changes (including tagging at creation for example), we will have to rework this part.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
olivergrabinski: Ah, this option will actually never be empty, it's just an option because that's what `ModifyCommand` expects. Weird it slipped through the cracks

olivergrabinski: I tried to address it gracefully, cf. new commits

imsdu (approved):

PR #4419, opened 2023-10-24
Migrate composite views to Cats Effect

### LINKED ISSUE #4329: Migrate composite views module to CE
In a first PR, migrate only the state machine / the routes part

imsdu:
Fixes #4329

### DIFF
--- a/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
+++ b/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
@@ -1,11 +1,11 @@
 package ch.epfl.bluebrain.nexus.delta.plugins.compositeviews
 
-import cats.effect.Clock
+import cats.effect.{Clock, IO}
 import ch.epfl.bluebrain.nexus.delta.kernel.effect.migration.ioToTaskK
 import ch.epfl.bluebrain.nexus.delta.kernel.kamon.KamonMetricComponent
 import ch.epfl.bluebrain.nexus.delta.kernel.search.Pagination.FromPagination
 import ch.epfl.bluebrain.nexus.delta.kernel.syntax.kamonSyntax
-import ch.epfl.bluebrain.nexus.delta.kernel.utils.{IOUtils, UUIDF}
+import ch.epfl.bluebrain.nexus.delta.kernel.utils.{IOInstant, UUIDF}
 import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.CompositeViews._
 import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.config.CompositeViewsConfig
 import ch.epfl.bluebrain.nexus.delta.plugins.compositeviews.indexing.CompositeViewDef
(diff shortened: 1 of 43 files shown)


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
```diff
@@ -45,24 +45,10 @@ final class CompositeViews private (
...
 
-  /**
-    * Create a new composite view with a generate id.
-    *
-    * @param project
-    *   the parent project of the view
-    * @param value
-    *   the view configuration
-    */
-  def create(
```
imsdu, line 60: Dead code

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
```diff
@@ -96,11 +81,11 @@ final class CompositeViews private (
...
     */
-  def create(project: ProjectRef, source: Json)(implicit caller: Caller): IO[CompositeViewRejection, ViewResource] = {
+  def create(project: ProjectRef, source: Json)(implicit caller: Caller): IO[ViewResource] = {
     for {
-      pc           <- fetchContext.onCreate(project)
-      (iri, value) <- sourceDecoder(project, pc, source)
-      res          <- eval(CreateCompositeView(iri, project, value, source.removeAllKeys("token"), caller.subject, pc.base))
+      pc           <- toCatsIO(fetchContext.onCreate(project))
+      (iri, value) <- toCatsIO(sourceDecoder(project, pc, source))
+      res          <- eval(CreateCompositeView(iri, project, value, source, caller.subject, pc.base))
```
shinyhappydan, line 88: Just want to check that the `removeAllKeys("token")` is correct to remove?

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/mu/ce/CatsEffectAssertions.scala
```diff
@@ -189,6 +194,24 @@ trait CatsEffectMUnitAssertions { self: Assertions =>
...
 
+    def eventually(expected: A, retryWhen: Throwable => Boolean)(implicit patience: PatienceConfig): IO[Unit] = {
+      val strategy = RetryStrategy[Throwable](
+        MaximumCumulativeDelayConfig(patience.timeout, patience.interval),
+        retryWhen,
+        onError = (_, _) => UIO.unit
+      )
+      assertEquals(expected, patience.timeout)
+        .retry(strategy)
+    }
```
shinyhappydan, line 205: I wonder if it's more flexible to pass an assertion here. 

Perhaps in order to keep it both flexible and consistent: 

- This could be renamed `eventuallyEquals`
- A "generic" assertion could be called `eventually`

I wonder if actually you could achieve syntax like this:

`doAnOperation.assertEquals(1).eventually` ? With an implicit class taking `IO[Assertion]`

Of course this can be done another day but it's interesting

shinyhappydan (approved):

delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeViews.scala
imsdu: Well, there is no token anymore in the payload, this was just a way so as not to save it unencrypted.

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/mu/ce/CatsEffectAssertions.scala
imsdu: Maybe, I did not want to modify that now as it will make it harder to migrate the other tests depending on it.
And we will to go back at it when we migrate the retry strategy (there is still a `UIO.unit` here)

PR #4424, opened 2023-10-25
Add pre and post synaptic pathways to BG -> ES projection

shinyhappydan:


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/CompositeSink.scala
```diff
@@ -123,6 +123,7 @@ final class Batch[SinkFormat](
   /** Replaces the graph of a provided [[GraphResource]] by extracting its new graph from the provided (full) graph. */
   private def replaceGraph(gr: GraphResource, fullGraph: Graph) = {
     implicit val api: JsonLdApi = JsonLdJavaApi.lenient
+    implicit val options        = JsonLdOptions.AlwaysEmbed
```
olivergrabinski, line 126: I think putting it here means it would only apply to the `BatchSink` and not the `SingleSink`.

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/GraphResourceToDocument.scala
```diff
@@ -33,8 +33,9 @@ final class GraphResourceToDocument(context: ContextValue, includeContext: Boole
 
   /** Given a [[GraphResource]] returns a JSON-LD created from the merged graph and metadata graph */
   def graphToDocument(element: GraphResource): Task[Option[Json]] = {
-    val graph = element.graph ++ element.metadataGraph
-    val json  =
+    implicit val jsonLdOptions = JsonLdOptions.AlwaysEmbed
```
olivergrabinski, line 36: Do we need to set this here? Is it not possible to make it an argument to the whole class, so that we can set it to be `@always` only from the CompositeSink?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/ResourceShift.scala
```diff
@@ -89,9 +89,10 @@ abstract class ResourceShift[State <: ScopedState, A, M](
   private def toGraphResource(project: ProjectRef, resource: ResourceF[A])(implicit
       cr: RemoteContextResolution
   ): IO[GraphResource] = {
-    val content  = resourceToContent(resource)
-    val metadata = content.metadata
-    val id       = resource.resolvedId
+    implicit val jsonLdOptions = JsonLdOptions.AlwaysEmbed
```
olivergrabinski, line 92: Similar question, would this not have potential consequences outside of composite views?

tests/src/test/resources/kg/search/synapse.json
```diff
@@ -0,0 +1,93 @@
...
+  "@context": "https://bbp.neuroshapes.org",
+  "@id": "https://bbp.epfl.ch/data/synapse",
+  "@type": [
+    "Entity",
+    "Parameter",
+    "SchafferCollateralAnatomyParameter",
+    "SynapsesPerConnection",
+    "ExperimentalSynapsesPerConnection"
+  ],
+  "brainLocation": {
```
olivergrabinski, line 11: Minor but there is no real use of having most of these fields here, `synapticPathway` should be enough

tests/src/test/resources/kg/search/synapse.json
```diff
@@ -0,0 +1,93 @@
...
+    "weight": {
+      "maxValue": 300,
+      "minValue": 200,
+      "unitCode": "grams"
+    }
+  },
+  "synapticPathway": {
+    "postSynaptic": [
+      {
+        "@id": "http://api.brain-map.org/api/v2/data/Structure/454",
```
imsdu, line 78: I think one can appear in both no ?
It would be nice to test that and make sure it works fine

shinyhappydan: I'll add another test, it probably makes sense to test that they can be both the same and different

tests/src/test/resources/kg/search/synapse.json
shinyhappydan: Can do, I guess you are right it isn't needed

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/GraphResourceToDocument.scala
shinyhappydan: Let me have a think about this one

shinyhappydan: I did what you suggested and it works. I think given that this is the smallest possible scope of change, you're right

olivergrabinski (approved):

PR #4425, opened 2023-10-25
Migrate Storages to Cats Effect

shinyhappydan:


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
```diff
@@ -87,11 +87,11 @@ final class Storages private (
...
       projectRef: ProjectRef,
       source: Json
-  )(implicit caller: Caller): IO[StorageRejection, StorageResource] = {
+  )(implicit caller: Caller): IO[StorageResource] = {
     for {
-      pc            <- fetchContext.onCreate(projectRef)
-      iri           <- expandIri(id, pc)
-      storageFields <- sourceDecoder(projectRef, pc, iri, source)
+      pc            <- fetchContext.onCreate(projectRef).toCatsIO
+      iri           <- expandIri(id, pc).toCatsIO
```
dantb, line 93: Could potentially reuse `FileId` in these methods - it encapsulates the Iri expansion and CE conversion. Might need a better name though 😅

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
```diff
@@ -242,11 +242,13 @@ final class Storages private (
     * @param project
     *   the project where the storage belongs
     */
-  def fetch[R](
+  def fetch[R <: Throwable](
       resourceRef: ResourceRef,
       project: ProjectRef
-  )(implicit rejectionMapper: Mapper[StorageFetchRejection, R]): IO[R, StorageResource] =
-    fetch(IdSegmentRef(resourceRef), project).mapError(rejectionMapper.to)
+  )(implicit rejectionMapper: Mapper[StorageFetchRejection, R]): IO[StorageResource] =
```
dantb, line 248: Do we still need this `R` and mapper? Looks like all it does is wrap the exception in a `WrappedStorageRejection` and is called only from `Files`

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
```diff
@@ -311,21 +316,23 @@ final class Storages private (
...
-        .hideErrors >> UIO.unit
+        .drain >> UIO.unit
     }
 
-  private def logFailureAndContinue[A](io: IO[StorageRejection, A]): UIO[Unit] =
-    io.mapError(err => logger.warn(err.reason)).attempt >> UIO.unit
+  private def logFailureAndContinue[A](io: IO[A]): IO[Unit] = {
+    io.onError { case err: StorageRejection =>
+      logger.warn(err.reason)
+    }.attemptNarrow[StorageRejection] >> IO.unit
```
dantb, line 325: `.void` also works

dantb (commented):
Realised this is probably still WIP as things like `SaveFile` aren't migrated - will leave the current comments anyway 😄

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
shinyhappydan: Possibly not, but I've been trying to just migrate rather than change the logic. I wonder if @imsdu might know if this is safe

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
shinyhappydan: Right, I knew there was a method for this. I looked for `.unit`

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
shinyhappydan: It's possible that we could have a broad trait (something suggesting a "full" id) and then specific classes like FileId extending it? I'm not sure whether to do that just for this though

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
imsdu: We should try to remove those but this can be done in later PRs (unless they are blocking in some way), this one is gonna be complex enough.

imsdu: Can we open a ticket for it ?
So that we can address that after the current migration ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
dantb: Yeah agreed - just working around `Resources` and a similar thing is done everywhere. I'll see what refactoring more generally would look like

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
```diff
@@ -111,10 +110,10 @@ final class Storages private (
       id: IdSegment,
       projectRef: ProjectRef,
       storageFields: StorageFields
-  )(implicit caller: Caller): IO[StorageRejection, StorageResource] = {
+  )(implicit caller: Caller): IO[StorageResource] = {
     for {
-      pc    <- fetchContext.onCreate(projectRef)
-      iri   <- expandIri(id, pc)
+      pc    <- fetchContext.onCreate(projectRef).toCatsIO
+      iri   <- expandIri(id, pc).toCatsIO
```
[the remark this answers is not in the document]
dantb: As an aside this `ExpandIri` could definitely be refactored, probably once the migration is done.  It doesn't do any actual effects - I prefer using less powerful types like `Either`/`Option` with the translation to the effect system delayed until the call sites. It makes things easier to understand / refactor / test IMO! E.g. in this case all it's doing is moving between data types.

This might be a larger larger discussion though? Sort of a stylistic preference

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/DiskStorageAccess.scala
```diff
@@ -1,28 +1,28 @@
...
-    def failWhen(condition: Boolean, err: => String) =
-      if (condition) IO.raiseError(StorageNotAccessible(id, err))
-      else IO.unit
+    def failWhen(condition: Boolean, err: => String) = {
+      IO.raiseWhen(condition)(StorageNotAccessible(id, err))
+    }
 
     for {
-      exists      <- UIO.delay(Files.exists(storage.volume.value))
+      exists      <- IO.delay(Files.exists(storage.volume.value))
```
dantb, line 21: There's a nice functional wrapper lib for file operations  - https://github.com/akiomik/cats-nio-file

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3StorageLinkFile.scala
```diff
@@ -5,49 +5,53 @@ import akka.http.scaladsl.model.Uri
...
-                dig,
-                origin = Storage
-              )
-            }
-          case None                 => Future.failed(fileNotFoundException)
-        }
-    ).mapError {
+    IO.fromFuture(
+      IO.delay(
+        S3.download(storage.value.bucket, URLDecoder.decode(key.toString, UTF_8.toString))
```
dantb, line 34: Another nice little wrapper library for S3 operations - https://github.com/laserdisc-io/fs2-aws

Not expecting you to go use all of these now haha 😛 was the S3 integrations one of the things we were going to ditch in the long term?

dantb (approved):
LGTM. The only change which may be worth doing is the `currentStorages` thing, it should be quite simple AFAICT

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StorageDeletionTask.scala
```diff
@@ -29,40 +33,43 @@ final class StorageDeletionTask(currentStorages: ProjectRef => Stream[Task, Stor
...
     new StorageDeletionTask(project =>
-      storages.currentStorages(project).evalMapFilter {
-        _.map(_.value).toTask
-      }
+      storages
+        .currentStorages(project)
+        .evalMapFilter {
+          _.map(_.value).toTask
+        }
+        .translate(taskToIoK)
```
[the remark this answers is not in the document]
shinyhappydan: I'll try this

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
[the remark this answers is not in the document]
shinyhappydan: Yeah I suppose I'm not clear why it's an IO

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/DiskStorageAccess.scala
shinyhappydan: I think that's a nice change but I'd rather do it in a separate PR, this is is already taking forever 😓

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3StorageLinkFile.scala
shinyhappydan: I think S3 is dead, yes

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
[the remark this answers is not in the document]
shinyhappydan: I agree the change you made was nice but I'm going to be lazy and not fix this

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
shinyhappydan: https://github.com/BlueBrain/nexus/issues/4441

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/DiskStorageAccess.scala
dantb: Yeah I got the easy part migrating files 😂

dantb (approved):

PR #4426, opened 2023-10-26
Fix bug where filename in location could differ from `_filename` attribute

### LINKED ISSUE #4379: Mismatch between _filename and _location in File resource metadata when special character like [ or ] are present in the file name
**Describe the bug**
There is a mismatch between _filename and _location property values in File resource metadata when special character like '[' or ']' are present in the file name. For example, a file named "[PH]layer_1.nrrd" is registered: 
- _filename = "[PH]layer_1.nrrd"  
- _location = "file:///.../a/%5BPH%5Dlayer_1.nrrd"
Here is an [example file  resource](https://bbp.epfl.ch/nexus/v1/files/bbp/atlas/https:%2F%2Fbbp.epfl.ch%2Fdata%2Fbbp%2Fatlas%2F5d8d6060-7a11-4166-85c0-a5f69a630f49)
The filename on remote disk is encoded while the value of _filename in the metadata is kept as is. This can cause issues for clients relying on _filename to download the file.

- [ ] Delta

**To Reproduce**
Steps to reproduce the behavior, e.g. for Fusion:
1. Create a file with name "[PH]layer_1.nrrd" in a project with remote storage configured as default.
2. Check the file metadata

**Expected behavior**
_filename should be equal to the file name on the remote disk

olivergrabinski:
Fixes #4379

If a file contained special characters, then in the local storage the file would be saved using akka.Uri.Path which had URI encoded characters in it. Hence a filename like `[tag]file.txt` would be saved as `%5Btag%5Dfile.txt` on disk.

### DIFF
--- a/delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FilesSpec.scala
+++ b/delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FilesSpec.scala
@@ -43,6 +43,8 @@ import monix.execution.Scheduler
 import org.scalatest.DoNotDiscover
 import org.scalatest.concurrent.Eventually
 
+import java.net.URLDecoder
+
 @DoNotDiscover
 class FilesSpec(docker: RemoteStorageDocker)
     extends TestKit(ActorSystem("FilesSpec"))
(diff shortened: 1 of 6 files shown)


delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FilesSpec.scala
```diff
@@ -161,6 +163,18 @@ class FilesSpec(docker: RemoteStorageDocker)
         actual shouldEqual expected
       }
 
+      "succeed when the file has special characters" in {
+        val specialFileName = "-._~:?#[ ]@!$&'()*,;="
```
olivergrabinski, line 167: Note that:
1. I believe there is still a problem if a filename contains `%`
2. If a `+` is in the filename, it will end up being a space when saved

imsdu (approved):

PR #4449, opened 2023-10-31
Migrate ElasticSearch routes and state machine to Cats Effect

### LINKED ISSUE #4408: Move Elasticsearch state machine and routes to Cats effect
Not in scope: Elasticsearch indexing process

dantb:
Fixes #4408

### DIFF
--- a/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/package.scala
+++ b/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/package.scala
@@ -2,7 +2,7 @@ package ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch
 
 import cats.effect.IO
 import ch.epfl.bluebrain.nexus.delta.kernel.utils.CatsEffectsClasspathResourceUtils
-import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.{nxv, schemas, contexts => nxvContexts}
+import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.{contexts => nxvContexts, nxv, schemas}
 import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceF
 import ch.epfl.bluebrain.nexus.delta.sdk.permissions.Permissions
 import ch.epfl.bluebrain.nexus.delta.sdk.permissions.model.Permission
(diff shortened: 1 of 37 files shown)


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/package.scala
```diff
@@ -52,37 +52,34 @@ package object model {
 
   implicit private val logger: Logger = Logger("ElasticSearchPlugin")
 
+  // TODO can do this only once and inject them where they're need instead of memoizing?
```
dantb, line 55: Got another PR incoming to do this TODO, thought it was worth splitting out

olivergrabinski (changes requested):
Looks OK, though I think you need to `attemptNarrow` on some endpoints like `_mapping` because it leads to a test failure otherwise

dantb:
> Looks OK, though I think you need to `attemptNarrow` on some endpoints like `_mapping` because it leads to a test failure otherwise

Thanks, also had forgotten the query routes - fixed in https://github.com/BlueBrain/nexus/pull/4449/commits/5e5e464c4cbfa6394962b80d47bdd65ae9899df0

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -264,8 +264,11 @@ object Resources {
     * Expands the segment to a [[ResourceRef]]
     */
   def expandResourceRef(segment: IdSegment, context: ProjectContext): IO[ResourceRef] =
+    expandResourceRef(segment, context, InvalidResourceId)
+
+  def expandResourceRef(segment: IdSegment, context: ProjectContext, notFound: String => Throwable): IO[ResourceRef] =
     IO.fromOption(segment.toIri(context.apiMappings, context.base).map(ResourceRef(_)))(
-      InvalidResourceId(segment.asString)
+      notFound(segment.asString)
     )
```
olivergrabinski, line 272: Ok for me but why the second method here?

olivergrabinski (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
dantb: Ah yeah, was going to reuse it in the ES module but forgot to call it - will open another PR as I think it's a bit bigger

PR #4451, opened 2023-10-31
Cache ElasticSearch files and inject into use sites

dantb:
We've spoken about this before, does a couple of things
- ~Reinstates the behaviour of Monix `memoize` for files loaded from resources in `FilesCache`~
- Loads ES files in at startup
- Injects these files in where they're needed


shinyhappydan:
Just for some context, why don't we load the files using distage and just pass the results to classes directly? 

Naively it feels like this would remove the need to cache at all, but there's probably something I'm missing

shinyhappydan:
Yeah now I'm looking on a big screen I think it would be possible to just load these files with distage and pass them into constructors, without the need for any caching. Each file is only used once and construction fails anyway if the file isn't there

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/FilesCache.scala
```diff
@@ -0,0 +1,28 @@
...
+}
+
+object FilesCache {
+
+  implicit val log: log4cats.Logger[IO] = Logger.cats[FilesCache]
+
+  def mk(fetchFromResources: String => IO[JsonObject])(implicit t: Timer[IO]): IO[FilesCache] =
+    MemoryCache.ofSingleImmutableMap[IO, String, JsonObject](None).map(mk(fetchFromResources, _))
+
+  def mk(fetchFromResources: String => IO[JsonObject], cache: MemoryCache[IO, String, JsonObject]): FilesCache =
```
shinyhappydan, line 23: I like that `mk` suggests construction rather than `apply`

dantb:
Yeah you're right @shinyhappydan, injecting properly makes the caching redundant. Will refactor

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ElasticSearchFiles.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  /** Settings for the event metrics index */
+  def metricsSettings: MetricsSettings
+}
+
+final case class DefaultMapping(value: JsonObject)
+final case class DefaultSettings(value: JsonObject)
+final case class EmptyResults(value: Json)
+final case class MetricsMapping(value: JsonObject)
+final case class MetricsSettings(value: JsonObject)
```
dantb, line 33: Could also use [monix newtypes](https://newtypes.monix.io/) and opaque types once we go to Scala 3

dantb: I agree it's not always right, [names are not type safety](https://lexi-lambda.github.io/blog/2020/11/01/names-are-not-type-safety/). But here it's pushing the potential mistake to one place (`ElasticSearchFiles`) rather than at the 5 use sites - I actually broke a test due to this 😂 (it's probably a me problem, but when I see two parameters of the same type alarm bells go off)

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/ce/CatsRunContext.scala
```diff
@@ -6,7 +6,7 @@ import scala.concurrent.ExecutionContext
 
 trait CatsRunContext {
 
-  implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
-  implicit val timer: Timer[IO]               = IO.timer(ExecutionContext.global)
+  implicit lazy val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
+  implicit lazy val timer: Timer[IO]               = IO.timer(ExecutionContext.global)
```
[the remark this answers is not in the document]
dantb: Think I had null pointer test failure, not sure if this was what resolved it though 🤔 I'll remove in the next PR (it depends on this one)

PR #4466, opened 2023-11-07
Remove Kamon monitoring for Monix

imsdu:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/acls/AclsImpl.scala
```diff
@@ -28,8 +28,7 @@ final class AclsImpl private (
   override def fetch(address: AclAddress): IO[AclResource] =
     log
       .stateOr(address, AclNotFound(address))
-      .toBIO[AclNotFound]
```
imsdu, line 31: Last part preventing us to remove the Monix implementation

dantb (approved):

PR #4467, opened 2023-11-07
Proposal: do not allow underscore to be used in operations that require a specific resolver

shinyhappydan:
When we treat underscore as a normal string segment, we can get bad error messages if a user provides underscore for something which requires a specific value. I have implemented a change on the ResolverRoutes, as that is where I first noticed the problem.

This solution definitely helps this route be more understandable, as well as ensuring that error messages returned to the user are understandable


dantb (commented):
Makes sense to me but don't know enough about this to approve. Could update the docs? Only if this is a behaviour users have been relying on (intentionally or not)

PR #4468, opened 2023-11-07
Remove DeltaDirectives for Monix

imsdu:
+ Migrating FileResponse to use Cats IO


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJsonLd.scala
```diff
@@ -104,8 +104,17 @@ object ResponseToJsonLd extends FileBytesInstances {
...
-        onSuccess(flattened.runToFuture) {
+        val flattened = io.flatMap {
+          _.traverse { fr =>
+            fr.content.map {
+              _.map { s =>
+                fr.metadata -> s
+              }
+            }
+          }
+        }
```
shinyhappydan, line 115: I don't think it's a problem for this PR, but I think this indicates a problem with our design. Perhaps it's just something that is fixed easily if we ditch akka

shinyhappydan (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/directives/ResponseToJsonLd.scala
imsdu: Yep, I did not want to spend too much time on it as I think we should try to get closer to what is expected in Akka/http4s.
Especially that now, we don't have 2 error channels.
The idea would be to get rid of the `attemptNarrow` too among other things

PR #4476, opened 2023-11-08
Migrate Transactors to Cats Effect

### LINKED ISSUE #4471: Migrate Transactors to Cats Effect

shinyhappydan:
Fixes #4471, #4474, #4475, #4430


delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/IOUtils.scala
```diff
@@ -16,20 +16,23 @@ trait IOUtils {
     */
   def instant(implicit clock: Clock[UIO]): UIO[Instant] =
     clock.realTime(TimeUnit.MILLISECONDS).map(Instant.ofEpochMilli)
+
+  implicit class IoOps[A](val io: IO[A]) {
```
imsdu, line 20: Not to fix in that PR but we have other places where we had operation to IO

delta/kernel/src/test/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/IOUtilsSuite.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import cats.effect.{Clock, IO}
+import cats.implicits.toFlatMapOps
+import ch.epfl.bluebrain.nexus.delta.kernel.utils.IOUtils.IoOps
+import munit.FunSuite
+
+import java.time.Instant
+import scala.concurrent.duration._
+
+class IOUtilsSuite extends FunSuite {
```
imsdu, line 11: We can use a CatsEffectsSuite here

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/mu/ce/CatsEffectAssertions.scala
```diff
@@ -230,7 +230,7 @@ trait CatsEffectAssertions { self: Assertions =>
       *   IO.unit.assert // OK
       * }}}
       */
-    def assert(implicit loc: Location): IO[Unit] =
+    def assertUnit(implicit loc: Location): IO[Unit] =
```
imsdu, line 233: This is coming from the Cats Effect Munit library.
When we use CE3, it is likely that we use directly the library and get rid of this class.
Could we keep the same name for now ?

delta/kernel/src/test/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/IOUtilsSuite.scala
shinyhappydan: We can't because it's not visible in `kernel`

delta/testkit/src/main/scala/ch/epfl/bluebrain/nexus/testkit/mu/ce/CatsEffectAssertions.scala
shinyhappydan: Can I merge this and rename again in another PR? I've got a backlog

imsdu (approved):

PR #4481, opened 2023-11-08
Add mapping for synaptic pathways

olivergrabinski:


tests/docker/config/mapping.json
```diff
@@ -99,6 +99,28 @@
         "endedAt": { "type": "date" }
       }
     },
+    "preSynapticPathway": {
+      "properties": {
+        "postSynaptic": {
+          "properties": {
+            "about": { "type": "text" },
```
imsdu, line 106: We want those to be only accessible via full text ?

PR #4483, opened 2023-11-08
Add bulk create by copy operation for files

### LINKED ISSUE #4400: Add a copy operation to files
**Motivation**
When moving a project from the private scope to the public scope, it is needed to make a copy (both of the file metadata and the physical file) from one project to another.

So far, they need to download the file to upload it again so it is slow and error prone.

**Acceptance criteria**
1. The selected file to copy can be a latest, revisioned and tagged version of a file
2. The copy is only possible from local to local or remote to remote
3. By default, the file is copied to the default storage of the target project but the user can also provide
4. The same constraints as the other write operations on files (write permission, enough space on the target storage)
5. The file metadata (digest/size/...) is preserved excluding the auditing ones (createdAt, createdBy, ...)
6. Only to be implemented for local and remote storages
7. The file can be optionally tagged like in the create/link operations

dantb:
Fixes #4400 with a bulk file copy operation.

This PR unfortunately got massive and I'd be happy to pair with someone to go through anything not clear.

Docs still to come


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutes.scala
```diff
@@ -232,6 +247,30 @@ final class FilesRoutes(
...
+    parameters("rev".as[Int], "storage".as[IdSegment].?, "tag".as[UserTag].?).tmap[FilePutRequest] {
+      case (rev, storage, tag) =>
+        FilePutRequest.Update(rev, storage, tag)
+    }
+
+  private def createFileOperation: Directive1[FilePutRequest] =
+    parameters("storage".as[IdSegment].?, "tag".as[UserTag].?).tmap[FilePutRequest] { case (storage, tag) =>
+      FilePutRequest.Create(storage, tag)
+    }
+
```
shinyhappydan, line 273: I really like the idea here, but I think this is maybe hiding a little too much? I'll try to explain better IRL

shinyhappydan (commented):
Regarding the permissions: I would imagine you should only need files/read on the source and files/write on the destination

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -197,6 +197,39 @@ final class Files(
     } yield res
   }.span("createLink")
 
+  // TODO comments
+  def copy(
+      id: FileId,
+      targetProject: ProjectRef,
+      targetStorageId: Option[IdSegment],
+      tagForCopy: Option[UserTag]
```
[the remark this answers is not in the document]
imsdu: The user can optionally provide a id for the file or a new one can be generated.
One of the reasons behind that is now most of the time, the org/project appears in the id so that an id is unique in a deployment so keeping the id would break that.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -197,6 +197,39 @@ final class Files(
...
+  )(implicit caller: Caller): IO[FileResource] = {
+    // TODO what if there's an ID clash? Generate one, or let the rejection happen as normal?
+    val newFileId = FileId(id.id.value, targetProject)
+    for {
+      (iri, targetProjectContext) <- newFileId.expandIri(fetchContext.onCreate)
+      (storageRef, storage)       <- fetchActiveStorage(targetStorageId, targetProject, targetProjectContext)
+      file                        <- fetch(id)
+      _                           <- validateStorageTypeForCopy(file.value.storageType, storage)
+      attributes                   = file.value.attributes
```
imsdu, line 214: A copy of the physical file must also be done both in the case of local storage and remote storage (that means having to change the storage app too)
We also need a new uuid for this new file

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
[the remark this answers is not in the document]
dantb: Right so something like `destinationFileId: Option[IdSegment]` passed into the endpoint?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -197,6 +197,60 @@ final class Files(
...
+  }.span("copyFile")
+
+  private def fetchSourceFile(id: FileId)(implicit c: Caller) =
+    for {
+      file                        <- fetch(id)
+      (iri, _)                    <- id.expandIri(fetchContext.onRead)
+      sourceStorage               <- storages.fetch(file.value.storage, id.project)
+      _                           <- validateAuth(id.project, sourceStorage.value.storageValue.readPermission)
+      attributes                   = file.value.attributes
+      sourceBytes                 <- fetchFile(sourceStorage.value, attributes, file.id)
```
imsdu, line 229: We should have to fetch the file itself to push it back.
For example for remote storage, it will mean get one file from one datacenter to another to push again to the first one.
Even for the local storage, we can just copy the file without having fetch the file to Delta.
Like we have a `SaveFile` and a `LinkFile`, we need a copy file to handle that.
That means also changes on the storage app to add a route for this copy operation

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/HttpClient.scala
```diff
@@ -139,25 +144,25 @@ class HttpClient private (baseUrl: Uri, httpExt: HttpExt)(implicit
...
 
-  def requestAssert[A](
+  def requestAssertAndReturn[A](
       method: HttpMethod,
       url: String,
       body: Option[Json],
       identity: Identity,
       extraHeaders: Seq[HttpHeader] = jsonHeaders
-  )(assertResponse: (A, HttpResponse) => Assertion)(implicit um: FromEntityUnmarshaller[A]): IO[Assertion] = {
+  )(assertResponse: (A, HttpResponse) => (A, Assertion))(implicit um: FromEntityUnmarshaller[A]): IO[(A, Assertion)] = {
```
shinyhappydan, line 153: I really dislike our `HttpClient` for reasons similar to why you have done this, but can I suggest an alternative:

Just return `A` and then assert on that in a map or whatever

I guess when the assertion is outside you lose a bit of the debug context but I think our client would generally be more usable if it were working this way. We have some horrible cases where we are running unsafe inside one of these assertion blocks because we want to assert the state of two requests

shinyhappydan (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -197,6 +197,60 @@ final class Files(
...
+
+  private def validateStorageTypeForCopy(source: StorageType, destination: Storage): IO[Unit] =
+    IO.raiseWhen(source == StorageType.S3Storage)(
+      WrappedStorageRejection(
+        InvalidStorageType(destination.id, source, Set(StorageType.DiskStorage, StorageType.RemoteDiskStorage))
+      )
+    ) >>
+      IO.raiseUnless(source == destination.tpe)(
+        WrappedStorageRejection(DifferentStorageType(destination.id, found = destination.tpe, expected = source))
+      )
```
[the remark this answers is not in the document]
olivergrabinski: The thing is that we shouldn't have the file go through Delta

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
olivergrabinski: Just to clarify: we _shouldn't_ have to fetch the file 😅

docs/src/main/paradox/docs/delta/api/files-api.md
```diff
@@ -89,6 +89,62 @@ Request
 Response
 :   @@snip [created-put.json](assets/files/created-put.json)
 
+## Create copy using POST or PUT
+
+Create a file copy based on a source file potentially in a different organization. No `MIME` details are necessary since this is not a file upload. Metadata such as the size and digest of the source file are preserved.
+
+The caller must have the following permissions:
+- `files/read` on the source project.
+- `storages/write` on the storage in the destination project.
```
imsdu, line 98: You need the permission defined on the destination storage

delta/sdk/src/main/resources/contexts/bulk-operation.json
```diff
@@ -0,0 +1,10 @@
+{
+  "@context": {
+    "_total": "https://bluebrain.github.io/nexus/vocabulary/total",
```
imsdu, line 3: Not sure if we need the total, we have in the search results because, there is pagination

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -195,6 +197,79 @@ final class Files(
     } yield res
   }.span("createLink")
 
+  def copyFiles(
```
imsdu, line 200: Files being already quite long and complex, it feels like we could introduce a `BatchFiles` that would take care of all the validation and would pass along the commands to Files ?
It may also make the testing easier ?

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/HttpClient.scala
dantb: Definitely agree that's the best way, but it would change a lot of places or create some duplication and this PR is already too big... So "for later" 👀

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -195,6 +197,79 @@ final class Files(
...
+      pc: ProjectContext,
+      dest: CopyFileDestination,
+      destStorageRef: ResourceRef.Revision,
+      destStorageTpe: StorageType,
+      destFilesAttributes: NonEmptyList[FileAttributes]
+  )(implicit c: Caller): IO[NonEmptyList[FileResource]] =
+    destFilesAttributes.traverse { destFileAttributes =>
+      for {
+        iri      <- generateId(pc)
+        command   = CreateFile(iri, dest.project, destStorageRef, destStorageTpe, destFileAttributes, c.subject, dest.tag)
```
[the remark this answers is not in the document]
dantb: Good shout, Cristina mentioned wanting that - this is already massive so I'm thinking do in a follow up?

docs/src/main/paradox/docs/delta/api/files-api.md
dantb: Thanks found that out during testing as well, gonna rewrite these docs once it's finished

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePluginModule.scala
```diff
@@ -147,6 +151,11 @@ class StoragePluginModule(priority: Int) extends ModuleDef {
 
   many[ResourceShift[_, _, _]].ref[Storage.Shift]
 
+  // TODO refactor Files to depend on this rather than constructing it
```
imsdu, line 154: Still a TODO here

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchCopy.scala
```diff
@@ -0,0 +1,125 @@
...
+      for {
+        _ <- IO.raiseWhen(sourcesBytes.exists(_ > maxSize))(SourceFileTooLarge(maxSize, destStorage.id))
+        _ <- validateRemainingStorageCapacity(destStorage, sourcesBytes)
+      } yield ()
+    }
+
+    private def validateRemainingStorageCapacity(destStorage: Storage, sourcesBytes: NonEmptyList[Long]) =
+      for {
+        spaceLeft <- storagesStatistics.getStorageAvailableSpace(destStorage)
+        totalSize  = sourcesBytes.toList.sum
```
imsdu, line 75: `sourceBytes.reduce` ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/BatchFilesRoutes.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  def routes: Route =
+    baseUriPrefix(baseUri.prefix) {
+      pathPrefix("bulk") {
+        pathPrefix("files") {
+          extractCaller { implicit caller =>
+            resolveProjectRef.apply { projectRef =>
+              (post & pathEndOrSingleSlash & parameter("storage".as[IdSegment].?) & indexingMode & tagParam) {
+                (storage, mode, tag) =>
+                  operationName(s"$prefixSegment/files/{org}/{project}") {
```
imsdu, line 60: Not needed, they still remain in some places but in the end it is disabled as in some cases:
*  this could not properly override the default behaviour (which is recording metrics for the given url)
* Some urls like the ones with a resource id for example were introducing a cardinality that Prometheus could not handle
* Prometheus would crash when there were too many resource reads
* Delta would also eventually crash as the collected metrics would continue to pile up until an OOM

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/StorageFileRejection.scala
```diff
@@ -65,6 +69,46 @@ object StorageFileRejection {
         extends FetchAttributeRejection(rejection.loggedDetails)
   }
 
+  /**
+    * Rejection returned when a storage cannot fetch a file's attributes
+    */
+  sealed abstract class CopyFileRejection(loggedDetails: String) extends StorageFileRejection(loggedDetails)
```
imsdu, line 75: Could we have it only on the `FileRejection` side ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchCopy.scala
```diff
@@ -0,0 +1,125 @@
...
+      storagesStatistics: StoragesStatistics,
+      diskCopy: DiskStorageCopyFiles,
+      remoteDiskCopy: RemoteDiskStorageCopyFiles
+  )(implicit uuidF: UUIDF): BatchCopy = new BatchCopy {
+
+    override def copyFiles(source: CopyFileSource, destStorage: Storage)(implicit
+        c: Caller
+    ): IO[NonEmptyList[FileAttributes]] =
+      destStorage match {
+        case disk: Storage.DiskStorage         => copyToDiskStorage(source, disk)
```
shinyhappydan, line 45: Maybe better as 2 classes dealing with each case alone

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchFiles.scala
```diff
@@ -0,0 +1,81 @@
...
+    override def copyFiles(source: CopyFileSource, dest: CopyFileDestination)(implicit
+        c: Caller
+    ): IO[NonEmptyList[FileResource]] = {
+      for {
+        pc                            <- fetchContext.onCreate(dest.project)
+        (destStorageRef, destStorage) <- fetchFileStorage.fetchAndValidateActiveStorage(dest.storage, dest.project, pc)
+        destFilesAttributes           <- batchCopy.copyFiles(source, destStorage).adaptError { case e: CopyFileRejection =>
+                                           CopyRejection(source.project, dest.project, destStorage.id, e)
+                                         }
+        fileResources                 <- evalCreateCommands(pc, dest, destStorageRef, destStorage.tpe, destFilesAttributes)
```
shinyhappydan, line 52: I realise that we are "evaling" create "commands", but maybe it's better to just say we are creating the files? Just as a more domain-oriented reading of what is happening

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/BatchFilesRoutes.scala
```diff
@@ -0,0 +1,108 @@
...
+                    }
+                  }
+              }
+            }
+          }
+        }
+      }
+    }
+
+  private def copyFile(mode: IndexingMode, source: CopyFileSource, dest: CopyFileDestination)(implicit
```
shinyhappydan, line 74: Isn't this copy fileS?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchCopy.scala
shinyhappydan: I like that `.sum` is explicit

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/DiskStorageSaveFile.scala
```diff
@@ -70,13 +70,23 @@ object DiskStorageSaveFile {
       disk: DiskStorageValue,
       uuid: UUID,
       filename: String
+  ): IO[(Path, Path)] =
+    for {
+      (resolved, relative) <- computeLocation(project, disk, uuid, filename)
+      dir                   = resolved.getParent
+      _                    <- IO.blocking(Files.createDirectories(dir)).adaptError(couldNotCreateDirectory(dir, _))
```
shinyhappydan, line 77: Maybe it's worth Files.createDirectories be blocking rather than needing the caller to block

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/package.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import java.net.URI
+import java.nio.file
+import java.nio.file.Paths
+
+package object disk {
+
+  def absoluteDiskPathFromAttributes(attr: FileAttributes): IO[file.Path] = absoluteDiskPath(attr.location.path)
+
+  def absoluteDiskPath(relative: Path): IO[file.Path] = IO(Paths.get(URI.create(s"file://$relative")))
```
shinyhappydan, line 15: My instinct here is that stuff touching the file system would be better off in a class so it's mockable, perhaps that doesn't apply here

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchCopySuite.scala
```diff
@@ -0,0 +1,293 @@
...
+    val (sourceFileRes, sourceStorage) = genFileResourceAndStorage(sourceFileId, sourceProj.context, diskVal)
+    val (user, aclCheck)               = userAuthorizedOnProjectStorage(sourceStorage.value)
+
+    val batchCopy                = mkBatchCopy(
+      fetchFile = stubbedFetchFile(sourceFileRes, events),
+      fetchStorage = stubbedFetchStorage(sourceStorage, events),
+      aclCheck = aclCheck,
+      stats = stubbedStorageStats(storageStatEntry, events),
+      diskCopy = stubbedDiskCopy(stubbedFileAttr, events)
+    )
```
shinyhappydan, line 55: This is a big setup

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchCopySuite.scala
```diff
@@ -0,0 +1,293 @@
...
+    val destStorage: DiskStorage = genDiskStorage()
+
+    batchCopy.copyFiles(source, destStorage)(caller(user)).map { obtained =>
+      val obtainedEvents = events.toList
```
shinyhappydan, line 59: I wonder if an abstraction around the events list might be a bit clearer

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchFilesSuite.scala
```diff
@@ -0,0 +1,118 @@
...
+
+    // resources returned are based on file command evaluation
+    assertEquals(obtained, expectedCmds.map(genFileResourceFromCmd))
+
+    val expectedActiveStorageFetched = ActiveStorageFetched(destination.storage, destProj.ref, destProj.context, c)
+    val expectedBatchCopyCalled      = BatchCopyCalled(source, destStorage.storage, c)
+    val expectedCommandsEvaluated    = expectedCmds.toList.map(FileCommandEvaluated)
+    val expectedEvents               = List(expectedActiveStorageFetched, expectedBatchCopyCalled) ++ expectedCommandsEvaluated
+    assertEquals(events.toList, expectedEvents)
+  }
```
shinyhappydan, line 56: Not sure what to suggest but this feels like a really big test

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/BulkOperationResults.scala
```diff
@@ -0,0 +1,24 @@
+package ch.epfl.bluebrain.nexus.delta.sdk.jsonld
+
+import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.{contexts, nxv}
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.ContextValue
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.encoder.JsonLdEncoder
+import io.circe.syntax._
+import io.circe.{Encoder, Json, JsonObject}
+
+final case class BulkOperationResults[A](results: Seq[A])
```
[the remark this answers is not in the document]
shinyhappydan: I would go the opposite direction and call it `CopyFilesResult` 😆

shinyhappydan (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchCopy.scala
dantb: 🤷

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/StorageFileRejection.scala
dantb: Ahh as in ditch the existing pattern since it just adds more error types than are necessary? I guess this came from monix types in the past

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/BulkOperationResults.scala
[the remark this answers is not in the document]
dantb: Is it not? The only purpose of this type was to encode the `bulk-operation` context in a contained way

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/TransactionalFileCopier.scala
```diff
@@ -1,27 +1,41 @@
...
+  private val logger = Logger[TransactionalFileCopier]
+
+  def mk(): TransactionalFileCopier = files => copyAll(files)
+
+  private def copyAll(files: NonEmptyList[CopyBetween]): IO[Unit] =
     Ref.of[IO, Option[CopyOperationFailed]](None).flatMap { errorRef =>
       files
         .parTraverse { case c @ CopyBetween(source, dest) =>
-          copySingle(source, dest).onError(_ => errorRef.set(Some(CopyOperationFailed(c))))
+          copySingle(source, dest).onError(e => errorRef.set(Some(CopyOperationFailed(c, e))))
```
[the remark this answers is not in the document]
dantb: Don't think I understand the question 🤔

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePluginModule.scala
```diff
@@ -185,6 +194,41 @@ class StoragePluginModule(priority: Int) extends ModuleDef {
...
+        fetchContext: FetchContext[ContextRejection],
+        files: Files,
+        filesLog: FilesLog,
+        batchCopy: BatchCopy,
+        uuidF: UUIDF
+    ) =>
+      BatchFiles.mk(
+        files,
+        fetchContext.mapRejection(FileRejection.ProjectContextRejection),
+        FilesLog.eval(filesLog),
```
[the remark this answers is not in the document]
dantb: Hmm yeah I did, was just somewhere to prevent duplication but it's not too important. I.e. "any saving of file events / states goes through this" 🤷

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchFiles.scala
dantb: The actual files already exist so maybe create file resources 👍

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchCopy.scala
dantb: Thought about this, it felt like just more verbosity since most of the logic is shared. Also do not want to change the tests again 👀  (they're maybe probably a little overkill tbh)

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/DiskStorageSaveFile.scala
dantb: That's a Java method (ideally it should use fs2 but I didn't want to change this everywhere here)

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/package.scala
dantb: The fs2 `Files[IO]` probably is this abstraction (or maybe a smaller wrapper round it). This was just removing duplication again

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/TransactionalFileCopier.scala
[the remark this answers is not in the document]
shinyhappydan: Why are we setting a ref when we could just return the error?

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchCopySuite.scala
dantb: There is actually a really nice pure FP abstraction for this kinda of testing (no mutable state), but it relies on tagless final / monad transformers so not a good fit here. I'll think about how to make it clearer tho 👍

shinyhappydan: I don't think the mutable state is the problem, it just needs to read better. I was thinking just a tiny class / enclosure

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchFiles.scala
```diff
@@ -0,0 +1,81 @@
...
+        dest: CopyFileDestination,
+        destStorageRef: ResourceRef.Revision,
+        destStorageTpe: StorageType,
+        destFilesAttributes: NonEmptyList[FileAttributes]
+    )(implicit c: Caller): IO[NonEmptyList[FileResource]] =
+      destFilesAttributes.traverse { destFileAttributes =>
+        for {
+          iri      <- generateId(pc)
+          command   =
+            CreateFile(iri, dest.project, destStorageRef, destStorageTpe, destFileAttributes, c.subject, dest.tag)
```
imsdu, line 67: Not sure if we want to preserve the original tag.
It would be nice in some way to know that this file has been copied from another

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/BulkOperationResults.scala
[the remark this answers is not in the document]
imsdu: If it is a `CopyFileResult`, it should go in the storage plugin

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/DiskStorageSpec.scala
```diff
@@ -8,7 +8,7 @@ import ch.epfl.bluebrain.nexus.tests.iam.types.Permission
 import io.circe.Json
 import org.scalatest.Assertion
 
-class DiskStorageSpec extends StorageSpec {
+class DiskStorageSpec extends StorageSpec with CopyFilesSpec {
```
imsdu, line 11: It could be other test classes dedicated for the batch operation ?
Those are already complicated, it may be better to split ?

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/TransactionalFileCopier.scala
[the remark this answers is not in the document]
dantb: I'll refactor this afterwards

PR #4485, opened 2023-11-09
Migrate Http Client to Cats Effect

### LINKED ISSUE #4433: Migrate HttpClient to CatsEffect

shinyhappydan:
Fixes #4433, #4434

### DIFF
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
@@ -35,9 +35,9 @@ import ch.megard.akka.http.cors.scaladsl.settings.CorsSettings
 import com.typesafe.config.Config
 import izumi.distage.model.definition.{Id, ModuleDef}
 import monix.bio.UIO
-import monix.execution.Scheduler
 import org.slf4j.{Logger, LoggerFactory}
 
+import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.DurationInt
 
 /**
@@ -62,7 +62,7 @@ class DeltaModule(appCfg: AppConfig, config: Config)(implicit classLoader: Class
   make[StrictEntity].from { appCfg.http.strictEntityTimeout }
   make[ServiceAccount].from { appCfg.serviceAccount.value }
 
-  implicit val scheduler: Scheduler = Scheduler.global
+  implicit val executionContext: ExecutionContext = ExecutionContext.global
 
   make[Transactors].fromResource { (cs: ContextShift[IO]) =>
     Transactors.init(appCfg.database)(classLoader, cs)
@@ -114,7 +114,7 @@ class DeltaModule(appCfg: AppConfig, config: Config)(implicit classLoader: Class
   make[Clock[IO]].from(Clock.create[IO])
   make[EvaluationExecution].from(EvaluationExecution(_, _))
   make[UUIDF].from(UUIDF.random)
-  make[Scheduler].from(scheduler)
+  make[ExecutionContext].from(executionContext)
   make[JsonKeyOrdering].from(
     JsonKeyOrdering.default(topKeys =
       List("@context", "@id", "@type", "reason", "details", "sourceId", "projectionId", "_total", "_results")
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
@@ -114,7 +114,6 @@ class DeltaModule(appCfg: AppConfig, config: Config)(implicit classLoader: Class
   make[Clock[IO]].from(Clock.create[IO])
   make[EvaluationExecution].from(EvaluationExecution(_, _))
   make[UUIDF].from(UUIDF.random)
-  make[ExecutionContext].from(executionContext)
   make[JsonKeyOrdering].from(
     JsonKeyOrdering.default(topKeys =
       List("@context", "@id", "@type", "reason", "details", "sourceId", "projectionId", "_total", "_results")
@@ -136,6 +135,7 @@ class DeltaModule(appCfg: AppConfig, config: Config)(implicit classLoader: Class
     }
     Resource.make(make)(release)
   }
+  make[ExecutionContext].from((as: ActorSystem[Nothing]) => as.executionContext)
 
   make[Materializer].from((as: ActorSystem[Nothing]) => SystemMaterializer(as).materializer)
   make[Logger].from { LoggerFactory.getLogger("delta") }
--- a/delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/RetryStrategy.scala
+++ b/delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/RetryStrategy.scala
@@ -1,10 +1,7 @@
 package ch.epfl.bluebrain.nexus.delta.kernel
 
-import cats.effect.IO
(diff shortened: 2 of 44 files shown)


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
```diff
@@ -62,7 +62,7 @@ class DeltaModule(appCfg: AppConfig, config: Config)(implicit classLoader: Class
   make[StrictEntity].from { appCfg.http.strictEntityTimeout }
   make[ServiceAccount].from { appCfg.serviceAccount.value }
 
-  implicit val scheduler: Scheduler = Scheduler.global
+  implicit val executionContext: ExecutionContext = ExecutionContext.global
```
dantb, line 65: Think this would ideally be a small fixed threadpool , assuming everywhere it's used is for non-blocking IO (like the HTTP client). According to the [CE docs](https://typelevel.org/cats-effect/docs/2.x/concurrency/basics#choosing-thread-pool) the key thing is to have a small number of threads dedicated to handling the async notification (the "event dispatcher" pool) and immediately shift back to the optimised compute pool (the one from `IOApp`). I always go back to this gist - https://gist.github.com/djspiewak/46b543800958cf61af6efa8e072bfd5c

The problem with the global EC is we don't know what else is scheduled on it (e.g. by libraries) so it could add latency when these notifications aren't processed immediately.

Having said that, it's [apparently](https://typelevel.org/cats-effect/docs/2.x/concurrency/basics#asynchronous--semantic-blocking) fine to use the default CE compute pool for non-blocking HTTP clients like this. Note the big caveat that the HTTP client must be non-blocking, which Akka HTTP is I think. I'm going to ask in Discord, because I've always heard both recommendations and it feels like a bit of a contradiction 🤷 

Sorry, a bit long, just feels important because `HttpClient` is used everywhere under the hood. As the docs say, `when it comes to concurrency noone has any idea what they're doing` 😂

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/RetryStrategy.scala
```diff
@@ -29,33 +27,40 @@ import scala.util.control.NonFatal
 final case class RetryStrategy[E](
     config: RetryStrategyConfig,
     retryWhen: E => Boolean,
-    onError: (E, RetryDetails) => BIO[E, Unit]
+    onError: (E, RetryDetails) => IO[Unit]
 ) {
-  val policy: RetryPolicy[BIO[E, *]] = config.toPolicy[E]
+  val policy: RetryPolicy[IO[*]] = config.toPolicy[E]
```
dantb, line 32: Maybe another one for later - but do we do anything bespoke enough to need this wrapper? If not we could use cats-retry directly

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FilesSpec.scala
```diff
@@ -38,34 +38,35 @@ import ch.epfl.bluebrain.nexus.delta.sourcing.model.Identity.{Anonymous, Authent
...
 import org.scalatest.concurrent.Eventually
+import org.scalatest.{Assertion, DoNotDiscover}
 
 import java.net.URLDecoder
 
 @DoNotDiscover
 class FilesSpec(docker: RemoteStorageDocker)
     extends TestKit(ActorSystem("FilesSpec"))
     with CatsEffectSpec
+    with BioRunContext
```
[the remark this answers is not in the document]
shinyhappydan: Yes, because of the execution context. I realise that's a bit off but I think it's one to fix later

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClient.scala
```diff
@@ -33,83 +30,95 @@ trait HttpClient {
...
 
   private[http] def apply(
       client: HttpSingleRequest
-  )(implicit httpConfig: HttpClientConfig, as: ActorSystem, scheduler: Scheduler): HttpClient =
+  )(implicit
+      httpConfig: HttpClientConfig,
+      as: ActorSystem,
+      ec: ExecutionContext,
+      timer: Timer[IO],
+      cs: ContextShift[IO]
```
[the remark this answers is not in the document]
shinyhappydan: Retry

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
shinyhappydan: My understanding was the `Scheduler.global` (the EC we were previously using) is the same as `ExecutionContext.global`, so I think my change doesn't make anything worse

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/RetryStrategy.scala
shinyhappydan: Probably but trying to no-brain this over the line

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
imsdu: It would be better to user the execution context provided by akka that you can access via `actorSystem.dispatcher`
That is the option we took in other cases

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClient.scala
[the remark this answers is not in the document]
imsdu: Could be done in a later PR but I think we don't use the retry strategy anymore in the HttpClient (we always pass the `alwaysGiveUp`)
The reason behind that is it was creating more trouble than it solved because this one is very low-level and could be included in other retries, creating in the end timeouts for clients, etc...

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/remote/client/RemoteDiskStorageClient.scala
```diff
@@ -68,7 +67,7 @@ final class RemoteDiskStorageClient(client: HttpClient, getAuthToken: AuthTokenP
       val req      = Head(endpoint).withCredentials(authToken)
       client(req) {
         case resp if resp.status.isSuccess() => IO.delay(resp.discardEntityBytes()).void.toUIO
```
imsdu, line 69: Feels weird we still have this `toUIO`

shinyhappydan: Yeah it's probably fixed in another commit I've already made 😓

dantb (approved):
LGTM - FWIW the answer seems to be use the Akka dispatcher for the EC, which is on `ActorSystem`. But it's probably better to deploy separately. I have no idea how it'll interact with `IO.fromFuture` in terms of threading 🤷

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
dantb: Beat me to it! [Docs](https://doc.akka.io/docs/akka-http/current/handling-blocking-operations-in-akka-http-routes.html). Just not sure what happens when going to IO

PR #4493, opened 2023-11-10
Cats Effect 3 Migration

shinyhappydan:


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala
```diff
@@ -55,7 +54,7 @@ final class ProjectsRoutes(
     cr: RemoteContextResolution,
     ordering: JsonKeyOrdering,
     fusionConfig: FusionConfig,
-    contextShift: ContextShift[IO]
+    runtime: IORuntime
```
imsdu, line 57: You only need to import:
```
import cats.effect.unsafe.implicits.global
```
Where we have a call to an unsafe method

build.sbt
```diff
@@ -25,16 +25,16 @@ val awsSdkVersion           = "2.17.184"
...
-val catsEffectVersion       = "2.5.5"
-val catsRetryVersion        = "2.1.1"
+val catsEffectVersion       = "3.5.2"
+val catsRetryVersion        = "3.1.0"
 val catsVersion             = "2.10.0"
 val circeVersion            = "0.14.6"
 val circeOpticsVersion      = "0.15.0"
 val circeExtrasVersions     = "0.14.3"
 val classgraphVersion       = "4.8.164"
 val distageVersion          = "1.0.10"
```
imsdu, line 35: Update distage version to get support for CE3

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/ClasspathResourceUtils.scala
```diff
@@ -14,7 +14,9 @@ import scala.jdk.CollectionConverters._
 
 trait ClasspathResourceUtils {
 
-  final def absolutePath(resourcePath: String)(implicit classLoader: ClassLoader): IO[String] =
+  private val classLoader: ClassLoader = getClass.getClassLoader
+
+  final def absolutePath(resourcePath: String): IO[String] =
```
imsdu, line 19: Removing the classloader is very likely to cause some problems when loading some files in the plugins.
It is a good opportunity to see if it is really mandatory though :)

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala
shinyhappydan: The text you quoted the other day:

> Instead of calling unsafe methods directly, consider using cats.effect.IOApp, which runs your IO. If integrating with non-functional code or experimenting in a REPL / Worksheet, add the following import:
> `import cats.effect.unsafe.implicits.global`

Doesn't that imply that we should NOT do that in production?

build.sbt
shinyhappydan: This does help with the class not found problem, thanks

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/stream/StreamConverter.scala
```diff
@@ -45,12 +44,12 @@ object StreamConverter {
       .evalMap(publish)
       .unNoneTerminate
       .onFinalizeCase {
-        case ExitCase.Completed | ExitCase.Canceled => complete
-        case ExitCase.Error(e)                      => fail(e)
+        case ExitCase.Succeeded | ExitCase.Canceled => complete
```
imsdu, line 47: The fs2 finalize methods accept a `Resource.ExitCase`
A bit weird they renamed one `Outcome` but not this one

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/HttpClient.scala
```diff
@@ -55,11 +56,23 @@ class HttpClient private (baseUrl: Uri, httpExt: HttpExt)(implicit
   )(implicit um: FromEntityUnmarshaller[A]): IO[Assertion] =
     requestAssert(POST, url, Some(body), identity, extraHeaders)(assertResponse)
 
+  def post2[A](url: String, body: IO[Json], identity: Identity, extraHeaders: Seq[HttpHeader] = jsonHeaders)(
```
imsdu, line 59: Temporary ?

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutes.scala
imsdu: The implicit:
```
object implicits {
  implicit def global: IORuntime = IORuntime.global
}
```
Which is the same value as the one you pass down

I think what they meant is that you should not use it unless you get away from the holy path of functional code.
And akka falls into that category because you need to call `unsafeFuture`

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/HttpClient.scala
shinyhappydan: Will be renamed :) but it's sort of needed

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/ClasspathResourceUtils.scala
shinyhappydan: This was an issue, the fix is pushed

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/DeltaModule.scala
```diff
@@ -101,28 +101,26 @@ class DeltaModule(appCfg: AppConfig, config: Config)(implicit classLoader: Class
       .merge(otherCtxResolutions.toSeq: _*)
   }
 
-  make[JsonLdApi].from { contextShift: ContextShift[IO] =>
-    new JsonLdJavaApi(appCfg.jsonLdApi)(contextShift)
+  make[JsonLdApi].from { () =>
```
imsdu, line 104: It is possible to use `fromValue` here

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/stream/StreamConverter.scala
```diff
@@ -45,12 +46,12 @@ object StreamConverter {
...
+        case ExitCase.Errored(e)                    => fail(e)
       }
   }
 
-  def apply[A](stream: Stream[IO, A])(implicit contextShift: ContextShift[IO]): Graph[SourceShape[A], NotUsed] = {
+  def apply[A](stream: Stream[IO, A]): Graph[SourceShape[A], NotUsed] = {
```
imsdu, line 54: This one could be removed no ?
We don't have a use for it anymore do we ?

imsdu (approved):

PR #4499, opened 2023-11-15
Publish full storage service jar to sonatype

olivergrabinski:
Should publish the full storage service jar to sonatype

### DIFF
--- a/.github/workflows/ci-release-sonatype.yml
+++ b/.github/workflows/ci-release-sonatype.yml
@@ -28,4 +28,5 @@ jobs:
           java-version: '17'
           check-latest: true
       - name: Publish To Sonatype
-        run: sbt -Dsbt.color=always -Dsbt.supershell=false ci-release
+        run: sbt -Dsbt.color=always -Dsbt.supershell=false ci-release \
+          storage/Universal/publish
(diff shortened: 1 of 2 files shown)


shinyhappydan (approved):

.github/workflows/ci-release-sonatype.yml
```diff
@@ -28,4 +28,5 @@ jobs:
           java-version: '17'
           check-latest: true
       - name: Publish To Sonatype
-        run: sbt -Dsbt.color=always -Dsbt.supershell=false ci-release
\ No newline at end of file
+        run: sbt -Dsbt.color=always -Dsbt.supershell=false ci-release \
+          storage/Universal/publish
```
imsdu, line 32: I hope it is gonna be reliable enough or it will have to go in its own workflow

imsdu (approved):

.github/workflows/ci-release-sonatype.yml
olivergrabinski: `storage/Universal/publish` was already present in the ci-snapshot workflow for a week or so and I don't see any failure because of it. So it should be ok. But if it's problematic we can change it very easily

PR #4503, opened 2023-11-16
Reject ambiguous organization DELETE requests

### LINKED ISSUE #4502: Using query parameters `prune=true&rev={rev}` should not deprecate an org
**Motivation**
`pure=true` is supposed to indicate that the org is to be deleted. However, if a rev is present it will hit the deprecation endpoint instead even if `pure=true` is there.

dantb:
Fixes #4502


olivergrabinski (approved):
As a side note the example for org deletion in the doc is wrong since it has just the `rev` query parameter but not `prune=true`

PR #4506, opened 2023-11-17
Do not create a revision when no change is detected

### LINKED ISSUE #4404: Do not create a revision of a resource if the payload or the remote contexts have not changed
**Motivation**
Users sometimes push identical resources over and over and this results in the creation of revisions that bring noise in Nexus as well extra used resources

**Acceptance criteria**
1. When the payload has changed, an update event is emitted and a new revision is created
2. When the payload has not changed but one of the remote contexts has been updated, a refresh event is emitted and a new revision is created
3. If no change is detected, no update/refresh event is done
4. If a tag is provided, it is always performed either with the update/refresh event or alone if no change has been detected

imsdu:
Fixes #4404


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -423,40 +423,59 @@ object Resources {
...
+          for {
+            (schemaRev, schemaProject) <- validate(u.id, expanded, schemaRef, s.project, u.caller)
+            time <- IOInstant.now
+          } yield ResourceUpdated(u.id, u.project, schemaRev, schemaProject, types, u.source, compacted, expanded, remoteContextRefs, s.rev + 1, time, u.subject, u.tag)
+        } else {
+          // If no changes have been detected on the resource itself, we fallback on other commands or raise an error
+          (u.schemaOpt, u.tag) match {
+            case (Some(schema), tagOpt) => updateResourceSchema(UpdateResourceSchema(u.id, u.project, schema, u.rev, u.caller, tagOpt))
+            case (None, Some(tagValue)) => tag(TagResource(u.id, u.project, u.schemaOpt, s.rev, tagValue, u.rev, u.subject))
+            case (None, None) => IO.raiseError(NoChangeDetected(s))
```
imsdu, line 439: It would be better to change the signature of the evaluate method instead of raising an error we recover afterwards.
I made some tests about it but those are some impacting changes and it is better to postpone them waiting for the other big changes to be merged

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -422,40 +422,57 @@ object Resources {
...
       // format: on
     }
 
     def tag(c: TagResource) = {
       for {
-        s    <- stateWhereRevisionExists(c, c.targetRev)
-        time <- clock.realTimeInstant
+        s                        <- stateWhereRevisionExists(c, c.targetRev)
+        tagAlreadyExistAtGivenRev = s.tags.value.exists { case (t, r) => t == c.tag && r == c.targetRev }
+        _                        <- IO.raiseWhen(tagAlreadyExistAtGivenRev)(NoChangeDetected(s))
```
olivergrabinski, line 474: I don't understand why this is a NoChangeDetected error here

olivergrabinski: I misread which operation this was on first pass

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -422,40 +422,57 @@ object Resources {
...
+      import r.jsonld._
       // format: off
       for {
-        s                          <- stateWhereResourceIsEditable(c)
-        _                          <- raiseWhenDifferentSchema(c, s)
-        (schemaRev, schemaProject) <- validate(c.id, expanded, c.schemaOpt.getOrElse(s.schema), s.project, c.caller)
+        s                          <- stateWhereResourceIsEditable(r)
+        _                          <- raiseWhenDifferentSchema(r, s)
+        (schemaRev, schemaProject) <- validate(r.id, expanded, r.schemaOpt.getOrElse(s.schema), s.project, r.caller)
+        noChangeDetected           = s.remoteContexts == remoteContextRefs && s.expanded == expanded
```
olivergrabinski, line 463: As above, if the user wants to refresh but nothing changed, then validation doesn't matter?

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/Resources.scala
```diff
@@ -422,40 +422,57 @@ object Resources {
...
 
     def updateResourceSchema(u: UpdateResourceSchema) = {
       for {
-        s                          <- stateWhereResourceIsEditable(u)
-        (schemaRev, schemaProject) <- validate(u.id, u.expanded, u.schemaRef, s.project, u.caller)
-        types                       = u.expanded.getTypes.getOrElse(Set.empty)
+        s <- stateWhereResourceIsEditable(u)
+
+        (schemaRev, schemaProject) <- validate(u.id, s.expanded, u.schemaRef, s.project, u.caller)
+        _                          <- IO.raiseWhen(s.schema.iri == u.schemaRef.iri)(NoChangeDetected(s))
```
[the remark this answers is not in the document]
imsdu: For the update schema case, I think you are right even if it seems there is a small bug as in this implementation as you can't update a schema to use another tag of the same schema for example.
So I will remove it for now because we need to also save in the resource the schema (latest/tag/revision) requested by the user besides the revision of the schema the resource has been validated with.

For the refresh case, we want to raise an error if the resource is validated by the latest version of a schema as this one may have changed since the last update.

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImplSpec.scala
```diff
@@ -410,24 +421,34 @@ class ResourcesImplSpec
...
+          _                      <- resources.create(contextId, projectRef, schema, contextSource, None)
+          createdResult          <- resources.create(idRefresh, projectRef, schema, refreshSource, None)
+          refreshedNoChange      <- resources.refresh(idRefresh, projectRef, None)
+          _                       = refreshedNoChange shouldEqual createdResult
+          updatedContext         <- resources.update(contextId, projectRef, None, 1, contextUpdatedSource, None)
+          _                       = updatedContext.rev shouldEqual 2
+          refreshedContextChange <- resources.refresh(idRefresh, projectRef, None)
+          _                       = refreshedContextChange shouldEqual createdResult.copy(rev = 2)
+        } yield ()
+        io.accepted
```
olivergrabinski, line 451: This seems to be perhaps too many tests in one IMO

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/model/ResourceEvent.scala
```diff
@@ -200,6 +263,32 @@ object ResourceEvent {
       subject: Subject
   ) extends ResourceEvent
 
+  object ResourceRefreshed {
+    def apply(
```
olivergrabinski, line 267: This apply method could be used in `Resources`

olivergrabinski (dismissed):

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/shacl/ValidationReport.scala
```diff
@@ -38,8 +38,9 @@ object ValidationReport {
 
   private val shaclCtx: ContextValue = ContextValue(contexts.shacl)
 
-  final def apply(report: Resource)(implicit api: JsonLdApi, rcr: RemoteContextResolution): IO[ValidationReport] = {
-    val tmpGraph = Graph.unsafe(DatasetFactory.create(report.getModel).asDatasetGraph())
+  final def apply(report: Resource)(implicit rcr: RemoteContextResolution): IO[ValidationReport] = {
+    implicit val api: JsonLdApi = JsonLdJavaApi.lenient
```
imsdu, line 42: We only need to use the configured `JsonLdApi` for write operations of resources.
For read operations or like in this case, it is generated by an underlying library, there is no reason to be strict and enforce extra checks

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdAssembly.scala
```diff
@@ -0,0 +1,78 @@
...
+  * @param compacted
+  *   its compacted json-ld representation
+  * @param expanded
+  *   its expanded json-ld representation
+  * @param graph
+  *   its graph representation
+  * @param remoteContexts
+  *   it
+  */
+final case class JsonLdAssembly(
```
imsdu, line 32: I introduces this one so that:
* We gather all computed representations in the same place
* To be able to test for their equality

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ResourcesImplSpec.scala
imsdu: The idea is to check that the refresh operation on the resource only creates a revision if one of the remote contexts has been updated.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdAssembly.scala
```diff
@@ -0,0 +1,78 @@
...
+    id: Iri,
```
olivergrabinski, line 33: I find it a bit strange at first sight to move the id here; i.e. if we have a command, we now get the id from the `JsonLdAssembly`.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdAssembly.scala
```diff
@@ -0,0 +1,78 @@
...
+    *   - If the remote contexts and the local context are the same, then the compacted form will be the same
+    *   - If the graph forms are isomorphic then, the expanded form will be the same
+    */
+  implicit val jsonLdAssemblyEq: Eq[JsonLdAssembly] = Eq.instance { (jsonld1, jsonld2) =>
+    jsonld1.id == jsonld2.id &&
+    jsonld1.remoteContexts == jsonld2.remoteContexts &&
+    jsonld1.compacted.ctx == jsonld2.compacted.ctx &&
+    jsonld1.graph.isIsomorphic(jsonld2.graph) &&
+    jsonld1.source == jsonld2.source
+  }
```
olivergrabinski, line 61: Why do we need to check the `id` here? Doesn't the graph isomorphism guarantee that already?

The following passes:

```scala
    "not be isomorphic when the id is different" in {
      val graph2 = graph.replaceRootNode(iri"http://other.id")
      graph.isIsomorphic(graph2) shouldEqual false
    }
```

imsdu: It is a fail fast thing, I tried to put the least expensive checks first

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdAssembly.scala
imsdu: At creation, we already did that if the id was in the payload or if it is generated.
For updates, there is also a check if the id provided in the segment is equal to the one in the payload.
So the idea was instead of extracting it to copy it aside is to just use it directly to make the signature a bit simpler

olivergrabinski (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdAssembly.scala
olivergrabinski: Ok make sense. I wanted to check if there wasn't something I was misunderstanding about `isIsomorphic`

PR #4511, opened 2023-11-20
Add bulk copy operation to nexus storage service

dantb:
This PR was getting too invasive so I've left some stuff for later:
1. Using `Files[IO]` from fs2 and the corresponding `Path` type everywhere.
2. Stop mixing error channels - just use `IO`. Then all the `EitherT` can be removed. I wanted to first just refactor it to be more modular / pure without breaking anything.
3. Breaking apart the tests for `DiskStorage` and `ValidateFiles` .
4. Capturing the effects properly for things like checking path / bucket existence (it's all intertwined with the directives).

Summary of changes
- Extract out validation logic into `ValidateFiles`.
- Add `CopyFiles` to transactionally copy a bunch of files, rolling back to avoid halfway states.
- Try to encapsulate effects properly. Includes fs2 `Files[IO]` and manual blocking in a few places.
- Tries to make throwing / rejecting a bit more explicit with a view to refactor later.
- Add POST endpoint for copying multiple files, accepting a list of source/dest paths and returning a list of the created absolute paths. 
  I know you mentioned issues with bulk copying @imsdu  but I figured this gives us options when calling from delta. If it ends up being a problem we could always cap the number of copied files in delta's API (even to just 1). This also avoids the big feedback loop when deploying the storage service.
- Introduce munit to storage module.


storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/StorageError.scala
```diff
@@ -95,6 +96,11 @@ object StorageError {
     */
   final case class OperationTimedOut(override val msg: String) extends StorageError(msg)
 
+  final case class CopyOperationFailed(name: String, source: Fs2Path, dest: Fs2Path)
```
shinyhappydan, line 99: It would be better to only have domain types in the errors, if we use something other than FS2 then we shouldn't have to change the error here

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/Storages.scala
```diff
@@ -72,14 +87,13 @@ trait Storages[Source] {
...
-    *   Left(rejection) or Right(fileAttributes). The file attributes contain the metadata (bytes and location) wrapped
-    *   in an F effect type
+    *   Left(rejection) or Right(fileAttributes). The file attributes contain the metadata (bytes and location)
     */
   def moveFile(
       name: String,
       sourcePath: Uri.Path,
       destPath: Uri.Path
-  )(implicit bucketEv: BucketExists): IO[Either[Rejection, FileAttributes]]
+  )(implicit bucketEv: BucketExists): IO[RejOrAttributes]
```
shinyhappydan, line 96: I find the explicit `Either` approx 1000% more useful than the type alias here

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/Storages.scala
```diff
@@ -140,9 +141,13 @@ object Storages {
...
-    final case object PathExists       extends PathExistence
-    final case object PathDoesNotExist extends PathExistence
-    type PathExists       = PathExists.type
+    final case object PathExists       extends PathExistence {
+      val exists = true
+    }
+    final case object PathDoesNotExist extends PathExistence {
+      val exists = false
+    }
+    type PathExists = PathExists.type
```
shinyhappydan, line 150: I didn't realise you can do this, I was always using traits to achieve the same thing. Nice

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/Storages.scala
```diff
@@ -153,157 +158,126 @@ object Storages {
...
+        computeSizeAndMoveFile(absSourcePath, absDestPath, isDir)
+
+    private def fixPermissions(path: Path): IO[Unit] =
+      if (config.fixerEnabled) {
+        val absPath = path.toAbsolutePath.normalize.toString
+        val logger  = StringProcessLogger(config.fixerCommand, absPath)
+        val process = Process(config.fixerCommand :+ absPath)
+
+        for {
+          exitCode <- IO.delay(process ! logger)
```
imsdu, line 239: This one should be blocking

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/Storages.scala
```diff
@@ -54,14 +54,29 @@ trait Storages[Source] {
...
+    * Copy files between locations inside the nexus folder. Attributes are neither recomputed nor fetched; it's assumed
+    * clients already have this information from the source files.
+    *
+    * @param name
+    *   the storage bucket name
+    * @param files
+    *   a list of source/destination files. The source files should exist under the nexus folder, and the destination
+    *   files will be created there. Both must be files, not directories.
+    */
+  def copyFile(
```
shinyhappydan, line 75: I think `copyFiles`

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFileOutput.scala
```diff
@@ -0,0 +1,19 @@
...
+import io.circe.Encoder
+import io.circe.generic.semiauto.deriveEncoder
+
+import java.nio.file.Path
+
+final case class CopyFileOutput(
+    sourcePath: Uri.Path,
+    destinationPath: Uri.Path,
+    absoluteSourceLocation: Path,
+    absoluteDestinationLocation: Path
```
shinyhappydan, line 14: I would just want to double check this response information would be used by a client?

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFiles.scala
```diff
@@ -0,0 +1,49 @@
+package ch.epfl.bluebrain.nexus.storage.files
+
+import cats.data.NonEmptyList
+import cats.effect.{IO, Ref}
+import cats.implicits.catsSyntaxParallelTraverse1
+import ch.epfl.bluebrain.nexus.storage.StorageError.CopyOperationFailed
+import fs2.io.file.{CopyFlag, CopyFlags, Files, Path}
+
+trait CopyFiles {
+  def copyValidated(name: String, files: NonEmptyList[ValidatedCopyFile]): IO[Unit]
```
shinyhappydan, line 10: It's probably cleared is this is bucket rather than name, I think there are other places where this is true also

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFileSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+import cats.syntax.all._
+import ch.epfl.bluebrain.nexus.storage.StorageError.CopyOperationFailed
+import ch.epfl.bluebrain.nexus.storage.files.CopyFiles.{parent, CopyBetween}
+import fs2.io.file.{Files, Path}
+import munit.CatsEffectSuite
+import munit.catseffect.IOFixture
+
+import java.util.UUID
+
+class CopyFileSpec extends CatsEffectSuite {
```
imsdu, line 14: CopyFileSuite

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFiles.scala
```diff
@@ -0,0 +1,49 @@
...
+}
+
+object CopyFiles {
+  def mk(): CopyFiles = (name, files) =>
+    copyAll(name, files.map(v => CopyBetween(Path.fromNioPath(v.absSourcePath), Path.fromNioPath(v.absDestPath))))
+
+  final private[files] case class CopyBetween(source: Path, dest: Path)
+
+  def copyAll(name: String, files: NonEmptyList[CopyBetween]): IO[Unit] =
```
shinyhappydan, line 19: This shouldn't be public, the test can just use an instance

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFileSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+    for {
+      _     <- givenAFileAtPath(source)
+      _     <- givenAFileAtPath(dest2)
+      error <- CopyFiles.copyAll(bucket, files).intercept[CopyOperationFailed]
+      _     <- filesShouldNotExist(dest1, dest3, parent(dest1), parent(dest3))
+      _     <- filesShouldExist(dest2)
+    } yield assertEquals(error, CopyOperationFailed(bucket, source, dest2))
+  }
+
+  test("successfully perform multiple copies") {
```
imsdu, line 37: We could complete this by setting the original as read-only and check that that those attributes are preserved

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFiles.scala
shinyhappydan: Also I don't know if it's necessarily helpful to check the list is nonempty here. I think practically speaking nothing bad will happen if an empty list is passed here. Sometimes it's easier in calling code if you just allow it and do nothing

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFiles.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  def parent(p: Path): Path = Path.fromNioPath(p.toNioPath.getParent)
+
+  private def copySingle(source: Path, dest: Path): IO[Unit] =
+    for {
+      _ <- Files[IO].createDirectories(parent(dest))
+      _ <- Files[IO].copy(source, dest, CopyFlags(CopyFlag.CopyAttributes))
+    } yield ()
+
+  private def rollbackCopiesAndRethrow(
```
imsdu, line 37: Not sure you can rollback if you copy read-only files

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/routes/StorageRoutes.scala
```diff
@@ -33,38 +34,54 @@ class StorageRoutes()(implicit storages: Storages[AkkaSource], hc: HttpConfig) {
...
-                      case Right((source, Some(_))) => complete(HttpEntity(`application/octet-stream`, source))
-                      case Right((source, None))    => complete(HttpEntity(`application/x-tar`, source))
-                      case Left(err)                => complete(err)
+                  )
+                }
+              },
+              operationName(s"/${hc.prefix}/buckets/{}/files") {
+                post {
+                  // Copy files within protected directory
+                  entity(as[NonEmptyList[CopyFile]]) { files =>
```
imsdu, line 75: The thing that bugs me is that copying a lot of files complicates things as it makes it more likely to hit some timeouts

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFileSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+    val bucket          = genString
+    val source1         = tempDir / genString / "file.txt"
+    val source2         = tempDir / genString / "file.txt"
+    val dest1           = tempDir / genString / "file.txt"
+    val dest2           = tempDir / genString / "file.txt"
+    val dest3           = tempDir / genString / "file.txt"
+    val files           = NonEmptyList.of(CopyBetween(source1, dest1), CopyBetween(source2, dest2), CopyBetween(source1, dest3))
+
+    for {
+      _ <- givenAFileAtPathWithContents(source1, source1Contents)
```
shinyhappydan, line 49: consider:

```
(source1, source1Contents) <- givenThereIsAFile
```
For less noise

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFileSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+      _ <- givenAFileAtPathWithContents(source2, source2Contents)
+      _ <- CopyFiles.copyAll(bucket, files)
```
shinyhappydan, line 51: I don't love passing the bucket here just for the error, is it really useful?

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/DiskStorageSpec.scala
```diff
@@ -317,6 +330,84 @@ class DiskStorageSpec
       }
     }
 
+    "copying" should {
```
imsdu, line 333: We could complete this by setting the original as read-only and check that that those attributes are preserved

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFileSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+      _ <- filesShouldExistWithContents(source1Contents, dest1, dest3)
+      _ <- filesShouldExistWithContents(source2Contents, dest2)
```
shinyhappydan, line 53: Maybe change this to take a single file and just repeat is clearer? Saying files when you're passing only one reads wrong

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/routes/StorageRoutes.scala
dantb: Hm yeah we could keep track of this and add a limit in delta if it's a problem - do we have metrics for individual endpoints? If not I could add logging

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/DiskStorageSpec.scala
dantb: Done in https://github.com/BlueBrain/nexus/pull/4511/commits/b774a6e488fcfb1f7f709a71775f158ad1b02668

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFileOutput.scala
dantb: Yeah I was trying to figure that out - it seems in delta we take the absolute path from the `location` field of a regular move / create remote file response. I wasn't sure whether it can be determined without it being returned? So I figured stick to the same pattern 🤷

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFiles.scala
dantb: The problem here was that `ValidatedCopyFile` is intentionally hard to construct, and tests for copying shouldn't need to care about the validation logic.

On the NEL point I just tend to prefer a stricter data structure when possible. It communicates more information. Once the parsing into a stricter structure is done at the edges then assumptions we have about that data are explicit in the type rather than implicit. In this case I'd rather reject the request in the endpoint for example

storage/src/test/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFileSpec.scala
dantb: 💯

storage/src/main/scala/ch/epfl/bluebrain/nexus/storage/files/CopyFiles.scala
dantb: I generally agree, it was also because I had my eye on reusing this code in delta so it will be moved to the kernel. The answer is probably to change the abstraction to use that function. I'll do that when I move it in the delta change ?

imsdu (approved):

PR #4522, opened 2023-11-22
Introduce munit-cats-effect

### LINKED ISSUE #4513: Introduce Cats-Effect munit
**Motivation**
We needed to duplicate some of its classes into our testkit while we depended on Cats Effect 2.

Now that the migration to Cats Effect 3 is complete, we can now introduce it as a dependency and remove those duplicated classes

shinyhappydan:
Fixes #4513


delta/kernel/src/test/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/ClasspathResourceLoaderSpec.scala
```diff
@@ -1,66 +1,65 @@
...
-    "fail when resource does not exists" in {
-      reject(loader.contentOf("resource2.txt", "value" -> "v")).getMessage should (include("not found") and include(
-        "resource2.txt"
-      ))
-    }
+  test("fail when a resource does not exist") {
+    loader
+      .contentOf("resource2.txt", "value" -> "v")
+      .intercept[Throwable]
+      .assert(e => e.getMessage.contains("not found") && e.getMessage.contains("resource2.txt"))
```
dantb, line 63: Not a new change and I'm not sure about asserting on library errors anyway, but if we are gonna do that could do something like

```suggestion
      .interceptMessage[IOException]("Resource resource2.txt not found")
```

dantb (approved):

delta/kernel/src/test/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/ClasspathResourceLoaderSpec.scala
shinyhappydan: I wrote it this way because the message changed slightly when I switched libraries and I wanted the assertion to be a bit looser (say that it wasn't found, say the name of the thing)

imsdu (approved):

PR #4528, opened 2023-11-23
Use IO.blocking in HttpClient, Storage and integration with Akka when needed

imsdu:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClient.scala
```diff
@@ -76,7 +76,7 @@ object HttpClient {
 
   private[http] object HttpSingleRequest {
     def default(implicit as: ActorSystem): HttpSingleRequest =
-      (request: HttpRequest) => IO.fromFuture(IO.delay(Http().singleRequest(request)))
+      (request: HttpRequest) => IO.fromFuture(IO.blocking(Http().singleRequest(request)))
```
shinyhappydan, line 79: It feels like this might not be blocking? If you take out the future, which should be dealt with by `fromFuture`?

shinyhappydan (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/http/HttpClient.scala
dantb: Think you're right, you'd hope `fromFuture` deals with the fact that this may be run on an external threadpool like Akka's one

imsdu: I wasn't worried about the future itself but things that could surround it.
But it is true that Akka in particular should handle this correctly and in the implementation, it just send a message mode so it should be fine

delta/plugins/jira/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/jira/JiraClient.scala
```diff
@@ -146,7 +146,7 @@ object JiraClient {
                             case None                      => IO.raiseError(NoTokenError)
                             case Some(_: AccessToken)      => IO.raiseError(RequestTokenExpected)
                             case Some(RequestToken(value)) =>
-                              IO {
+                              IO.blocking {
```
shinyhappydan, line 149: It might be clearer to put only the execute part in blocking, it was confusing initially

shinyhappydan (dismissed):

olivergrabinski (approved):

PR #4530, opened 2023-11-28
Add un-deprecate project operation

### LINKED ISSUE #4507: Allow to undeprecate a project
**Motivation**
A project can be deprecated by mistake or people may change their mind.
Currently when this happens, it has to be patched in database which is also error prone

**Acceptance criteria**
1. An new endpoint allows to undeprecate a project
2. Resources can be created/updated again in this project

shinyhappydan:
Fixes #4507 

image

### DIFF
--- a/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutesSpec.scala
+++ b/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutesSpec.scala
@@ -15,7 +15,8 @@ import ch.epfl.bluebrain.nexus.delta.sdk.identities.model.Caller
 import ch.epfl.bluebrain.nexus.delta.sdk.implicits._
 import ch.epfl.bluebrain.nexus.delta.sdk.organizations.model.Organization
 import ch.epfl.bluebrain.nexus.delta.sdk.organizations.model.OrganizationRejection.{OrganizationIsDeprecated, OrganizationNotFound}
-import ch.epfl.bluebrain.nexus.delta.sdk.permissions.Permissions.{events, projects => projectsPermissions, resources}
+import ch.epfl.bluebrain.nexus.delta.sdk.permissions.Permissions.{projects => projectsPermissions, resources}
+import ch.epfl.bluebrain.nexus.delta.sdk.permissions.model.Permission
 import ch.epfl.bluebrain.nexus.delta.sdk.projects.Projects.FetchOrganization
 import ch.epfl.bluebrain.nexus.delta.sdk.projects.model.ProjectRejection.WrappedOrganizationRejection
 import ch.epfl.bluebrain.nexus.delta.sdk.projects.model._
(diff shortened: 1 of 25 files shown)


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ProjectsSpec.scala
```diff
@@ -0,0 +1,5 @@
+package ch.epfl.bluebrain.nexus.tests.kg
+
+import ch.epfl.bluebrain.nexus.tests.BaseIntegrationSpec
+
+class ProjectsSpec extends BaseIntegrationSpec {}
```
imsdu, line 5: To be deleted

docs/src/main/paradox/docs/delta/api/projects-api.md
```diff
@@ -185,6 +185,29 @@ Request
 
 Response
 :   @@snip [deprecated.json](assets/projects/deprecated.json)
+~~~~
+## Undeprecate
```
imsdu, line 189: We also need a paragraph in the release notes

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutesSpec.scala
```diff
@@ -534,11 +538,73 @@ class ProjectsRoutesSpec extends BaseRouteSpec with IOFromMap {
...
+      val (org, project) = thereIsAProject
+      Put(s"/v1/projects/$org/$project/undeprecate?rev=1") ~> as(superUser) ~> routes ~> check {
+        status shouldEqual StatusCodes.BadRequest
+        response.asJson shouldEqual jsonContentOf(
+          "projects/errors/project-not-deprecated.json",
+          "org"  -> org,
+          "proj" -> project
+        )
+      }
+    }
```
olivergrabinski, line 566: Is it fair to say this does not check that the project is actually still active? It should also be possible to test this in `ProjectsImplSpec` instead and leave this as an error serialization test only

shinyhappydan: Good spot :)

I was trying to avoid changing `ProjectsImplSpec` though as I was anticipating the same issues I had with this one. Do you think it's worth the investment?

imsdu (dismissed):

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/ProjectsRoutesSpec.scala
olivergrabinski: I don't think it needs to be done here. But in my experience the ImplSpec were easier to work with than the Route ones

imsdu: Maybe something in the integration tests ?

shinyhappydan: I purposely didn't add edge cases to the integration tests as they are so much slower, and now really better than testing at this level. The integration test has the "happy" path only, which seems like the correct way to test given that these route specs are almost end-to-end tests

imsdu (approved):

PR #4531, opened 2023-11-28
Add a way to reset all default Elasticsearch views

### LINKED ISSUE #4410: Add a way to migrate default elasticsearch views after 1.8 changes
**Motivation**
Mapping and settings of default elasticsearch views have changed, we need a way to update the existing ones and trigger the reindexing including for the deprecated projects

**Acceptance criteria**
1. All views are recreated/updated with the new mapping/setting
2. Data is reindexed

olivergrabinski:
Closes #4410


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchDefaultViewsResetter.scala
```diff
@@ -0,0 +1,110 @@
...
+            case e                   =>
+              logger.warn(
+                s"There was an error when attempting to delete the default ES index in project '$project'. Reason: ${e.getMessage}"
+              )
+          }
+          .void
+
+      private def deleteEventsStatesOffsets(project: ProjectRef): doobie.ConnectionIO[Unit] =
+        sql"""
+          DELETE FROM scoped_events WHERE type = 'elasticsearch' AND id = $defaultEsViewId AND org = ${project.organization} AND project = ${project.project};
```
imsdu, line 97: What about adding this operation in sourcing so that it can be reusable in other cases ?

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchDefaultViewsResetter.scala
```diff
@@ -0,0 +1,110 @@
...
+          DELETE FROM scoped_states WHERE type = 'elasticsearch' AND id = $defaultEsViewId AND org = ${project.organization} AND project = ${project.project};
+          DELETE FROM projection_offsets WHERE module = 'elasticsearch' AND resource_id = $defaultEsViewId AND project = $project;
```
imsdu, line 99: I know that would make two transactions instead of a single one but it is possible to use `Projections.reset`

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
```diff
@@ -136,6 +136,23 @@ final class ElasticSearchViews private (
...
+    * Creates a new ElasticSearchView without checking whether the provided project is deprecated or not.
+    *
+    * @param iri
+    *   id of the view to be created
+    * @param project
+    *   project in which the view has to be created
+    * @param value
+    *   configuration of the view to be created
+    */
+  def unsafeCreate(
```
imsdu, line 149: We have a similar method for deprecation called `internalDeprecate` so we could align those ?

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchDefaultViewsResetterSuite.scala
```diff
@@ -0,0 +1,189 @@
...
+
+  override def munitFixtures: Seq[AnyFixture[_]] = List(doobie)
+  private lazy val xas                           = doobie()
+
+  private val defaultEsViewId   = "https://bluebrain.github.io/nexus/vocabulary/defaultElasticSearchIndex"
+  private val defaultIndexLabel = IndexLabel.unsafe("default")
+  private val project           = ProjectRef.unsafe("org", "proj")
+
+  test("The reseter should delete scoped events for the default view") {
+    initPartitions(xas, project) >>
```
imsdu, line 33: The initPartitions should come in a fixture

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
olivergrabinski: Updated

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchDefaultViewsResetterSuite.scala
imsdu: As it is a setup thing, it would be better if it did not happen in the first test but we can address it in a later PR

olivergrabinski: Ideally it should be in beforeAll, but I'm running into `The fixture `doobie` was not instantiated` when doing that. So putting it as the first test for now...

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchDefaultViewsResetter.scala
```diff
@@ -0,0 +1,104 @@
...
+          case activeView: ActiveViewDef => deleteIndex(activeView.index)
+          case _: DeprecatedViewDef      => IO.pure(true)
+        }
+
+      private def deleteEventsStatesOffsets(project: ProjectRef): doobie.ConnectionIO[Unit] =
+        sql"""
+          DELETE FROM scoped_events WHERE type = 'elasticsearch' AND id = ${defaultEsViewId.toString} AND org = ${project.organization} AND project = ${project.project};
+          DELETE FROM scoped_states WHERE type = 'elasticsearch' AND id = ${defaultEsViewId.toString} AND org = ${project.organization} AND project = ${project.project};
+          DELETE FROM projection_offsets WHERE module = 'elasticsearch' AND resource_id = ${defaultEsViewId.toString} AND project = $project;
+        """.stripMargin.update.run.void
```
olivergrabinski, line 94: Given the fragile nature of the resetter tests, I prefer to hardcode the default view id in here so that we cannot delete something else by mistake. We should handle resource deletion properly in the future

imsdu (approved):

PR #4536, opened 2023-11-29
Resource registering error (#4532)

imsdu:
* added file_resource_mapping to avoid error when registering

* undid non-relevant changes (python version, etc...)

---------


NicoRicardi:
I tried to make this a draft because at least another notebook needs the same change. I could not make it a draft, perhaps because I am part of the organisation (yet)

PR #4544, opened 2023-11-30
Add operation to un-deprecate an organisation

### LINKED ISSUE #4508: Allow to undeprecate an org
**Motivation**
An org can be deprecated by mistake or people may change their mind.
Currently when this happens, it has to be patched in database which is also error prone

**Acceptance criteria**
1. An new endpoint allows to undeprecate an org
2. Resources can be created/updated again in this org

shinyhappydan:
Fixes #4508 

image

image


delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/OrganizationsRoutesSpec.scala
```diff
@@ -35,18 +33,34 @@ class OrganizationsRoutesSpec extends BaseRouteSpec with IOFromMap {
...
     Set(orgsPermissions.write, orgsPermissions.read)
   )
 
   private lazy val orgs                            = OrganizationsImpl(Set(aopd), config, xas, clock)
   private lazy val orgDeleter: OrganizationDeleter = id => IO.raiseWhen(id == org1.label)(OrganizationNonEmpty(id))
 
-  private val caller = Caller(alice, Set(alice, Anonymous, Authenticated(realm), Group("group", realm)))
-
-  private val identities = IdentitiesDummy(caller)
+  private val superUser                      = User("superUser", Label.unsafe(genString()))
```
imsdu, line 44: There is not really a need to generate a realm for each user

shinyhappydan: It's not necessary but it shows that the realm is not relevant to the test

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/OrganizationsRoutesSpec.scala
```diff
@@ -271,6 +293,93 @@ class OrganizationsRoutesSpec extends BaseRouteSpec with IOFromMap {
...
+
+      latestRevisionOfOrganization(org) should not(be(deprecated))
+    }
+
+    "fail to undeprecate an organisation without organizations/write permission" in {
+      val org = thereIsADeprecatedOrganization
+
+      Put(s"/v1/orgs/$org/undeprecate?rev=2") ~> routes ~> check {
+        response.shouldBeForbidden
+      }
```
olivergrabinski, line 313: Not sure it's too important but I would have a preference for using the user with org/read permission instead of anonymous

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/model/OrganizationRejection.scala
```diff
@@ -77,6 +77,15 @@ object OrganizationRejection {
   final case class OrganizationIsDeprecated(label: Label)
       extends OrganizationRejection(s"Organization '$label' is deprecated.")
 
+  /**
+    * Signals an attempt to update/deprecate an organization that is already deprecated.
```
olivergrabinski, line 81: ...an organization that is already active?

olivergrabinski (approved):

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/OrganizationsRoutesSpec.scala
shinyhappydan: Each of the other users only have one permission, so this should be safe

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/organizations/model/OrganizationRejection.scala
shinyhappydan: I'll fix this in another PR, good spot

PR #4565, opened 2023-12-05
Don’t allow updates of default views

### LINKED ISSUE #4549: Disallow the updates/deprecation of default elasticsearch and blazegraph views
**Motivation**
Those views are meant to have a consistent behaviour across projects.

In the case of the Elasticsearch views, they even power the listing operations.
So any writes operations on those default views must be prohibited to avoid errors and inconsistencies.

**Acceptance criteria**
1. When a user attempts any kind of update on a default Elasticsearch or Blazegraph views, an error is returned

shinyhappydan:
Fixes #4549

### DIFF
--- a/delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
+++ b/delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
@@ -134,6 +134,7 @@ final class BlazegraphViews(
     for {
       pc        <- fetchContext.onModify(project)
       iri       <- expandIri(id, pc)
+      _         <- validateNotDefaultView(iri)
       viewValue <- sourceDecoder(project, pc, iri, source)
       res       <- eval(UpdateBlazegraphView(iri, project, viewValue, rev, source, caller.subject))
       _         <- createNamespace(res)
--- a/delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
+++ b/delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
@@ -211,9 +211,14 @@ final class BlazegraphViews(
     for {
       pc  <- fetchContext.onModify(project)
       iri <- expandIri(id, pc)
+      _   <- validateNotDefaultView(iri)
       res <- eval(DeprecateBlazegraphView(iri, project, rev, subject))
     } yield res
   }.span("deprecateBlazegraphView")
+
+  private def validateNotDefaultView(iri: Iri): IO[Unit] = {
+    IO.raiseWhen(iri == defaultViewId)(ViewIsDefaultView)
+  }
 
   /**
     * Undeprecate a view.
--- a/delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
+++ b/delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
@@ -189,6 +189,7 @@ final class BlazegraphViews(
     for {
       pc  <- fetchContext.onModify(project)
       iri <- expandIri(id, pc)
+      _   <- validateNotDefaultView(iri)
       res <- eval(TagBlazegraphView(iri, project, tagRev, tag, rev, subject))
       _   <- createNamespace(res)
     } yield res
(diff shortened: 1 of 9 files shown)


delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
```diff
@@ -211,10 +211,15 @@ final class BlazegraphViews(
     for {
       pc  <- fetchContext.onModify(project)
       iri <- expandIri(id, pc)
+      _   <- validateNotDefaultView(iri)
```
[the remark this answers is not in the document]
shinyhappydan: Total brain fail, wrote the test but didn't make it pass 🤦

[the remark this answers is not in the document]
imsdu: It would be better to move it in the evaluate function so that it is not isolated from the other checks

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViews.scala
```diff
@@ -201,6 +201,7 @@ final class ElasticSearchViews private (
   )(implicit caller: Caller): IO[ViewResource] = {
     for {
       (iri, pc) <- expandWithContext(fetchContext.onModify, project, id)
+      _         <- validateNotDefaultView(iri)
```
imsdu, line 204: Same, it would be better to have it in the evaluate function

shinyhappydan: Yeah, looking at the other checks, they are all related to the state. Personally, I think it's clearer to put the check in the main method

shinyhappydan: Unfortunately this means I now need to make sure in the ElasticSearchViewsSpec that we create the default view (currently we don't do this)

shinyhappydan: This ended up causing design issues around our project deletion code. We're going to merge what we have now, and when you're back I can show you what we came up against. I still have the code so it'll be easy to pursue if we can figure out a way forward

delta/plugins/blazegraph/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViewsSpec.scala
```diff
@@ -418,6 +418,21 @@ class BlazegraphViewsSpec extends CatsEffectSpec with DoobieScalaTestFixture wit
...
+      }
+
+      "updating" in {
+        views.update(defaultViewId, projectRef, 1, indexingSource).rejected shouldEqual ViewIsDefaultView
+      }
+
+      "tagging" in {
+        views.tag(defaultViewId, projectRef, tag, tagRev = 1, 1).rejected shouldEqual ViewIsDefaultView
+      }
+    }
```
olivergrabinski, line 434: Shouldn't these then assert that the actions have in fact not been performed

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/ElasticSearchViewsSpec.scala
```diff
@@ -535,6 +535,22 @@ class ElasticSearchViewsSpec extends CatsEffectSpec with DoobieScalaTestFixture
...
+
+      "updating" in {
+        val source = json"""{"@type": "ElasticSearchView", "mapping": $mapping}"""
+        views.update(defaultViewId, projectRef, 1, source).rejectedWith[ViewIsDefaultView]
+      }
+
+      "tagging" in {
+        views.tag(defaultViewId, projectRef, UserTag.unsafe("tag"), 1, 1).rejectedWith[ViewIsDefaultView]
+      }
+    }
```
olivergrabinski, line 552: Shouldn't these then assert that the actions have in fact not been performed

olivergrabinski (approved):

delta/plugins/blazegraph/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViewsSpec.scala
olivergrabinski: I guess this should be covered elsewhere

PR #4575, opened 2023-12-07
Add query, files and tests for morphology features

crisely09:
This is to index the morphology features that exist as separate annotations.

Following what was requested in [this ticket](https://bbpteam.epfl.ch/project/issues/browse/BBPP134-856):
1. separate axon and dendritic total lengths not just total length of the whole morphology
2. maximum branch order (Strahler index) [for axon, apical/basal dendrites]
3. xyz spatial extent of dendrites (bbox)
4. soma radius
5. arbor asymmetry (partition asymmetry) for apical and basal dendrites


shinyhappydan (approved):

crisely09:
@shinyhappydan how is that the integration tests didn't run?

imsdu:
Could you check the query is valid ?
If it is not it is possible it prevents Delta to start

tests/docker/config/construct-query.sparql
```diff
@@ -657,4 +710,115 @@ CONSTRUCT {
                                                   OPTIONAL { ?preSynaptic     skos:notation   ?preSynapticNotation    . } .
 
     } .
+
+  # Morphology features
+  OPTIONAL {
```
imsdu, line 715: This is only for neuron morphologies or a subset of types right ?
If yes, we could add something like:
```
?id  a           nsg:NeuronMorphology ;
```
to act as a short circuit so that the query is faster for the other types ?

crisely09: Yes, this is a good idea.

tests/docker/config/construct-query.sparql
```diff
@@ -657,4 +665,25 @@ CONSTRUCT {
...
+      GRAPH ?g {
+        ?featureAnnotation    bmo:compartment       ?featureCompartment ;
+                              nsg:hasBody           ?featureBody        .
+        ?featureBody          nsg:isMeasurementOf / rdfs:label ?featureLabel ;
+                              nsg:series            ?featureSeries .
+        ?featureSeries        nsg:statistic         ?featureStat   ;
+                              schema:value          ?featureValue  ;
+                              schema:unitCode       ?featureUnit   .
+
+        FILTER (?featureLabel IN ("Soma Radius", "Total Length", "Total Weight", "Total Depth", "Section Strahler Orders", "Partition Asymmetry")) .
```
olivergrabinski, line 685: I'm not 100% sure on this one, but I think if the filter could be done as soon as `?featureLabel` is available to avoid looking trying to look up series and annotation? I don't remember if that's how it plans the query

tests/docker/config/construct-query.sparql
```diff
@@ -657,4 +665,25 @@ CONSTRUCT {
...
     } .
+
+  # Morphology features
+  OPTIONAL {
+    ?id a nsg:NeuronMorphology  ;
+          ^nsg:hasSource / ^nsg:hasTarget ?featureAnnotation    .
+    ?featureAnnotation   a   bmo:NeuronMorphologyFeatureAnnotation  .
+
+    OPTIONAL {
+      GRAPH ?g {
```
olivergrabinski, line 676: I think we can get rid of the `GRAPH` here

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigIndexingSpec.scala
```diff
@@ -871,6 +876,188 @@ class SearchConfigIndexingSpec extends BaseIntegrationSpec {
...
+    "have the correct morphology mean length" in {
+      val query    = queryField(neuronMorphologyId, "morphologyFeature")
+      val expected =
+        json"""[
+                {
+                 "value" : 1131.963147431612,
+                 "unit" : "μm",
+                 "label" : "Total Length",
+                 "statistic" : "mean",
+                 "compartment" : "NeuronMorphology",
```
olivergrabinski, line 889: The comma on this line will cause the test to fail

olivergrabinski (commented):
If I understand correctly, the idea is that a neuron morphology will link to different `NeuronMorphologyFeatureAnnotation`; i.e. we want these annotations to appear in the morphology docs indexed in ES.

Then I think several changes will have to be made:

* The annotations are not "main resources", they are not indexed as their own documents, but are needed by other resources. So the references to the new annotations should be moved from `mainResources` to `otherResources`. Not great I admit but so far this is how we keep this difference in this test. This change will make the `index all data` test pass.
* Then we can focus on the tests. Right now you are querying the available `neuronMorphologyId`, and expecting to find some annotations. But then we need to also modify the neuron morphology file to reference these annotations.
* We might also be in a situation again where the queries that will be made require the fields to be indexed to be set as nested; maybe you can check what kind of queries will be made.

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigIndexingSpec.scala
crisely09: comma removed

tests/docker/config/construct-query.sparql
crisely09: move it higher in the query

tests/docker/config/construct-query.sparql
```diff
@@ -657,4 +665,24 @@ CONSTRUCT {
...
+  OPTIONAL {
+    ?id a nsg:NeuronMorphology  ;
+          ^nsg:hasSource / ^nsg:hasTarget ?featureAnnotation    .
+    ?featureAnnotation   a   bmo:NeuronMorphologyFeatureAnnotation  .
+
+    OPTIONAL {
+      ?featureAnnotation    bmo:compartment                    ?featureCompartment ;
+                            nsg:hasBody                        ?featureBody        .
+      ?featureBody          nsg:isMeasurementOf / rdfs:label   ?featureLabel       .
+      FILTER (?featureLabel IN ("Soma Radius", "Total Length", "Total Weight", "Total Depth", "Section Strahler Orders", "Partition Asymmetry")) .
```
ssssarah, line 679: Missing Total Width

Issue #4350, opened 2023-10-10
Improve full text search for default views


imsdu:
**Motivation**
Performing full text search in the Fusion project page does not bring satisfying results due to the way data is indexed. We search only one field, `allFields` - [here](https://github.com/BlueBrain/nexus/blob/master/delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/QueryBuilder.scala#L76) in the QueryBuilder. This contains all data so fields that have more weight such as id/name/description gets diluted by everything else.

A simple example of the problem is that passing a resource Id in the search query can produce mixed results, such as other resources linked to it. It should instead return that resource as the top result. The same goes for querying the name of a resource.

In the API this is supported by the `q` parameter on list operations [here](https://bluebrainnexus.io/docs/delta/api/resources-api.html#within-a-project). In Fusion it's in the default resource listing here: https://bbp.epfl.ch/nexus/web/orgs/public/sscx

The code change should be very simple - just modifying the `QueryBuilder`. Most of the work will be experimenting with indexing and querying of elastic search based on different tokenizers / analyzers, such as:
- For iris - https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-uaxurlemail-tokenizer.html
- For natural language fields like description - https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenizer.html
- For querying across multiple fields - https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html

**Acceptance criteria**
1. Description is indexed as an independent field.
2. The tokenizer/analyzer is changed so as to better index the resource Id, for example.
3. The full text search part of the listing operation (`q` param) is changed to target several fields (already done in Global search)

dantb:
@imsdu might pick up this next, although it might be more suited to someone with domain knowledge... Would it be ok to allow feeding the weightings through an endpoint with a small UI? Otherwise the feedback loop is long and it'll be hard to iterate towards a "better" search

Also who uses the search feature? I.e. who can verify that the results are better

imsdu:
The idea is not to break retro-compatibility and not to introduce any weighting for now but to change the full-text search part of the listing operation (the q param) so that it does not target only the `allFields` fields but also the other fields like the id, the name, the type that are defined in the default mapping.
That will be a first good step as the default TF/IDF algorithm should already help a lot.
[...]

PR #4582, opened 2023-12-13
Improve full text search

olivergrabinski:
#4350

* Fields that reasonably expect an IRI (like @id) use the path_hierarchy analyzer
* Fields label, name, description use an "autocomplete" analyzer
* Adding `*.fulltext` multi-field to replace `_all_fields`.


delta/plugins/elasticsearch/src/main/resources/defaults/default-mapping.json
```diff
@@ -1,170 +1,141 @@
 {
   "properties": {
     "@type": {
-      "type": "keyword",
```
imsdu, line 4: This is introducing a breaking change, we need to introduce a multi-field where the full text search would be run in the subfield.
It may make the full text search easier as we can only check on those fields (*.fulltext maybe) ?

delta/plugins/elasticsearch/src/main/resources/defaults/default-mapping.json
```diff
@@ -1,170 +1,141 @@
...
-      "copy_to": "_all_fields"
+      "type": "text"
     },
     "_filename": {
-      "type": "keyword",
-      "copy_to": "_all_fields"
+      "type": "text"
     },
     "_digest": {
       "type": "nested",
```
imsdu, line 105: We can get rid of the nested there, it is not needed

delta/plugins/elasticsearch/src/main/resources/defaults/default-mapping.json
olivergrabinski: The nice thing about this is that we no longer need to match on `_*` but `*.fulltext` and so we don't match on ES metadata

delta/plugins/elasticsearch/src/main/resources/defaults/default-settings.json
```diff
@@ -1,14 +1,45 @@
...
+      },
+      "autocomplete_search": {
+        "tokenizer": "lowercase"
+      }
+    },
+    "tokenizer": {
+      "autocomplete": {
+        "type": "edge_ngram",
+        "min_gram": 2,
+        "max_gram": 10,
```
imsdu, line 35: We may have longer words, maybe we should extend that a bit

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/QueryBuilder.scala
```diff
@@ -137,14 +137,20 @@ final case class QueryBuilder private[client] (private val query: JsonObject) {
...
     JsonObject("terms" -> Json.obj(k -> values.asJson))
 
-  private def matchPhrasePrefix[A: Encoder](k: String, value: A): JsonObject =
-    JsonObject("match_phrase_prefix" -> Json.obj(k -> Json.obj("query" -> value.asJson)))
+  private def multiMatch[A: Encoder](value: A): JsonObject                  =
+    JsonObject(
+      "multi_match" -> Json.obj(
+        "query"  -> value.asJson,
+        "type"   -> Json.fromString("phrase_prefix"),
+        "fields" -> json"""[ "@id.fulltext^10", "*", "*.fulltext", "_tags", "_original_source", "_uuid" ]"""
```
imsdu, line 151: description is missing here

olivergrabinski: From what I tried and what I understand it will be matched by `*`. Ideally I would have liked to have just `"@id.fulltext^10", "*", "*.fulltext"`. The issue with `"_tags", "_original_source", "_uuid"` is that they were already `text` fields; but they are not caught by the `*` (because they start with `_`), and they cannot be caught by `*.fulltext` because you cannot add a multi-field which is a duplicate of the main type

delta/plugins/elasticsearch/src/main/resources/defaults/default-settings.json
olivergrabinski: Extended it to 20

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/QueryBuilder.scala
olivergrabinski: There is a test now that matches on the description that shows it works

olivergrabinski: I also removed the multiplier on `@id`, because I don't think we need it now that IRIs use the path_hierarchy search analyzer

delta/plugins/elasticsearch/src/main/resources/defaults/default-settings.json
```diff
@@ -1,14 +1,45 @@
 {
   "analysis": {
     "analyzer": {
-      "nexus": {
+      "default": {
```
imsdu, line 4: As the default settings, may be used by custom views, we need to keep a reference to the nexus analyzer

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ListingsSpec.scala
```diff
@@ -382,4 +388,87 @@ final class ListingsSpec extends BaseIntegrationSpec {
 
   }
 
+  "Fulltext search" should {
```
dantb, line 391: Nice tests 💯

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ListingsSpec.scala
```diff
@@ -382,4 +388,87 @@ final class ListingsSpec extends BaseIntegrationSpec {
...
+          val results = listing._results.getOption(json).value
+          results.head should have(`@id`(id))
+        }
+      }
+    }
+
+    "find match by description" in {
+      val id       = s"http://bbp.epfl.ch/${genString()}"
+      val resource =
+        json"""{ "@id": "$id", "http://schema.org/description": "Northumberland is a ceremonial county bordering Scotland."}"""
```
dantb, line 450: For understanding, why do name and description need their full context?

olivergrabinski: This resource has no context, so if we put `"description"` then it's not going to be `http://schema.org/description` but instead it's going to be using the vocab defined in the project. If the non schema.org is pushed to ES then the fields is not `"description"` and the mapping does not apply to it.

Usually, in prod projects this is not specified in each resource, instead the context is defined in a separate resource and just referenced by `@id`.

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ListingsSpec.scala
olivergrabinski: @shinyhappydan MVP

delta/plugins/elasticsearch/src/main/resources/defaults/default-settings.json
olivergrabinski: I've made the change. The `_createdAt` and `_updatedAt` use the `nexus` analyzer but they should probably be changed to use a pattern one because `1970-01-01T00:00:00Z` is tokenized as follows

```
1970-01-01t00, 1970, 01, 01t00, 00, 00z
```

imsdu (approved):

PR #4583, opened 2023-12-14
Pass only the show location part of storages config for file encoding

dantb:
Noticed while creating a new route that file encoding requires only a small part of the config. This makes that more explicit / helps with testing


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesConfig.scala
```diff
@@ -63,12 +63,16 @@ object StoragesConfig {
...
-      tpe match {
-        case StorageType.DiskStorage       => Some(disk)
-        case StorageType.S3Storage         => amazon
-        case StorageType.RemoteDiskStorage => remoteDisk
-      }
+    def showFileLocation: ShowFileLocation =
+      ShowFileLocation(disk.showLocation, remoteDisk.exists(_.showLocation), amazon.exists(_.showLocation))
+  }
+
+  final case class ShowFileLocation(disk: Boolean, remote: Boolean, s3: Boolean) {
```
imsdu, line 70: maybe a set of storage type instead of these 3 booleans ?
And then, you call contains ?

dantb: Ok, done

imsdu (approved):

PR #4608, opened 2023-12-20
Remove the ability to fetch all events as SSEs

### LINKED ISSUE #4601: Remove the ability to fetch all events as SSE
**Motivation**
This endpoint allows to fetch all events as SSEs (realm, acl, resource) in a single endpoint.

It is not used and introduces complexity and coupling between all the different kind of events.

It has been marked as deprecated in 1.9 so that we can get rid of it instead of making it an extra task in the Akka migration

** Acceptance criteria **
* Code and tests are removed
* Documentation is updated

olivergrabinski:
Closes #4601


docs/src/main/paradox/docs/releases/v1.10-release-notes.md
```diff
@@ -46,6 +47,10 @@ Previously deprecated storages can now be undeprecated.
 
 @ref:[More information](../delta/api/storages-api.md#undeprecate)
 
+### Deprecations
```
imsdu, line 50: This would be more under a SSE part

olivergrabinski: 👍

imsdu (approved):

PR #4613, opened 2023-12-21
Remove SSEs for global entities

### LINKED ISSUE #4602: Remove SSEs for global entities
**Motivation**
Those are unused endpoints and instead of migrating them to the Akka alternative, it is better to get rid of them and the complexity they are introducing.

**Acceptance criteria**
1. Code is removed and refactored
2. Documentation is updated

olivergrabinski:
Closes #4602


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/sse/SseEventLog.scala
```diff
@@ -84,15 +84,10 @@ trait SseEventLog {
...
 
-  /**
-    * Return all SSE selectors
-    */
-  def allSelectors: Set[Label]
-
   /**
     * Returns SSE selectors related to ScopedEvents
     */
-  def scopedSelectors: Set[Label]
```
olivergrabinski, line 95: I don't think there is a reason to keep this distinction

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/EventsRoutes.scala
```diff
@@ -90,18 +83,17 @@ class EventsRoutes(
...
-                      operationName(s"$prefixSegment/$selector/{org}/{proj}/events") {
-                        authorizeFor(project, events.read).apply {
-                          emit(sseEventLog.streamBy(selector, project, offset).attemptNarrow[ProjectRejection])
-                        }
-                      },
-                      (head & authorizeFor(project, events.read)) {
-                        complete(OK)
+                (resolveSelector & projectRef & pathPrefix("events") & pathEndOrSingleSlash) { (selector, project) =>
+                  concat(
+                    operationName(s"$prefixSegment/$selector/{org}/{proj}/events") {
```
imsdu, line 88: The operation names can go away because we will keep kamon disabled for this part

delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/AclsModule.scala
```diff
@@ -61,8 +59,6 @@ object AclsModule extends ModuleDef {
 
   many[ProjectDeletionTask].add { (acls: Acls) => Acls.projectDeletionTask(acls) }
 
-  many[SseEncoder[_]].add { base: BaseUri => AclEvent.sseEncoder(base) }
```
imsdu, line 64: The encoder itself in AclEvent can go away too

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/sse/SseEncoder.scala
```diff
@@ -7,15 +7,11 @@ import ch.epfl.bluebrain.nexus.delta.sourcing.event.Event.ScopedEvent
 import ch.epfl.bluebrain.nexus.delta.sourcing.model.{EntityType, Label, ProjectRef}
 import io.circe.{Decoder, Encoder, JsonObject}
 
-import scala.reflect.ClassTag
-
-abstract class SseEncoder[E <: Event](implicit c: ClassTag[E]) {
```
imsdu, line 12: We could reduce it to ScopedEvent

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/sse/SseEncoder.scala
```diff
@@ -7,15 +7,11 @@ import ch.epfl.bluebrain.nexus.delta.sourcing.event.Event.ScopedEvent
...
+abstract class SseEncoder[E <: Event] {
```
imsdu, line 10: We could reduce it to ScopedEvent

delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/routes/EventsRoutesSpec.scala
```diff
@@ -32,17 +32,14 @@ class EventsRoutesSpec extends BaseRouteSpec {
   private val identities = IdentitiesDummy(caller)
   private val asAlice    = addCredentials(OAuth2BearerToken("alice"))
 
-  private val acl       = Label.unsafe("acl")
   private val project   = Label.unsafe("project")
-  private val resources = Label.unsafe("resources")
+  private val resources = Label.unsafe("resource")
```
imsdu, line 36: The `resource` selector does not exist, it is just `resources` that must return all resource event (including schema, resolver, etc...)
Might be worth to check if it still works and if it is tested

olivergrabinski: It's just that here the implementation of streamBy is filtering based on the type defined when defining the SSE, and it was `Project` and `Resource`. In fact there were several tests that were passing before but for the wrong reasons. And `project` was also in singular. Anyway, I will just make it plural to avoid any confusion.

Also I have changed the ACL test to be a test on resources sse so that will be covered. I will check in the integration tests

imsdu (approved):

PR #4615, opened 2023-12-21
Refactor integration tests for storages & files

### LINKED ISSUE #4614: Rework integration tests for storages and files
- Create DSLs for storages and files, to be consistent with the other DSLs
- Remove duplication
- Keep related things in a common package

dantb:
1. Created `StoragesDsl` and `FilesDsl` for common parts and removed some duplication. They're probably not strictly DSLs since they do assertions and suchlike but named that for consistency 🤷 
2. New package for everything related to files (maybe archives should be in there?)
3. Made `BatchFilesSpec` tests independent from the other storages tests and eachother - e.g. we now mk/delete buckets each time

Fixes #4614


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/RemoteStorageSpec.scala
```diff
@@ -61,60 +54,26 @@ class RemoteStorageSpec extends StorageSpec with CopyFilesSpec {
...
-               "storageBase" -> externalEndpoint,
-               "id"          -> storId
-             )
-             filterKey("credentials")(json) should equalIgnoreArrayOrder(expected)
-
-           }
-      _ <- permissionDsl.addPermissions(
-             Permission(storName, "read"),
-             Permission(storName, "write")
-           )
```
dantb, line 109: I removed these calls for each storage type tests and nothing failed... So I guess they're not needed, but I don't know why?

imsdu: Maybe a bug ?
When you create a storage, you can define custom permissions and those permissions need to have been previoully defined.

dantb: This was the issue - I nuked the DB and it correctly failed afterwards. Fixed now. I have to remember there's no cleanup so the tests aren't independent 😅  maybe with generated data everywhere they would be unlikely to overlap

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/BatchFilesSpec.scala
```diff
@@ -0,0 +1,188 @@
...
+import ch.epfl.bluebrain.nexus.tests.kg.files.model.FileInput._
+import ch.epfl.bluebrain.nexus.tests.HttpClient._
+import ch.epfl.bluebrain.nexus.tests.Identity.storages.Coyote
+import ch.epfl.bluebrain.nexus.tests.kg.files.BatchFilesSpec.{CopyStorageType, Response, StorageDetails}
+import ch.epfl.bluebrain.nexus.tests.{BaseIntegrationSpec, Optics}
+import io.circe.syntax.KeyOps
+import io.circe.{Decoder, DecodingFailure, Json, JsonObject}
+import org.scalatest.Assertion
+
+class BatchFilesSpec extends BaseIntegrationSpec {
```
shinyhappydan, line 19: I think it's at least batch copy?

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/FilesDsl.scala
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Base64
+import scala.concurrent.ExecutionContext
+
+class FilesDsl(deltaClient: HttpClient)(implicit mat: Materializer, ec: ExecutionContext)
+    extends CirceUnmarshalling
+    with Matchers
+    with OptionValues
+    with ScalaFutures {
+
+  def uploadFile(
```
shinyhappydan, line 27: The assert function being a parameter here makes this harder to compose, it's really better to just map to an assertion

shinyhappydan: This might be better referred to as creating a file

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/StorageSpec.scala
```diff
@@ -84,41 +68,41 @@ abstract class StorageSpec extends BaseIntegrationSpec {
...
 
     "be uploaded" in {
-      uploadFile(jsonFileNoContentType, None)(expectCreated)
+      filesDsl.uploadFile(jsonFileNoContentType, projectRef, storageId, None)(expectCreated)
     }
 
     "be downloaded" in {
       deltaClient.get[ByteString](s"/files/$projectRef/attachment:attachment.json", Coyote, acceptAll) {
-        expectDownload("attachment.json", ContentTypes.`application/json`, jsonFileContent)
+        filesDsl.expectDownload("attachment.json", ContentTypes.`application/json`, jsonFileContent)
```
shinyhappydan, line 89: I realise the name is the same as before, but `expectDownload` isn't great, and the hiding of the response and headers isn't helpful to the reader

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/StoragesDsl.scala
```diff
@@ -0,0 +1,91 @@
...
+class StoragesDsl(deltaClient: HttpClient) extends CirceUnmarshalling with Matchers with CirceEq {
+
+  private val loader = ClasspathResourceLoader()
+
+  def createStorage(payload: Json, projectRef: String): IO[Assertion] =
+    deltaClient.post[Json](s"/storages/$projectRef", payload, Coyote) { (_, response) =>
+      response.status shouldEqual StatusCodes.Created
+    }
+
+  def createDiskStorageDefaultPerms(id: String, projectRef: String): IO[Assertion] =
```
shinyhappydan, line 25: I think this needs to say `createDiskStorageWithDefaultPerms` because it sounds like you're creating the default permissions

shinyhappydan: (Same with the other methods)

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/DiskStorageSpec.scala
```diff
@@ -33,34 +32,19 @@ class DiskStorageSpec extends StorageSpec with CopyFilesSpec {
...
-      _ <- deltaClient.post[Json](s"/storages/$projectRef", payload2, Coyote) { (_, response) =>
-             response.status shouldEqual StatusCodes.Created
-           }
-      _ <- deltaClient.get[Json](s"/storages/$projectRef/nxv:$storageId2", Coyote) { (json, response) =>
-             val expected = storageResponse(projectRef, storageId2, storage2Read, storage2Write)
-             filterMetadataKeys(json) should equalIgnoreArrayOrder(expected)
-             response.status shouldEqual StatusCodes.OK
-           }
+      _ <- storagesDsl.createDiskStorageDefaultPerms(storId, projectRef)
+      _ <- storagesDsl.checkStorageMetadata(projectRef, storId, expectedStorage)
```
shinyhappydan, line 44: I thinnk this would be clearer as `storagesDsl.storageMetadata(project, storgage) shouldBe expectedStorage`, right now it's a bit unclear what the check is

shinyhappydan (approved):

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/FilesDsl.scala
dantb: I agree - working with this style is painful. When I went down this rabbit hole it ended up touching pretty much all the integration tests 😓 this change is mostly just trying to group together storage-related things

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/DiskStorageSpec.scala
dantb: Another thing caused by the assertion passing pattern 😞 maybe it's worth refactoring all the DSLs?

shinyhappydan: Ha well it would be nice but also probably an exercise in pain

PR #4616, opened 2023-12-21
Improve synchronous indexing

### LINKED ISSUE #4556: Improve synchronous indexing
**Motivation**
When a user opts for synchronous indexing, the resource is pushed to the different views defined in its project and a response is returned to the user only when the indexing completes

**Acceptance criteria**
1. If indexing for one view fails, it should not prevent other views to perform the indexing
2. The error should be an aggregation of the different errors
3. The indexing in the different views should be done again in parallel and using fibers like the regular indexing

imsdu:
Fixes #4556 

+ Fix flaky tests in `SupervisorSuite`


delta/sourcing-psql/src/test/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/SupervisorSuite.scala
```diff
@@ -48,30 +48,31 @@ class SupervisorSuite extends NexusSuite with SupervisorSetup.Fixture with Doobi
...
-        CompiledProjection.fromStream(
-          metadata,
-          strategy,
-          evalStream(started.set(true)).map(_ >> Stream.raiseError[IO](expectedException))
-        )
-      _         <- sv.run(projection).assertEquals(Running).eventually
-      _         <- started.get.assertEquals(true).eventually
+      started       <- Ref.of[IO, Boolean](false)
+      alreadyFailed <- Ref.of[IO, Boolean](false)
+      failingOnce    = Stream.eval(alreadyFailed.get).flatMap {
```
imsdu, line 64: To avoid the status to be flipping between `Running` and `Failed`

olivergrabinski (approved):

PR #4620, opened 2024-01-03
Save scope initialization errors 

olivergrabinski:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ProjectsImpl.scala
```diff
@@ -33,15 +34,29 @@ final class ProjectsImpl private (
...
+                          case Left(e)  =>
+                            errorStore
+                              .save(init.entityType, resource.value.ref, e)
+                              .onError(e =>
+                                logger.error(e)(
+                                  s"Failed to save error for '${init.entityType}' initialization step on project '${resource.value.ref}'"
+                                )
+                              )
+                          case Right(_) => IO.unit
+                        }
```
shinyhappydan, line 55: There will be something along the lines of `onError { case e: ScopeInitializationFailed => ???}` you could use

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ProjectsImpl.scala
```diff
@@ -33,15 +34,29 @@ final class ProjectsImpl private (
...
+                        .onProjectCreation(resource.value, caller)
+                        .attemptNarrow[ScopeInitializationFailed]
+                        .flatMap {
+                          case Left(e)  =>
+                            errorStore
+                              .save(init.entityType, resource.value.ref, e)
+                              .onError(e =>
+                                logger.error(e)(
+                                  s"Failed to save error for '${init.entityType}' initialization step on project '${resource.value.ref}'"
+                                )
```
shinyhappydan, line 52: The store should probably be doing this?

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ScopeInitializationErrorStoreSuite.scala
```diff
@@ -0,0 +1,68 @@
...
+    1,
+    entityType.value,
+    project.organization.value,
+    project.project.value,
+    scopeInitError.reason,
+    Instant.EPOCH
+  )
+
+  test("Inserting an error should succeed") {
+    assertStoreIsEmpty >>
```
shinyhappydan, line 41: Would it ever not be empty?

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ScopeInitializationErrorStoreSuite.scala
```diff
@@ -0,0 +1,68 @@
...
+  case class ScopeInitErrorRow(
+      ordering: Int,
+      entityType: String,
+      org: String,
+      project: String,
+      message: String,
+      instant: Instant
+  )
+
+  implicit class ScopeInitErrorStoreOps(store: ScopeInitializationErrorStore) {
```
shinyhappydan, line 55: I'm not sure what the implicit class is adding here, it's actually quite confusing that the store appears to have this method on it

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ProjectsImplSpec.scala
```diff
@@ -265,5 +265,27 @@ class ProjectsImplSpec extends CatsEffectSpec with DoobieScalaTestFixture with C
...
+      }
+
+      val inits    = Set(failingScopeInit, successfulScopeInit)
+      val projects = ProjectsImpl(fetchOrg, validateDeletion, inits, defaultApiMappings, config, xas, clock)
+
+      val createProject                    = projects.create(project, payload)
+      val assertSuccessfulInitStepExecuted =
+        successfulScopeInit.createdProjects.get.map(p => p shouldEqual Set(project))
+
+      createProject >> assertSuccessfulInitStepExecuted
```
shinyhappydan, line 288: Should it assert that a log was recorded?

shinyhappydan (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ProjectsImpl.scala
olivergrabinski: You're right, I already had the same issue at some point; I keep forgetting about `recover`

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ProjectsImpl.scala
olivergrabinski: Yes I was going to move that later

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ProjectsImplSpec.scala
olivergrabinski: Yes there should be a check for that. The reason I did not add it is because I didn't want to have to deal with the errorStore directly. In my mind the next step is the ability to check the health/errors for a given project which can then be used to do that check

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ScopeInitializationErrorStoreSuite.scala
olivergrabinski: If I were to add another test now, it would start with a non empty store. Should it be more like a `resetStore` method, here? Not sure

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ScopeInitializationErrorStoreSuite.scala
olivergrabinski: My thinking here was that I didn't want to add methods to the actual implementation that were only used in tests. However I agree it's not very successful, and it should be gone soon, if not in this PR then in the next one

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ScopeInitializationErrorStoreSuite.scala
shinyhappydan: I think the tests having a clean slate isn't really something you need to check at the beginning of the test, it's more for a beforeAll

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ScopeInitializationErrorStoreSuite.scala
shinyhappydan: I see what you mean, but I would suggest that a normal method would be clearer than the implicit class

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ProjectsImplSpec.scala
shinyhappydan: I think it's legit to deal with the error store (mocking it probably makes the most sense for this test), but your suggested next step could work too

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ScopeInitializationErrorStoreSuite.scala
imsdu: As those information should be part of the new endpoint allowing to check that projects are correctly initialized.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ProjectsImpl.scala
```diff
@@ -33,15 +34,20 @@ final class ProjectsImpl private (
...
+  )(implicit caller: Subject): IO[ProjectResource] =
     for {
       resource <- eval(CreateProject(ref, fields, caller)).span("createProject")
       _        <- scopeInitializations
-                    .parUnorderedTraverse(_.onProjectCreation(resource.value, caller))
+                    .parUnorderedTraverse { init =>
+                      init
+                        .onProjectCreation(resource.value, caller)
+                        .recoverWith { case e: ScopeInitializationFailed =>
+                          errorStore.save(init.entityType, resource.value.ref, e)
```
imsdu, line 45: The error should still be returned to the user

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/ProjectsImpl.scala
```diff
@@ -33,15 +34,20 @@ final class ProjectsImpl private (
   override def create(
       ref: ProjectRef,
       fields: ProjectFields
-  )(implicit caller: Subject): IO[ProjectResource] = {
+  )(implicit caller: Subject): IO[ProjectResource] =
     for {
       resource <- eval(CreateProject(ref, fields, caller)).span("createProject")
       _        <- scopeInitializations
-                    .parUnorderedTraverse(_.onProjectCreation(resource.value, caller))
+                    .parUnorderedTraverse { init =>
```
imsdu, line 41: A bit like `IndexingAction`, it could be interesting to refactor this by introducing some kind of aggregate there which would take care of the init steps and saving errors

PR #4621, opened 2024-01-04
Add adt for view type restrictions

### LINKED ISSUE #4374: Introduce ADT for `resourceType` and `resourceSchema` in view model
**Motivation**

In the view models, there is an assumption that and empty set of types or schemas means no restriction. This leads to unclear code that has to deal with the emptyset case in several locations.

* CompositeView
    * The sources and projections have `resourceTypes` and `resourceSchemas` as `Set[Iri]`
* ElasticSearchView
    * The `FilterByType` and `FilterBySchema` pipes rely on `Set[Iri]`
* BlazegraphView
    * Indexing view has `resourceTypes` and `resourceSchemas` as `Set[Iri]`

The idea would be to deal with this using an ADT with a specific case when there is no restriction (ADT such as `NoRestriction` and `RestrictedTo(types: Set[Iri])`)

dantb:
Fixes #4374


delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/model/BlazegraphView.scala
```diff
@@ -112,8 +112,8 @@ object BlazegraphView {
       description: Option[String],
       project: ProjectRef,
       uuid: UUID,
-      resourceSchemas: Set[Iri],
-      resourceTypes: Set[Iri],
+      resourceSchemas: ValidViewTypes,
```
imsdu, line 115: For `resourceSchemas`, the `Iri` defines the identifier of the schema 
For `resourceTypes`, the `Iri` defines the types of the resources eligible to index.
So we need a better name

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/model/ValidViewTypes.scala
```diff
@@ -0,0 +1,30 @@
+package ch.epfl.bluebrain.nexus.delta.sourcing.model
+
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.decoder.JsonLdDecoder
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.ValidViewTypes.RestrictedTo
+import io.circe.syntax.EncoderOps
+import io.circe.{Decoder, Encoder, Json}
+
+sealed trait ValidViewTypes {
```
imsdu, line 9: To rename
Also as we don't really care if it is an Iri there so we could have a type parameter ?

dantb: Since it's just incidental that they're iris? Is there a specific example where we could reuse it for a different type? Otherwise that might be over generifying. Considered potentially having different ADTs for schemas / types to make the difference explicit

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/model/ViewRestriction.scala
```diff
@@ -0,0 +1,30 @@
+package ch.epfl.bluebrain.nexus.delta.sourcing.model
+
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.decoder.JsonLdDecoder
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.ViewRestriction.RestrictedTo
+import io.circe.syntax.EncoderOps
+import io.circe.{Decoder, Encoder, Json}
+
+sealed trait ViewRestriction {
```
imsdu, line 9: Maybe `IriFilter` to have a naming closer to `SelectFilter` ?

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/model/ViewRestriction.scala
```diff
@@ -0,0 +1,30 @@
...
+  def asRestrictedTo: Option[RestrictedTo] = this match {
+    case ViewRestriction.None => None
+    case r: RestrictedTo      => Some(r)
+  }
+}
+
+object ViewRestriction {
+  final case object None                                  extends ViewRestriction
+  sealed abstract case class RestrictedTo(iris: Set[Iri]) extends ViewRestriction
```
imsdu, line 18: In other places, we tend to use the term `Include`

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/model/ValidViewTypes.scala
imsdu: You are right we can keep it like this for now

imsdu (approved):

PR #4622, opened 2024-01-04
Configure docker compose to run Delta in clustered node

imsdu:
Using env files and profiles from docker compose to run Delta in standalone or clustered nodes

The docker command looks like this:
```
# Standalone
docker compose -f tests/docker/docker-compose.yml up -d

# Clustered mode
docker compose -f tests/docker/docker-compose.yml --env-file tests/docker/.env.cluster --profile cluster up -d
```


olivergrabinski (approved):
Out of curiosity have you tried using the
```
deploy:
  replicas: 3
```
option in the docker compose file? I guess it would almost work though I'm not sure if we can specify a different `entrypoint` per replica and make them depend on the first one.

imsdu:
> Out of curiosity have you tried using the
> 
> ```
> deploy:
>   replicas: 3
> ```
> 
> option in the docker compose file? I guess it would almost work though I'm not sure if we can specify a different `entrypoint` per replica and make them depend on the first one.
People tried this before, I haven't tried again as it would surely mean more dev.

PR #4625, opened 2024-01-05
Add keywords to files during creation

### LINKED ISSUE #4512: Add the ability to add custom metadata to files during creation
**Motivation**
Users can't define custom metadata (like author, license, ...) on files so they either have to define it in an associated resource or they encode it in the filename which is not a great idea.

The idea is to allow them to add this metadata to be able to index them in order to make it easier to find them.
This metadata would look like:

```
{
  "name": "My File",
  "description": "Some description that may be useful in full-text search",
  "keywords": {
    "author": "Me",
    "license": "XXX",
    "emodel": "M1",
    "brainRegion": "Isocortex",
    "species": "Mouse"
  }
}
```

where keywords is a list of key/values

**Acceptance criteria**
1. When creating a file, metadata can optionally be provided
2. The key should follow the `Label` restriction
3. Values are only strings
4. This metadata is correctly indexed in Blazegraph/Elasticsearch using Delta context for files

**Not in scope**
* The ability to push this information when updating/link a file will be addressed in later tickets
* Keywords can't be patched, all of them must be provided every time

shinyhappydan:
Fixes #4512 

- [x] Check full text search works
- [ ] Name, description fields
- [x] Parsing metadata should return an error
- [x] Metadata fields should be moved?


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ResourcesSearchParams.scala
```diff
@@ -62,6 +65,17 @@ final case class ResourcesSearchParams(
 
 object ResourcesSearchParams {
 
+  case class FileUserMetadata(keywords: Map[String, String])
```
imsdu, line 68: `name` and `description` are missing.
The key should be a `Label` to restrain the possible values

delta/plugins/elasticsearch/src/main/resources/defaults/default-mapping.json
```diff
@@ -182,6 +182,9 @@
         }
       }
     },
+    "_keywords": {
```
imsdu, line 185: keywords.
Also:
* You will need a multi-field here to enable full text search properly
* `Flattened` could be an option here, we have to be careful that when searching, the ES query does not fail because a keyword value does not exist but return an empty value

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/QueryBuilder.scala
```diff
@@ -69,6 +69,10 @@ final case class QueryBuilder private[client] (private val query: JsonObject) {
     */
   def withFilters(params: ResourcesSearchParams)(implicit baseUri: BaseUri): QueryBuilder = {
     val (includeTypes, excludeTypes) = params.types.partition(_.include)
+    val value1 = params.fileUserMetadata.toList.flatMap(_.keywords).map { case (key, value) =>
```
imsdu, line 72: Can we address the change in the listing operation in a later PR and just make sure that data is properly indexed and full text search on keywords is working ?

shinyhappydan: I can't have a working test without the listing so I think it's needed. We can not do name/description if we want to slice it. Full text search could also be cut given that it's more advanced functionality and isn't mentioned in the ticket

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FormDataExtractor.scala
```diff
@@ -122,12 +127,18 @@ object FormDataExtractor {
...
-      private def extractFile(part: FormData.BodyPart): Future[Option[(FileDescription, BodyPartEntity)]] = part match {
+      private def extractFile(part: FormData.BodyPart): Future[Option[FileInformation]] = part match {
         case part if part.name == fieldName =>
-          val filename    = part.filename.getOrElse("file")
+          val filename = part.filename.getOrElse("file")
+
           val contentType = detectContentType(filename, part.entity.contentType)
+          val metadata    = part.dispositionParams.get("metadata").flatMap { metadata =>
+            println("metadata: " + metadata)
+            parser.parse(metadata).toOption.flatMap(_.as[FileUserMetadata].toOption)
```
imsdu, line 137: We should return an error here

delta/plugins/elasticsearch/src/main/resources/contexts/indexing-metadata.json
```diff
@@ -15,6 +15,8 @@
     },
     "_algorithm": "https://bluebrain.github.io/nexus/vocabulary/algorithm",
     "_attributes": "https://bluebrain.github.io/nexus/vocabulary/attributes",
+    "@vocab": "https://bluebrain.github.io/nexus/vocabulary/",
+    "_keywords": "https://bluebrain.github.io/nexus/vocabulary/keywords",
```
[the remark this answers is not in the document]
shinyhappydan: In what way does DefaultLabelPredicates need to be updated? It’s not clear what it is doing

`Pipe implementation that transforms the resource graph keeping only the specific predicates defined in DefaultLabelPredicates#defaultLabelPredicates`

If that were true then wouldn’t we already have the keyword predicate removed?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FormDataExtractor.scala
```diff
@@ -122,12 +127,18 @@ object FormDataExtractor {
...
+          }
+          println("metadata output: " + metadata)
           FileDescription(filename, contentType).unsafeToFuture().map { desc =>
-            Some(desc -> part.entity)
+            Some(FileInformation(metadata, desc, part.entity))
```
imsdu, line 141: We could just put this new field in `FileDescription` ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileCommand.scala
```diff
@@ -64,6 +65,7 @@ object FileCommand {
       storage: ResourceRef.Revision,
       storageType: StorageType,
       attributes: FileAttributes,
+      userMetadata: Option[FileUserMetadata],
```
imsdu, line 68: We could add more fields to `FileAttributes` ?
It would allow to get rid of the metadata confusion.
In all cases, we should avoid to import something from the elasticsearch plugin here

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/client/QueryBuilder.scala
imsdu: The data can be indexed and returned by the listing operation without having an extra param to filter for the moment.

shinyhappydan: I would prefer to implement a full slice of functionality, plus: it's already done

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FormDataExtractor.scala
shinyhappydan: My IDE says that nothing uses FileDescription 🤔

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileCommand.scala
shinyhappydan: I can't tell whether that's an improvement. All of those things are calculated, where this is user-supplied information. It feels like it might be better to keep them separate

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileEvent.scala
```diff
@@ -337,19 +342,21 @@ object FileEvent {
...
             case "subject"                             => nxv.eventSubject.prefix
             case field if metadataKeys.contains(field) => s"_$field"
             case other                                 => other
           })
-        implicit val subjectEncoder: Encoder[Subject]       = IriEncoder.jsonEncoder[Subject]
-        implicit val projectRefEncoder: Encoder[ProjectRef] = IriEncoder.jsonEncoder[ProjectRef]
+        implicit val subjectEncoder: Encoder[Subject]                    = IriEncoder.jsonEncoder[Subject]
+        implicit val projectRefEncoder: Encoder[ProjectRef]              = IriEncoder.jsonEncoder[ProjectRef]
+        implicit val userMetadataCodec: Codec.AsObject[FileUserMetadata] =
+          deriveConfiguredCodec[FileUserMetadata]
```
dantb, line 359: I wonder if this should be treated as a leaf node and not have configured codecs? Since it really is just a `Map[String, String]` wrapper rather than a KG thing (which I assume the custom circe config is intended for)

dantb: That sentence wasn't clear at all - just meant we don't want to accidentally map certain custom keys / values based on config like `transformMemberNames`

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ResourcesSearchParams.scala
```diff
@@ -49,6 +53,7 @@ final case class ResourcesSearchParams(
     updatedAt: TimeRange = TimeRange.Anytime,
     types: List[Type] = List.empty,
     typeOperator: TypeOperator = TypeOperator.Or,
+    fileUserMetadata: Option[FileUserMetadata] = None,
```
imsdu, line 56: `keywords` ?
In the end, it could also be extended to regular resources if users provide them

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/FilesSpec.scala
```diff
@@ -54,6 +69,37 @@ class FilesSpec extends BaseIntegrationSpec {
...
+        s"/files/$org/$project/$id",
+        "file content",
+        ContentTypes.`text/plain(UTF-8)`,
+        s"$id.json",
+        Writer,
+        Json.obj("keywords" := Json.obj("brainRegion" := brainRegion))
+      )
+      .map { case (json, response) =>
+        response.status shouldEqual StatusCodes.Created
+        json.hcursor.downField("@id").as[String].getOrElse(fail("Could not extract @id from response"))
```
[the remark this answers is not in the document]
shinyhappydan: I think it's planned that we remove the optics library, otherwise I would agree

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FormDataExtractor.scala
imsdu: `ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model.FileDescription`
Not sure anymore what I was thinking back then though

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileCommand.scala
imsdu: They can stay separated but it would be nice to have a naming involving the metadata term to avoid confusion

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ResourcesSearchParams.scala
shinyhappydan: Well it might be more than just keywords? I know that's the only field right now but soon we will have others

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileSerializationSuite.scala
```diff
@@ -70,18 +72,21 @@ class FileSerializationSuite extends SerializationSuite with StorageFixtures {
 
   private val filesMapping = List(
     (
+      "FileCreated",
```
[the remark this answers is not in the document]
shinyhappydan: I changed it because the test names were just wrong, but you're right it is a bit easier to read now

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ResourcesSearchParams.scala
imsdu: Name and description will only part of full-text search

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileState.scala
```diff
@@ -1,5 +1,6 @@
 package ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model
 
+import ch.epfl.bluebrain.nexus.delta.plugins.elasticsearch.model.ResourcesSearchParams.FileUserMetadata
```
imsdu, line 3: It should not use an import from the ES plugin

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/model/ResourcesSearchParams.scala
```diff
@@ -62,6 +67,20 @@ final case class ResourcesSearchParams(
 
 object ResourcesSearchParams {
 
+  case class FileUserMetadata(keywords: Map[String, String])
```
imsdu, line 70: The key should be a `Label`

imsdu (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileState.scala
shinyhappydan: Interesting. You are right, I see two possibilities:
- Put everything in FileAttributes
- Make a class in non-plugin code to represent the data

I think I'm leaning to using FileAttributes, what do you think @dantb @imsdu ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileEvent.scala
dantb: Yeah, if we're testing mapped fields as the metadata fields. Just a regular Codec would work I think, without the configured part e.g. `deriveCodec`

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileState.scala
dantb: `FileUserMetadata` feels like a file thingy in the way it's supported by the API, so that seems reasonable to me

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/batch/BatchFilesSuite.scala
```diff
@@ -53,6 +54,22 @@ class BatchFilesSuite extends NexusSuite with StorageFixtures with Generators wi
...
+    implicit val c: Caller     = Caller(genUser(), Set())
+    val batchFiles: BatchFiles = mkBatchFiles(events, destProj, destFileUUId, fetchFileStorage, batchCopy)
+
+    val sourceProject = genProjectRef()
+
+    batchFiles.copyFiles(
+      CopyFileSource(sourceProject, NonEmptyList.of(genFileId(sourceProject), genFileId(sourceProject))),
+      destination
+    )
+  }
```
dantb, line 71: Can delete now?

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/BatchCopySpec.scala
```diff
@@ -105,24 +108,45 @@ class BatchCopySpec extends BaseIntegrationSpec {
...
+  ): IO[Assertion] = {
+    ids
+      .zip(sourceFiles)
+      .parTraverse { case (id, FileInput(_, _, _, _, keywords)) =>
+        deltaClient.get[Json](s"/files/$destProjRef/${UrlUtils.encode(id)}", Coyote) { (json, response) =>
+          response.status shouldEqual StatusCodes.OK
+          json.hcursor.downField("keywords").as[Map[String, String]].toOption match {
+            case Some(value) => value shouldEqual keywords
+            case None        => fail("keywords missing")
+          }
```
dantb, line 142: could do
```suggestion
          json.hcursor.get[Map[String, String]]("keywords").value shouldEqual keywords
```

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/generators/FileGen.scala
```diff
@@ -48,9 +48,11 @@ trait FileGen { self: Generators with FileFixtures =>
 
   def genFileIdWithTag(projRef: ProjectRef): FileId = FileId(genString(), UserTag.unsafe(genString()), projRef)
 
-  def genAttributes(): NonEmptyList[FileAttributes] = {
+  def genAttributesAndMetadata(): NonEmptyList[(FileAttributes, Option[FileUserMetadata])] = {
```
dantb, line 51: I should've done these generators in a less half-arsed way or got scalacheck working 😬 does the job though

dantb (approved):

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/BatchCopySpec.scala
shinyhappydan: Yeah the failure message was not good when I had that, that's why I used this match

dantb: ah yeah I've found that with these tests too, I quite like just `actual shouldEqual Right(expected)`

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/ClasspathResourceLoader.scala
```diff
@@ -124,7 +121,7 @@ class ClasspathResourceLoader private (classLoader: ClassLoader) {
 }
 
 object ClasspathResourceLoader {
-  private[utils] val handleBars = new Handlebars().`with`(EscapingStrategy.NOOP)
+  private val handlebarsExpander = new HandlebarsExpander
```
imsdu, line 124: What was the issue here ?

delta/plugins/elasticsearch/src/main/resources/contexts/elasticsearch-indexing.json
```diff
@@ -4,7 +4,9 @@
     "prefLabel": "http://www.w3.org/2004/02/skos/core#prefLabel",
     "name": "http://schema.org/name",
     "label": "http://www.w3.org/2000/01/rdf-schema#label",
-    "description": "http://schema.org/description"
+    "description": "http://schema.org/description",
+    "keywords": "https://schema.org/keywords",
```
imsdu, line 8: Switch to the nexus vocabulary ?

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/ClasspathResourceLoader.scala
shinyhappydan: I had previously added something else to the handlebars instance which I then removed, although I think this interface is still better

delta/plugins/elasticsearch/src/main/resources/defaults/default-mapping.json
```diff
@@ -182,6 +182,10 @@
         }
       }
     },
+    "keywords": {
+      "type": "object",
```
imsdu, line 186: Did you try with the flattened type ?
We have to be careful with the dynamic mapping:
https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html
We don't want someone to push a value that can be parsed as a date/a number and have this weird case where the first occurence set it as such and have regular strings failing after
One alternative to flatten can be mapping templates and:
https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsDirectives.scala
```diff
@@ -79,6 +94,7 @@ trait ElasticSearchViewsDirectives extends UriDirectives {
               updatedAt,
               types,
               typeOperator,
+              fileUserMetadata,
```
imsdu, line 97: keywords ?

delta/plugins/elasticsearch/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/routes/ElasticSearchViewsDirectivesSpec.scala
```diff
@@ -48,6 +48,7 @@ class ElasticSearchViewsDirectivesSpec extends RouteHelpers with CirceMarshallin
   implicit val paramTypeOperatorCodec: Codec[ResourcesSearchParams.TypeOperator] =
     deriveConfiguredCodec[ResourcesSearchParams.TypeOperator]
   implicit val paramsCodec: Codec[ResourcesSearchParams]                         = deriveConfiguredCodec[ResourcesSearchParams]
+//  implicit val metadataCodec: Codec[FileUserMetadata]                            = deriveConfiguredCodec[FileUserMetadata]
```
imsdu, line 51: To remove

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FormDataExtractor.scala
```diff
@@ -118,23 +127,38 @@ object FormDataExtractor {
         .adaptError {
           case _: EntityStreamSizeException =>
             FileTooLarge(maxFileSize, storageAvailableSpace)
-          case th                           =>
+          case NotARejection(th)            =>
```
imsdu, line 130: A rejection can be risen here ?
Not sure if we should just drop this one
Especially to map it to a `MalformedRequestContentRejection`, this may fail for other reasons that do not depend on the client

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FormDataExtractor.scala
```diff
@@ -118,23 +127,38 @@ object FormDataExtractor {
...
-      private def extractFile(part: FormData.BodyPart): Future[Option[(FileDescription, BodyPartEntity)]] = part match {
-        case part if part.name == fieldName =>
+      private def extractFile(part: FormData.BodyPart): Future[Option[UploadedFileInformation]] = part match {
+        case part if part.name == FileFieldName =>
           val filename    = part.filename.getOrElse("file")
           val contentType = detectContentType(filename, part.entity.contentType)
-          FileDescription(filename, contentType).unsafeToFuture().map { desc =>
-            Some(desc -> part.entity)
+
+          val result = for {
```
imsdu, line 139: You can just map here

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileMetadata.scala
```diff
@@ -0,0 +1,11 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model
+
+import ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model.FileAttributes.FileAttributesOrigin
+
+case class FileMetadata(bytes: Long, digest: Digest, origin: FileAttributesOrigin)
```
imsdu, line 5: Dead code ?

delta/plugins/storage/src/main/resources/contexts/files.json
```diff
@@ -18,6 +19,7 @@
     },
     "_algorithm": "https://bluebrain.github.io/nexus/vocabulary/algorithm",
     "_attributes": "https://bluebrain.github.io/nexus/vocabulary/attributes",
+    "_keywords": "https://schema.org/keywords",
```
imsdu, line 22: Switch to the nexus vocabulary

delta/rdf/src/main/scala/ch/epfl/bluebrain/nexus/delta/rdf/Vocabulary.scala
```diff
@@ -99,6 +99,7 @@ object Vocabulary {
     val description       = schema + "description"
     val name              = schema + "name"
     val unitText          = schema + "unitText"
+    val keywords          = iri"https://schema.org/keywords"
```
imsdu, line 102: Same to move to the nexus vocab

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FormDataExtractorSpec.scala
```diff
@@ -21,57 +20,93 @@ class FormDataExtractorSpec
...
-
     val content = "file content"
     val iri     = iri"http://localhost/file"
 
     val customMediaType   = MediaType.parse("application/custom").rightValue
     val customContentType = ContentType(customMediaType, () => HttpCharsets.`UTF-8`)
     val mediaTypeDetector = MediaTypeDetectorConfig(Map("custom" -> customMediaType))
     val extractor         = FormDataExtractor(mediaTypeDetector)
+    val KeyThatIsTooLong  =
+      "this-key-is-too-long-to-be-a-label-lalalalalalalaalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalaalalalla"
```
imsdu, line 31: It could have been generated :)

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileMetadata.scala
shinyhappydan: It's used when copying

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileAttributes.scala
```diff
@@ -40,20 +39,48 @@ final case class FileAttributes(
...
     mediaType: Option[ContentType],
+    // TODO: Remove default after ??? migration
+    keywords: Map[Label, String] = Map.empty,
     bytes: Long,
     digest: Digest,
     origin: FileAttributesOrigin
-)
+) extends LimitedFileAttributes
+
+trait LimitedFileAttributes {
```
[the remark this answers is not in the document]
shinyhappydan: It's used when copying

delta/kernel/src/main/scala/ch/epfl/bluebrain/nexus/delta/kernel/utils/ClasspathResourceLoader.scala
imsdu: As the handlebars thing is only used here, maybe the `HandleBarsExpander` can be here too ?
It does not really matters though

delta/plugins/elasticsearch/src/main/resources/contexts/elasticsearch-indexing.json
```diff
@@ -4,7 +4,8 @@
     "prefLabel": "http://www.w3.org/2004/02/skos/core#prefLabel",
     "name": "http://schema.org/name",
     "label": "http://www.w3.org/2000/01/rdf-schema#label",
-    "description": "http://schema.org/description"
+    "description": "http://schema.org/description",
+    "@vocab": "https://bluebrain.github.io/nexus/keywords/"
```
imsdu, line 8: This one should go with indexing-metadata too

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileAttributes.scala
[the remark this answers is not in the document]
imsdu: We can't do without it ?
Or have something that would be in the copy operation scope ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileStorageMetadata.scala
```diff
@@ -0,0 +1,16 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model
+
+import akka.http.scaladsl.model.Uri
+import akka.http.scaladsl.model.Uri.Path
+import ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model.FileAttributes.FileAttributesOrigin
+
+import java.util.UUID
+
+case class FileStorageMetadata(
```
imsdu, line 9: `FileStorage` ?

shinyhappydan: It's not a file storage though, it's metadata regarding where the file is stored. FileStorage sounds too much like S3Storage etc

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileAttributes.scala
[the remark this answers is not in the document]
shinyhappydan: It's better with it. We can remove it when we stop storing the uuid

delta/plugins/elasticsearch/src/main/resources/contexts/elasticsearch-indexing.json
imsdu: It can now be deleted from here

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileRejection.scala
```diff
@@ -153,6 +153,13 @@ object FileRejection {
   final case class InvalidMultipartFieldName(id: Iri)
       extends FileRejection(s"File '$id' payload a Multipart/Form-Data without a 'file' part.")
 
+  /**
+    * Rejection returned when attempting to create/update a file with a Multipart/Form-Data payload that has user
+    * metadata which cannot be parsed
+    */
+  final case class InvalidUserMetadata(err: String)
```
imsdu, line 160: InvalidFileDescription ?
As it has been renamed ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileStorageMetadata.scala
imsdu: I just would like to drop/replace the metadata suffix where we can

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileRejection.scala
shinyhappydan: I'll call it `InvalidKeywords` for now, but might re-use it for the name/description later so it might get renamed back

PR #4629, opened 2024-01-08
Replace Envelope with Elem.SuccessElem

### LINKED ISSUE #3745: Remove envelope for elem
**Motivation**
There is no need for `Envelope` as its usages are also covered by `Elem`.
So `Envelope` should go

**Acceptance criteria**
`Envelope` disappeards from the codebase

dantb:
Fixes #3745


delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
```diff
@@ -339,8 +340,8 @@ final class BlazegraphViews(
       IO.pure(toIndexViewDef(envelope))
     }
 
-  private def toIndexViewDef(envelope: Envelope[BlazegraphViewState]) =
-    envelope.toElem { v => Some(v.project) }.traverse { v =>
+  private def toIndexViewDef(elem: Elem.SuccessElem[BlazegraphViewState]) =
+    elem.withProject(elem.value.project).traverse { v =>
```
dantb, line 344: Doesn't feel great that this is optional - most of the concrete types for which we're querying elems have a project and they are mandatory in the scoped tables. I guess it's because `Envelope` was used for both global and scoped?

shinyhappydan: It's confusing how the elem wouldn't already have the project set since it's coming from within it..?

shinyhappydan (commented):
What was a stream of Envelopes, and what is a stream of SuccessElems?

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/BlazegraphViews.scala
imsdu: You should be able to populate this field from the original SQL query.
Yes, it was because `Envelope` and `Elem` were designed to work for both global and scoped entities.
Not sure if there is still a use case for global entities though now that we got rid of the sse events for them.

dantb:
Looking into why the project was optional led me to try to understand the usage of elem / envelope before... My feeling is it's used in places it shouldn't be which leads to incidental complexity like streams of `Elem` where nothing on the `Elem` is ever used. The usages I've seen:
1. Streaming from global tables. Now we use `GlobalStateValue` purely for the `Read` instance when streaming from `global_states`. Only `value` is needed.
2. Streaming from scoped tables. This seems fine. However for the projections, the coordinators were using a stream of `Elem` rather than `SuccessElem` for views. This meant the type suggested that there could be failing / dropped elements, but there would never be. I've changed that now.
3. Streaming from `projection_restarts`. No project here, I haven't modified this.
4. Streaming from `composite_restarts`. There is a project here.
5. Streaming graph resources - haven't yet gone down this rabbit hole.

3 and 5 mean the project remains optional. I don't understand enough to comment on what should be used and where, but it's still pretty confusing in its current state. So I still can't really answer your question @shinyhappydan 😂  maybe @imsdu can!

Also deleted a load of unused streaming methods from the event stores and logs. Probably enough for now!

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/state/GlobalStateStore.scala
```diff
@@ -122,21 +110,19 @@ object GlobalStateStore {
         .option
         .transact(xas.read)
 
-    private def states(offset: Offset, strategy: RefreshStrategy): EnvelopeStream[S] =
-      StreamingQuery[Envelope[S]](
+    private def states(offset: Offset, strategy: RefreshStrategy): Stream[IO, S] =
+      StreamingQuery[GlobalStateValue[S]](
```
imsdu, line 114: Maybe you could use a tuple (S, Offset) here ?
Or make `GlobalStateValue`private ?

imsdu (approved):

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/state/GlobalStateStore.scala
dantb: Thanks, less code!

PR #4635, opened 2024-01-09
Add a project healing endpoint

olivergrabinski:
Closes #4580

Screenshot 2024-01-10 at 16 41 34

PR #4640, opened 2024-01-11
Clarify method in `ScopeInitializer`

olivergrabinski:
Small fix related to #4635


dantb (approved):
That the first error is taken may also be worth clarifying

shinyhappydan (approved):
That's really nice, thanks!

PR #4658, opened 2024-01-15
Fix website deploy with java 17

imsdu:


project/plugins.sbt
```diff
@@ -11,9 +11,11 @@ addSbtPlugin("com.lightbend.sbt" % "sbt-javaagent"       % "0.1.6")
...
 addSbtPlugin("com.eed3si9n"      % "sbt-buildinfo"       % "0.10.0")
 
-addSbtPlugin("com.typesafe.sbt"      % "sbt-site"                   % "1.4.1")
-addSbtPlugin("com.lightbend.paradox" % "sbt-paradox"                % "0.9.2")
-addSbtPlugin("io.github.jonas"       % "sbt-paradox-material-theme" % "0.5.1")
+addSbtPlugin(
+  ("com.typesafe.sbt"                % "sbt-site"                   % "1.4.1").exclude("org.scala-lang.modules", "scala-xml_2.12")
+)
+addSbtPlugin("com.lightbend.paradox" % "sbt-paradox"                % "0.10.3")
+addSbtPlugin("io.github.jonas"       % "sbt-paradox-material-theme" % "0.6.0")
```
olivergrabinski, line 18: Wasn't there some kind of issue with this version?

imsdu: Hum, I set the same one as 1.8 but there is indeed a newer one.
I only remember the problem with 0.9.x but I will set the newer version

shinyhappydan (approved):

project/plugins.sbt
olivergrabinski: I actually meant the question for `sbt-paradox-material-theme`, I thought it was breaking something on the website

imsdu: This one has not changed in a loooong time

olivergrabinski: Isn't it being updated from `0.5.1` to `0.6.0` in this PR?

imsdu: Sorry, I meant the plugin itself is not maintained anymore, I must have misread your question.
Those are both the paradox versions that we use in the main branch

PR #4670, opened 2024-01-18
Log database config at startup

imsdu:


dantb (approved):
LGTM - I'd be keen on a config endpoint, particularly with the double prod situation. I've used [ciris](https://cir.is/docs/quick-example) before for config loading and it ensures secrets are obfuscated (ideally we'd use it for all custom config, but that might be a lot of work)

imsdu:
> LGTM - I'd be keen on a config endpoint, particularly with the double prod situation. I've used [ciris](https://cir.is/docs/quick-example) before for config loading and it ensures secrets are obfuscated (ideally we'd use it for all custom config, but that might be a lot of work)

Yep, we can introduce this at some point, note that we use in the configs a `Secret` class where we override toString and have custom encoders.

PR #4671, opened 2024-01-18
Support passing JWK keys as properties to storage service

dantb:
Sorry, it's horrendous. This allows us to inject config to the storage service process via only properties - the issue was the `x5c` field is a list within a JSON object within a list, but the value gets parsed as a string. For the other fields you can do `"-Dapp.authorization.keys.0.kty=RSA"`. There might be a better way?

I tested overriding the docker-compose properties and this now works when setting it to
```
"-Dapp.authorization.keys.0.x5c=[\"blahblah\"]"
```

Hopefully it can be reverted once we have a better config solution in AWS.


olivergrabinski (approved):

dantb:
I'll leave this open for now @olivergrabinski @imsdu , maybe we'll never need it

imsdu:
Making it a draft to make sure we don't accidentally merge it :)

PR #4672, opened 2024-01-18
Add StressSpec

olivergrabinski:


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/StressSpec.scala
```diff
@@ -0,0 +1,57 @@
+package ch.epfl.bluebrain.nexus.tests.kg
+
+import akka.http.scaladsl.model.StatusCodes
+import cats.implicits._
+import ch.epfl.bluebrain.nexus.tests.BaseIntegrationSpec
+import ch.epfl.bluebrain.nexus.tests.Identity.listings.Bob
+import ch.epfl.bluebrain.nexus.tests.iam.types.Permission.Organizations
+import io.circe.Json
+
+class StressSpec extends BaseIntegrationSpec {
```
imsdu, line 10: How much time does it take to complete ?
It would need to be improved a bit but it can be done later

imsdu (approved):

PR #4673, opened 2024-01-18
Deprecate resolver tagging

### LINKED ISSUE #4638: Remove the ability to tag resolvers
**Motivation**
Tagging is an operation which has been generalized for the different types of resources.
However because it takes time to align the behaviour between the different types of resources, there are discrepancies on how tagging works between them.
For resolvers, the feature is not used so we can just drop it.

**Acceptance criteria**
1. The endpoint for tagging is removed for resolvers
2. The associated command is removed from the state machine
3. The documentation is updated
4. The event should remain for the moment.

olivergrabinski:
Closes #4638


delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resolvers/ResolversImplSpec.scala
```diff
@@ -602,39 +522,19 @@ class ResolversImplSpec extends CatsEffectSpec with DoobieScalaTestFixture with
...
     }
 
     val inProjectExpected    = resolverResourceFor(
       nxv + "in-project",
       projectRef,
       updatedInProjectValue,
       sourceWithoutId(updatedInProjectValue),
-      tags = Tags(tag -> 1, tag2 -> 4),
-      rev = 5,
+      tags = Tags.empty,
```
imsdu, line 536: We can get rid of it if we remove tags from the state too.

olivergrabinski: Made the change

imsdu (approved):

PR #4675, opened 2024-01-18
Remove the ability to tag storages

dantb:


delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesStmSpec.scala
```diff
@@ -102,19 +90,6 @@ class StoragesStmSpec extends CatsEffectSpec with StorageFixtures {
           StorageUndeprecated(dId, project, DiskStorageType, 4, epoch, alice)
       }
 
-      "reject with IncorrectRev" in {
-        val state    = storageState(dId, project, diskVal)
-        val commands = List(
-          UpdateStorage(dId, project, diskFields, Json.obj(), 2, alice),
-          TagStorage(dId, project, 1, UserTag.unsafe("tag"), 2, alice),
```
imsdu, line 109: You should only remove this line, others are still relevant

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesStmSpec.scala
```diff
@@ -186,17 +161,6 @@ class StoragesStmSpec extends CatsEffectSpec with StorageFixtures {
           .rejectedWith[ResourceAlreadyExists]
       }
 
-      "reject with StorageNotFound" in {
-        val commands = List(
-          UpdateStorage(dId, project, diskFields, Json.obj(), 2, alice),
-          TagStorage(dId, project, 1, UserTag.unsafe("tag"), 2, alice),
```
imsdu, line 192: Only this line should be removed

imsdu (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
```diff
@@ -464,16 +435,6 @@ object Storages {
         } yield StorageUpdated(c.id, c.project, value, c.source, s.rev + 1, instant, c.subject)
     }
 
-    def tag(c: TagStorage) = state match {
```
imsdu, line 467: In `StorageState`, we should not have the tags anymore and the next method should only update the auditing fields
For the definition that lives in the same file, we can switch to `ScopedEntityDefinition.untagged`

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/routes/StoragesRoutes.scala
```diff
@@ -175,21 +174,6 @@ final class StoragesRoutes(
                               .attemptNarrow[StorageRejection]
                               .rejectOn[StorageNotFound]
                           )
-                        },
-                        // Tag a storage
-                        (post & parameter("rev".as[Int])) { rev =>
```
imsdu, line 180: The endpoint to fetch tags should also be changed to directly reject

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
```diff
@@ -183,34 +182,6 @@ final class Storages private (
...
-    * @param projectRef
-    *   the project where the storage belongs
-    * @param tag
-    *   the tag name
-    * @param tagRev
-    *   the tag revision
-    * @param rev
-    *   the current revision of the storage
-    */
-  def tag(
```
imsdu, line 200: It should also not be possible to fetch by tag, we may have to introduce another type in `IdSegmentRef` that excludes the `Tag` one

imsdu (changes requested):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/routes/StoragesRoutes.scala
```diff
@@ -175,21 +174,6 @@ final class StoragesRoutes(
                               .attemptNarrow[StorageRejection]
                               .rejectOn[StorageNotFound]
                           )
-                        },
-                        // Tag a storage
```
imsdu, line 179: You can also remove the endpoint to fetch the tags defined above

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/Storages.scala
```diff
@@ -368,10 +335,6 @@ object Storages {
       s.copy(rev = e.rev, value = e.value, source = e.source, updatedAt = e.instant, updatedBy = e.subject)
     }
 
-    def tagAdded(e: StorageTagAdded): Option[StorageState] = state.map { s =>
```
imsdu, line 371: I think we still need to compute a new state and update the rev, we may break some history otherwise.
For example, this should still work:
* StorageCreated
* StorageTagged
* StorageUpdated
We should still have a storage with revision 3

imsdu (approved):

PR #4684, opened 2024-01-22
Fix computing diff on large resources during updates

### LINKED ISSUE #3291: Add the project default base and vocab in resources that contain no context
**Is your feature request related to a problem? Please describe.**
Currently, Delta injects automatically an `@context` object (containing the project base and vocab configuration values) during resource creation/update when no context is provided.

The issue, however, is that Delta falsely assumes that a resource doesn't have a context when after expanding the provided json the result contains no entries. The current logic works for most cases, with the following exception:

When creating/updating a resource having the following source:
```
{
  "@type": "SomeType"
}
```
... when expanding the json document, the result contains an entry due to the fact that the use of `@type` is handled in a special way. The value of `@type` must be a valid IRI so the JSON-LD processor will use the configured `base` value instead of a `vocab` to generate a correct value for the `@type` field. This leads to having an entry in the resulting expanded json-ld document and in turn a default `@context` is not automatically added by Delta.
This is a fairly common case, where resources without an `@context` would be "incorrectly" created in the sense that the user expectation is different. Further more, for resources like:
```
{
  "@type": "SomeType",
  "name": "Some Name",
  "age": 42
}
```
... the properties `name` and `age` will not be assigned an RDF prefix and will be excluded when indexing the resource in the RDF graph.

**Describe the solution you'd like**
Delta should inject the default `@context` when one does not exist.

### LINKED ISSUE #4676: Looking for changes during the update of the resource can lead to maxing CPU
**Describe the bug**
In some cases, it seems that the check is causing cpu starvation.

- [ ] Fusion
- [X] Delta
- [ ] CLI
- [ ] Documentation

**Expected behavior**
When there is no change, no new revision is created and the resource usage remains normal.

imsdu:
Fixes #4676 
Fixes #3291


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/jsonld/JsonLdSourceProcessor.scala
```diff
@@ -31,21 +31,22 @@ sealed abstract class JsonLdSourceProcessor(implicit api: JsonLdApi) {
...
+    * Expand the source document using the provided project context and remote context resolution.
+    *
+    * If the source does not provide a context, one will be injected from the project base and vocab.
+    */
+  protected def expandSource(projectContext: ProjectContext, source: Json)(implicit
+      rcr: RemoteContextResolution
+  ): IO[(ContextValue, ExplainResult[ExpandedJsonLd])] = {
+    implicit val opts: JsonLdOptions = JsonLdOptions(base = Some(projectContext.base.iri))
+    val sourceContext                = source.topContextValueOrEmpty
+    if (sourceContext.isEmpty) {
```
imsdu, line 44: This part is the change related to #3291 that allows to make the diff check easier

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/DetectChange.scala
```diff
@@ -0,0 +1,65 @@
...
+    */
+  private val Impl = new DetectChange {
+
+    override def apply(newValue: JsonLdAssembly, current: Current): IO[Boolean] =
+      IO.cede
+        .as(
+          newValue.types != current.types ||
+            newValue.remoteContexts != current.remoteContexts ||
+            newValue.compacted.ctx != current.compacted.ctx ||
+            newValue.source != current.source
```
olivergrabinski, line 58: Enough to guarantee that the expanded for is the same? We did check the graph before. I think it's ok just wondering

olivergrabinski (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/DetectChange.scala
imsdu: To me all the cases should be covered:
https://github.com/BlueBrain/nexus/pull/4684/files#diff-43071a23edb35dcb0a07b6c27a989adc39e7a4227b4a70cbb11b845404f89944
To have a change in the resulting expanded value and thus in the graph, there are 3 parameters: the source, the context defined in the resource, the base defined in the project config when the resource does not define one.

1. We check for a change in the source
2.  We check for the local context `.compacted.ctx` that may have been provided by the source or injected if the resource do not define one
3. We check for remote contexts defined in the resource context (We have all contexts here, even the transitive ones).
4. We check for the types, it allows to detect cases where the `@base` is not defined in the local context and so relies on the one defined in the project config for it.

olivergrabinski (approved):

PR #4690, opened 2024-01-24
Retrieve tagged value in a previous transaction, better error handling

imsdu:


delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/ScopedEventLog.scala
```diff
@@ -225,14 +227,20 @@ object ScopedEventLog {
...
+        def newTaggedState(event: E, state: S): IO[Option[(UserTag, S)]] =
+          tagger.tagWhen(event) match {
+            case Some((tag, rev)) if rev == state.rev =>
+              IO.some(tag -> state)
+            case Some((tag, rev))                     =>
               stateMachine
                 .computeState(eventStore.history(ref, id, Some(rev)))
-                .map(_.fold(noop) { s => stateStore.save(s, tag, Noop) })
+                .flatTap {
+                  case stateOpt if !stateOpt.exists(_.rev == rev) =>
```
dantb, line 238: To clarify, the error should be raised even when this returns `None`? Instead of a `noop`

imsdu: If it is none or if for some reason, the state could not be replayed up to the requested rev

dantb (approved):

PR #4712, opened 2024-02-02
Allow unconstrained resources to be updated even if schema are enforced

### LINKED ISSUE #4711: Enforcing schema does not behave as expected during updates
**Describe the bug**
When the configuration flag is enforced on a project and a user attempts to update the resource linked to an unconstrained schema without updating the schema, the resource can't be updated.

- [ ] Fusion
- [X] Delta
- [ ] CLI
- [ ] Documentation

**Expected behavior**
The resource should be updated

imsdu:
Fixes #4711

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ValidateResource.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ValidateResource.scala
@@ -3,17 +3,18 @@ package ch.epfl.bluebrain.nexus.delta.sdk.resources
 import cats.effect.IO
 import cats.syntax.all._
 import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
-import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.{contexts, schemas}
+import ch.epfl.bluebrain.nexus.delta.rdf.Vocabulary.contexts
 import ch.epfl.bluebrain.nexus.delta.rdf.jsonld.context.RemoteContextResolution
 import ch.epfl.bluebrain.nexus.delta.rdf.shacl.{ShaclEngine, ValidationReport}
+import ch.epfl.bluebrain.nexus.delta.sdk.identities.model.Caller
 import ch.epfl.bluebrain.nexus.delta.sdk.jsonld.JsonLdAssembly
 import ch.epfl.bluebrain.nexus.delta.sdk.model.ResourceF
 import ch.epfl.bluebrain.nexus.delta.sdk.resolvers.ResolverResolution.ResourceResolution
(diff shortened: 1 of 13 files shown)


delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/SchemaClaimSuite.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  test("Create with a schema runs validation") {
+    val claim = SchemaClaim.onCreate(project, definedSchema, caller)
+    assertValidation(claim)
+  }
+
+  test("Create unconstrained fails or skip validation") {
+    val claim = SchemaClaim.onCreate(project, unconstrained, caller)
+    failOrNoValidation(claim)
+  }
```
shinyhappydan, line 53: I would consider changing this test to make it clear what the conditions are. Really, this should be two tests. It's better that the test is more verbose but more understandable

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ValidateResource.scala
```diff
@@ -58,24 +59,18 @@ object ValidateResource {
...
-            assertNotReservedId(jsonld.id) >>
-            IO.pure(NoValidation(schemaClaim.project))
-        else
-          for {
-            schema <- resolveSchema(schemaClaim)
-            result <- apply(jsonld, schema)
-          } yield result
+      ): IO[ValidationResult] = {
+        val submitOnDefinedSchema: SubmitOnDefinedSchema = resolveSchema(_, _, _).flatMap(apply(jsonld, _))
+        assertNotReservedId(jsonld.id) >> schemaClaim.validate(enforceSchema)(submitOnDefinedSchema)
```
shinyhappydan, line 64: I think this pattern is not helping make clear what is going on here. I would suggest either `schemaClaim.validate` should return something that could be mapped on, or else just combine the functions. This callback style is really difficult to grok (despite what is happening being quite simple)

shinyhappydan (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ValidateResource.scala
shinyhappydan: I don't think this is equivalent to `IO.whenA`, as I don't have a problem with that. There are other patterns possible which would keep the validation logic in the same place. You don't have to do it, I'm just pointing out that this pattern makes the code harder to navigate and I don't think I'm the only one who would think this

imsdu: They are both callbacks to me :)

PR #4713, opened 2024-02-02
Allow to disable the creation of a default storage at project creation

### LINKED ISSUE #4634: Disable the creation of a default storage by configuration
**Motivation**
When creating a project, a new default local storage is provisioned with it with default values.
In some deployments, we would like to disable it because we want to set custom values for the quota or the permission for example

**Acceptance criteria**
1. A new configuration flag is introduced to enable the creation of this default storage
2. By default, the storage is created
3. Make sure that the error when creating a file and no storage is available is clear.

imsdu:
Fixes #4634

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePluginModule.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePluginModule.scala
@@ -306,7 +306,7 @@ class StoragePluginModule(priority: Int) extends ModuleDef {
   }
 
   many[ScopeInitialization].addSet { (storages: Storages, serviceAccount: ServiceAccount, cfg: StoragePluginConfig) =>
-    StorageScopeInitialization.when(cfg.enableDefault)(storages, serviceAccount, cfg.defaults).toSet
+    Option.when(cfg.enableDefaultCreation)(new StorageScopeInitialization(storages, serviceAccount, cfg.defaults)).toSet
   }
 
   many[ProjectDeletionTask].add { (storages: Storages) => StorageDeletionTask(storages) }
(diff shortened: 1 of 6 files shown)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StoragePluginModule.scala
```diff
@@ -305,13 +305,10 @@ class StoragePluginModule(priority: Int) extends ModuleDef {
         .getOrElse(Set.empty[RemoteStorageServiceDependency])
   }
 
-  make[StorageScopeInitialization].from {
-    (storages: Storages, serviceAccount: ServiceAccount, cfg: StoragePluginConfig) =>
-      new StorageScopeInitialization(storages, serviceAccount, cfg.defaults)
+  many[ScopeInitialization].addSet { (storages: Storages, serviceAccount: ServiceAccount, cfg: StoragePluginConfig) =>
+    StorageScopeInitialization.when(cfg.enableDefault)(storages, serviceAccount, cfg.defaults).toSet
```
shinyhappydan, line 309: IMHO this is less clear than just using `Option.when`

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StorageScopeInitializationSpec.scala
```diff
@@ -82,5 +82,13 @@ class StorageScopeInitializationSpec
       val resource = storages.fetch(nxv + "diskStorageDefault", project.ref).accepted
       resource.rev shouldEqual 1L
     }
+
+    "be instantiated when enabled" in {
+      StorageScopeInitialization.when(enabled = true)(storages, sa, defaults) shouldBe defined
```
shinyhappydan, line 87: I think unless you can test our distage somehow, then in order to test this properly we would need to refactor how scopes are created. 

My point here is that even if this function works correctly, this test doesn't ensure that this is the method that gets called

shinyhappydan (approved):

PR #4715, opened 2024-02-06
Do not store the endpoint for remote storages and use the one in the configuration

### LINKED ISSUE #4714: Do not store the endpoint for remote storages and use the one in the configuration
**Motivation**
Since #4062, remote storage credentials are not stored within the storage because it makes it simpler to update them when they change.
This is similar, we want to only rely on the endpoint defined in the configuration so that is endpoint can be easily changed.
The remote storage is also something internal to Nexus so it also makes more sense to keep it as an internal configuration.

**Acceptance criteria**
1. The calls to the remote storage rely on the endpoint defined in the configuration
2. The api do not provide the ability to define another endpoint for storages
3. The information is no longer stored in database.

imsdu:
Fixes #4714 

There would be further refactorings but I did not want to get the PR too complex.


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/remote/client/RemoteDiskStorageClient.scala
```diff
@@ -35,39 +37,20 @@ import scala.concurrent.duration._
 /**
   * The client to communicate with the remote storage service
   */
-final class RemoteDiskStorageClient(client: HttpClient, getAuthToken: AuthTokenProvider, credentials: Credentials)(
-    implicit as: ActorSystem
-) {
-  import as.dispatcher
-
-  private val serviceName = Name.unsafe("remoteStorage")
+trait RemoteDiskStorageClient {
```
imsdu, line 40: This is being a dependency of storages and files, that seems the best compromise to do the refactoring without spending too much time on it.

delta/plugins/storage/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/remotestorage/RemoteStorageClientFixtures.scala
```diff
@@ -1,14 +1,22 @@
...
+import ch.epfl.bluebrain.nexus.testkit.scalatest.BaseSpec
+import org.scalatest.BeforeAndAfterAll
 
 import java.nio.file.attribute.PosixFilePermissions
 import java.nio.file.{Files, Path}
 import scala.concurrent.duration.DurationInt
 import scala.jdk.DurationConverters.ScalaDurationOps
 
-trait RemoteStorageDocker extends BeforeAndAfterAll { this: Suite =>
+trait RemoteStorageClientFixtures extends BeforeAndAfterAll with ConfigFixtures { this: BaseSpec =>
```
imsdu, line 19: Same approach as for Postgres and Elasticsearch for example, we don't care about the docker container itself but about a client we can build from it

olivergrabinski (approved):

olivergrabinski:
We could have a migration todo list somewhere so we don't have to remember what we might need to do

imsdu:
> We could have a migration todo list somewhere so we don't have to remember what we might need to do

There is an epic, I will add a task to it

Issue #4663, opened 2024-01-16
Introduce type hierarchy resolution


imsdu:
**Motivation**
Most resources depend on a hierarchy of types.
They have:
* A concrete type (ex: `ReconstructedNeuronMorphology`)
* A list of a more abstract types (ex: `NeuronMorphology`)

Both are important in different use cases and until now, the clients would have to provide the list of types which is error-prone and difficult to maintain.

We would like the client to only take care of providing the concrete type and Delta would take care on resolving and introducing the hierarchy of types.

This first ticket deals with introducing a new global entity responsible for this mapping between a concrete type and its hierarchy.

**Acceptance criteria**
1. An endpoint is introduced to append/replace/delete entries in the same fashion as permissions: 
https://bluebrainnexus.io/docs/delta/api/permissions-api.html 
2. An endpoint is provided to fetch this mapping (latest and specific revision)

**Not in scope**
1. The implementation in resources

crisely09:
I believe there won't be an endpoint to provide the information, the backbone of the ontologies can be added as another file in the already existing resources

PR #4718, opened 2024-02-07
Add type hierarchy model

olivergrabinski:
Part of #4663


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/typehierarchy/model/TypeHierarchyRejection.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+    * Signals that the provided revision does not match the latest revision
+    *
+    * @param provided
+    *   provided revision
+    * @param expected
+    *   latest know revision
+    */
+  final case class IncorrectRev(provided: Int, expected: Int)
```
imsdu, line 38: We could drop this in this case as the risk of concurrent updates will be very low.

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/typehierarchy/model/TypeHierarchyEvent.scala
```diff
@@ -0,0 +1,76 @@
...
+    *   the type hierarchy mapping
+    * @param rev
+    *   the revision number
+    * @param instant
+    *   the instant when the event was recorded
+    * @param subject
+    *   the subject that created the event
+    */
+  final case class TypeHierarchyCreated(
+      id: Iri,
```
imsdu, line 39: We don't need to define an id as it will be a singleton (like permissions)

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/typehierarchy/model/TypeHierarchyRejection.scala
olivergrabinski: Ok but what do we do with the rejection that needs to be provided if a unique violation happens? It can't be a partial match there. Just some generic rejection, assuming it won't happen?

imsdu: Sorry, I forgot about this one :).
What I meant is that we don't have to handle it in the api endpoint and in the state machine

olivergrabinski: Done 👍

imsdu (approved):

PR #4720, opened 2024-02-07
Add emodel fields to search

imsdu:


tests/docker/config/construct-query.sparql
```diff
@@ -673,9 +680,10 @@ CONSTRUCT {
     ?id a bmo:NeuronMorphologyFeatureAnnotation .
 
     OPTIONAL {
-      ?id                   nsg:hasTarget / nsg:hasSource      ?neuronMorphologyId ;
+      ?id                   nsg:hasTarget / nsg:hasSource      ?neuronMorphologyIdTmp ;
                             bmo:compartment                    ?featureCompartment ;
                             nsg:hasBody                        ?featureBody        .
+      BIND(STR(?neuronMorphologyIdTmp) AS ?neuronMorphologyId) .
```
imsdu, line 686: There was a nice side effect because of batching:
When there was a feature and the emodel in the same batch, one was picking just the id and the other the id and the name and it led to indexing errors as we were indexing `neuronMorphologyId` as an object.
As we only want the id here, converting to string allow to solve

imsdu: You are right, in the end it will be cleaner

crisely09 (approved):

PR #4729, opened 2024-02-09
Add type hierarchy route

olivergrabinski:
Part of #4663


olivergrabinski:
The main thing remaining is to review if the `TypeHierarchyRoutesSpec` and `TypeHierarchySpec` are sufficient. The issues I'm running into:
* How to test the serialization of the responses without having the reimplement the `TypeHierarchy` logic in the `RoutesSpec`.
* Making the tests in `TypeHierarchySpec` is slightly tricky because there is only one `TypeHierarchy`. So it makes it harder to have a spec that can run twice in a row without failing. For now I have added a `Ref` but maybe there is a smarter way.
* Also, I realized that having `model.TypeHierarchy` and `TypeHierarchy` is quite confusion. At the same time, I haven't found a better name for the `TypeHierarchy` trait that allows to create/update.

Finally, a new doc page is missing.

imsdu (approved):

imsdu (approved):

PR #4745, opened 2024-02-19
Do not create an event when updating attributes and digest is not computed

### LINKED ISSUE #4743: When computing the file attributes asynchronously, events are created when they are not yet computed
**Describe the bug**
When a file is linked, its attributes (size + checksum) are computed asynchronously by a task.
This task currently creates events even if attributes are not yet computed

- [ ] Fusion
- [X] Delta
- [ ] CLI
- [ ] Documentation

**Expected behavior**
An event is registered only when the attributes are computed

imsdu:
Fixes #4743


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/FileAttributesUpdateStream.scala
```diff
@@ -0,0 +1,122 @@
...
+import fs2.Stream
+import retry.RetryPolicies.constantDelay
+import retry.implicits.retrySyntaxError
+
+import scala.concurrent.duration.FiniteDuration
+
+/**
+  * Stream that attempts to update file attributes asynchronously for linked files in remote storages
+  */
+sealed trait FileAttributesUpdateStream
```
imsdu, line 25: Separating from `Files` makes it easier and more consistent with the other long-running tasks (like indexing)

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesConfig.scala
```diff
@@ -206,7 +207,7 @@ object StoragesConfig {
       defaultWritePermission: Permission,
       showLocation: Boolean,
       defaultMaxFileSize: Long,
-      digestComputation: RetryStrategyConfig
+      digestComputationRetryDelay: FiniteDuration
```
imsdu, line 210: Here, we always want to retry so it is better to just leave the opportunity to define a delay between them.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -700,6 +624,7 @@ object Files {
       case Some(s) if s.rev != c.rev               => IO.raiseError(IncorrectRev(c.rev, s.rev))
       case Some(s) if s.deprecated                 => IO.raiseError(FileIsDeprecated(c.id))
       case Some(s) if s.attributes.digest.computed => IO.raiseError(DigestAlreadyComputed(s.id))
+      case Some(s) if !c.digest.computed           => IO.raiseError(DigestNotComputed(s.id))
```
olivergrabinski, line 627: It feels like the issue would have not occurred in the first place if the digest fields in the attribute was an Option maybe?

olivergrabinski (approved):

PR #4756, opened 2024-02-23
Refactor FetchContext + add resolver to import batch

### LINKED ISSUE #4752: Remove `WrappedOrganizationRejection` from ProjectRejection
**Motivation**
With Cats-Effect and the new way we have to handle rejections, we don't need as much to wrap rejections.
`WrappedOrganizationRejection` can be removed.

**Acceptance criteria**
1. `WrappedOrganizationRejection` is removed
2. The appropriate payload and error code is still returned to the user on a org error (ex: org not found/ org is deprecated)

imsdu:
Fixes #4752

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/FetchContext.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/FetchContext.scala
@@ -1,14 +1,17 @@
 package ch.epfl.bluebrain.nexus.delta.sdk.projects
 
 import cats.effect.IO
-import cats.syntax.all._
 import ch.epfl.bluebrain.nexus.delta.sdk.ProjectResource
-import ch.epfl.bluebrain.nexus.delta.sdk.organizations.Organizations
+import ch.epfl.bluebrain.nexus.delta.sdk.organizations.FetchActiveOrganization
 import ch.epfl.bluebrain.nexus.delta.sdk.projects.model.ProjectRejection._
-import ch.epfl.bluebrain.nexus.delta.sdk.projects.model.{ApiMappings, ProjectContext, ProjectRejection}
+import ch.epfl.bluebrain.nexus.delta.sdk.projects.model.{ApiMappings, ProjectContext, ProjectState}
 import ch.epfl.bluebrain.nexus.delta.sdk.quotas.Quotas
+import ch.epfl.bluebrain.nexus.delta.sourcing.Transactors
 import ch.epfl.bluebrain.nexus.delta.sourcing.model.Identity.Subject
 import ch.epfl.bluebrain.nexus.delta.sourcing.model.{Label, ProjectRef}
+import ch.epfl.bluebrain.nexus.delta.sourcing.state.ScopedStateGet
+import doobie.Get
+import doobie.implicits._
 
 /**
   * Define the rules to fetch project context for read and write operations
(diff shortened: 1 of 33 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/FetchContext.scala
```diff
@@ -49,40 +52,47 @@ abstract class FetchContext { self =>
 
 object FetchContext {
 
-  /**
-    * Create a fetch context instance from an [[Organizations]], [[Projects]] and [[Quotas]] instances
-    */
-  def apply(organizations: Organizations, projects: Projects, quotas: Quotas): FetchContext =
```
imsdu, line 55: To break some circular dependencies

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/state/ScopedStateGet.scala
```diff
@@ -0,0 +1,18 @@
+package ch.epfl.bluebrain.nexus.delta.sourcing.state
+
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.Tag.Latest
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.{EntityType, ProjectRef, Tag}
+import doobie.implicits._
+import doobie.{ConnectionIO, Get, Put}
+
+object ScopedStateGet {
+
+  def apply[Id: Put, S: Get](tpe: EntityType, project: ProjectRef, id: Id, tag: Tag): ConnectionIO[Option[S]] =
```
imsdu, line 10: Allow to fetch a state without having a dependency on the whole project module.
Meant to be extended to other cases to break more circular dependencies

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/state/GlobalStateGet.scala
```diff
@@ -0,0 +1,13 @@
+package ch.epfl.bluebrain.nexus.delta.sourcing.state
+
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.EntityType
+import doobie.implicits._
+import doobie.{ConnectionIO, Get, Put}
+
+object GlobalStateGet {
```
dantb, line 7: Wonder if this could be a single method trait and have `GlobalStateStore` implement it? That way the transactor is encapsulated and dependents like `FetchContext` and `FetchActiveOrganisation` can rely on the abstraction (but will be passed the regular `GlobalStateStore` in the production code)

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/state/ScopedStateGet.scala
dantb: Same suggestion as `GlobalStateGet`, could it be a trait?

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/state/GlobalStateGet.scala
imsdu: The thing is that the store is also relying on other dependencies to be able to stream data (`QueryConfig`) and we don't need it to get a single one.

olivergrabinski (approved):

Issue #4520, opened 2023-11-21
Allow to add custom metadata to files during updates


imsdu:
**Motivation**
Related to #4512
Now that custom metadata is introduced, users may want to define it directly during file updates so as not to multiply the calls.

**Acceptance criteria**
1. When custom metadata is provided, it is directly associated to the updated file
2. Providing a file during updates is now optional but at least the file or the metadata should be provided

PR #4775, opened 2024-03-04
Add method & endpoint to update custom file metadata

olivergrabinski:
Related to #4520 

Screenshot 2024-03-04 at 16 28 55


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -222,6 +222,17 @@ final class Files(
     } yield res
   }.span("updateFile")
 
+  def updateMetadata(
```
imsdu, line 225: We may want to tag it too.
(Can be done in a later PR/ticket)

imsdu (approved):

PR #4780, opened 2024-03-05
Fix Elasticsearch default views resetter

### LINKED ISSUE #4727: The `ElasticSearchDefaultViewsResetter` can fail to reset some views
**Motivation**
When the `ElasticSearchDefaultViewsResetter` is enabled, we often encounter the following error on several projects
```
2024-02-09 08:00:48 ERROR c.e.b.n.d.p.e.i.ElasticSearchDefaultViewsResetter - Could not create view. Message: 'Resource 'https://bluebrain.github.io/nexus/vocabulary/defaultElasticSearchIndex' already exists in project 'org/project'.'
```
For the concerned projects, the default ES view seems correctly deleted, but fails to be recreated.

Possible (unconfirmed) explanation: read happens on the replica where the delete did not got replicated 

Solution: Make sure fetching the state is done on the primary node before evaluating a command

imsdu:
Fixes #4727 
* Limit the amount of transactions by grouping the deletes
* Read from the write node during evaluation to avoid bugs due to replication delays


delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchDefaultViewsResetter.scala
```diff
@@ -92,15 +93,15 @@ object ElasticSearchDefaultViewsResetter {
       override def resetDefaultViews: IO[Unit] =
         resetTrigger.flatMap { triggered =>
           IO.whenA(triggered) {
-            views.compile.toList
-              .flatMap { _.traverse { resetView } }
-              .flatMap { _ => logger.info("Completed resetting default elasticsearch views.") }
+            deleteDatabaseEntries >>
```
olivergrabinski, line 96: So was the fact that ES indices were deleted before the entries also an issue?

olivergrabinski (approved):

delta/plugins/elasticsearch/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/elasticsearch/indexing/ElasticSearchDefaultViewsResetter.scala
imsdu: I think it is a combination of that and the fact that we were reading from replicas when recreating the views yes.
Reducing the number of transactions is not a bad thing anyway, it will also help the WAL to get replicated.

PR #4783, opened 2024-03-05
Break circular dependencies

olivergrabinski:


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/FetchResource.scala
```diff
@@ -0,0 +1,29 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.resources.model.{Resource, ResourceState}
+import ch.epfl.bluebrain.nexus.delta.sourcing.Transactors
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.{ProjectRef, ResourceRef}
+import ch.epfl.bluebrain.nexus.delta.sourcing.state.ScopedStateGet
+import doobie.implicits._
+import doobie.{Get, Put}
+
+trait FetchResource {
+
+  def latest(ref: ResourceRef, project: ProjectRef): FetchF[Resource]
```
imsdu, line 14: Here it is not only the latest but possibly a revision or a tag too

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/FetchResource.scala
```diff
@@ -0,0 +1,29 @@
...
+}
+
+object FetchResource {
+
+  implicit private val getValue: Get[ResourceState] = ResourceState.serializer.getValue
+  implicit private val putId: Put[Iri]              = ResourceState.serializer.putId
+
+  def apply(xas: Transactors): FetchResource = (ref: ResourceRef, project: ProjectRef) =>
+    ScopedStateGet
+      .latest(Resources.entityType, project, ref.iri)
```
imsdu, line 25: Depending on the case you may want to resolve any type of resource

olivergrabinski: Does this imply we should have a function in `ScopedStateGet` that allows fetching the state without specifying the entity type?

imsdu: So far, we have to focus on the one related to resources/schemas, the general one is only used for resolver endpoints afaik so we should just leave it as it is for now as we don't need it for the batch

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/FetchResource.scala
```diff
@@ -0,0 +1,33 @@
...
+import ch.epfl.bluebrain.nexus.delta.rdf.IriOrBNode.Iri
+import ch.epfl.bluebrain.nexus.delta.sdk.model.Fetch.FetchF
+import ch.epfl.bluebrain.nexus.delta.sdk.resources.model.{Resource, ResourceState}
+import ch.epfl.bluebrain.nexus.delta.sourcing.Transactors
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.{ProjectRef, ResourceRef}
+import ch.epfl.bluebrain.nexus.delta.sourcing.state.ScopedStateGet
+import doobie.implicits._
+import doobie.{Get, Put}
+
+trait FetchResource {
```
imsdu, line 12: You could declare it in the distage module and use directly where it is used instead of passing along the transactors which is a bit low level ?
Any blocker to that ?

delta/sdk/src/test/scala/ch/epfl/bluebrain/nexus/delta/sdk/resolvers/ResolverContextResolutionSuite.scala
```diff
@@ -62,7 +62,7 @@ class ResolverContextResolutionSuite extends NexusSuite {
     )
   )
 
-  def fetchResource: (ResourceRef, ProjectRef) => FetchResource[Resource] = { (r: ResourceRef, p: ProjectRef) =>
+  def fetchResource: (ResourceRef, ProjectRef) => FetchF[Resource] = { (r: ResourceRef, p: ProjectRef) =>
```
imsdu, line 65: In the tests, you could have the `FetchResource` type too

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/state/ScopedStateGet.scala
```diff
@@ -8,11 +8,22 @@ import doobie.{ConnectionIO, Get, Put}
 object ScopedStateGet {
 
   def apply[Id: Put, S: Get](tpe: EntityType, project: ProjectRef, id: Id, tag: Tag): ConnectionIO[Option[S]] =
-    sql"""SELECT value FROM scoped_states WHERE type = $tpe AND org = ${project.organization} AND project = ${project.project}  AND id = $id AND tag = $tag"""
+    sql"""SELECT value FROM scoped_states WHERE type = $tpe AND org = ${project.organization} AND project = ${project.project} AND id = $id AND tag = $tag"""
+      .query[S]
+      .option
+
+  def apply[Id: Put, S: Get](tpe: EntityType, project: ProjectRef, id: Id, rev: Int): ConnectionIO[Option[S]] =
```
imsdu, line 15: We don't store the state for every revision, we have to query for events and replay them in a Stream (the code is already in `ScopedStateStore`, it could be moved here)

Issue #4790, opened 2024-03-08
`S3Storage*Spec`s failing on basic GitHub runner


olivergrabinski:
The following three tests are the only ones preventing us from running the plugins pipeline on the basic GitHub runner

<img width="857" alt="Screenshot 2024-03-08 at 16 41 24" src="https://github.com/BlueBrain/nexus/assets/20188437/62d7863e-7ddf-4143-9362-43d484f7cc5a">

PR #4792, opened 2024-03-13
Use GitHub cloud runners for more workflows

olivergrabinski:
The workflows that are moved from self-hosted to the basic cloud runners now run faster; this frees up the self-hosted runners for integration tests.

The plugin workflow is almost ready to run on the cloud runners as well, only #4790 needs to be dealt with first.

### DIFF
--- a/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/MainSuite.scala
+++ b/delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/MainSuite.scala
@@ -18,6 +18,7 @@ import munit.catseffect.IOFixture
 import munit.{AnyFixture, CatsEffectSuite}
 
 import java.nio.file.{Files, Paths}
+import scala.concurrent.duration.Duration
 
 /**
   * Test class that allows to check that across core and plugins:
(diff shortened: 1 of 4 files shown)


delta/app/src/test/scala/ch/epfl/bluebrain/nexus/delta/MainSuite.scala
```diff
@@ -27,6 +28,9 @@ import java.nio.file.{Files, Paths}
   */
 class MainSuite extends NexusSuite with MainSuite.Fixture {
 
+  // The default timeout of 30s is slightly too short for the GitHub free runners
+  override val munitIOTimeout: Duration = Duration(60, "s")
+
```
olivergrabinski, line 33: Locally the test takes about 22s. On the cloud runner, it was timing out without this change, so must be >30s. However, even with this change, the workflow runs faster on the cloud runner than on the self-hosted one.

imsdu: Yep it is starting a ES node so it can take some time
It could me nice to cut this dependency on this test as it does not bring value here as we want to check configuration/ distage wiring here (it should be about the events metric IIRC)

imsdu (approved):

PR #4794, opened 2024-03-13
Don't start an Elasticsearch container when running MainSuite

imsdu:


olivergrabinski (approved):
We could remove the 60s timeout in the `MainSuite`, no? (Just a note)

PR #4797, opened 2024-03-14
Remove the linking operation for S3 storages

### LINKED ISSUE #4784: Remove the linking operation from the S3 storage
**Motivation**
This feature has been implemented a while back to mirror the feature of the GPFS integration.
However:
* It has never been used
* It does not ensure that the file is in a protected folder
* It downloads the file to compute the checksum
* It is not fit for large files

As it does not match our needs and as we need to use another S3 library (because akka switched its license), it is better to remove it and replace it later with an appropriate

**Acceptance criteria**
1. The related code / tests are removed
2. The documentation is updated

dantb:
Fixes #4784


imsdu (approved):

imsdu:
LGTM.
Only a mention in the release notes are missing

PR #4798, opened 2024-03-14
Fix hostname validation on remote sources for composite views

imsdu:
Also removing the retry strategy for the http client for composite views as it only brings more problems in the end

### DIFF
--- a/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/client/DeltaClient.scala
+++ b/delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/client/DeltaClient.scala
@@ -26,10 +26,11 @@ import ch.epfl.bluebrain.nexus.delta.sourcing.offset.Offset
 import ch.epfl.bluebrain.nexus.delta.sourcing.offset.Offset.Start
 import ch.epfl.bluebrain.nexus.delta.sourcing.stream.{Elem, RemainingElems}
 import io.circe.parser.decode
+import ch.epfl.bluebrain.nexus.delta.sdk.http.HttpClientError.HttpTimeoutError
 import fs2._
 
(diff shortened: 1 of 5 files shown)


delta/plugins/composite-views/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/compositeviews/client/DeltaClient.scala
```diff
@@ -114,9 +115,10 @@ object DeltaClient {
     override def checkElems(source: RemoteProjectSource): IO[Unit] = {
       for {
         authToken <- authTokenProvider(credentials)
-        result    <- client(Head(elemAddress(source)).withCredentials(authToken)) {
+        request    = Head(elemAddress(source)).withCredentials(authToken)
+        result    <- client(request) {
                        case resp if resp.status.isSuccess() => IO.delay(resp.discardEntityBytes()) >> IO.unit
-                     }
+                     }.timeoutTo(3.seconds, IO.raiseError(HttpTimeoutError(request, "Head request took more than 3s")))
```
imsdu, line 121: Capping it to 3 seconds so it won't result in another timeout during evaluation

tests/docker/docker-compose.yml
```diff
@@ -159,20 +155,16 @@ services:
...
-    ports:
-      # Now only required if you need to access LocalStack from the host
-      - "127.0.0.1:4566:4566"
-      # Now only required if you need to access LocalStack from the host
-      - "127.0.0.1:4510-4559:4510-4559"
+    image: localstack/localstack:3.2
     environment:
-      - DEBUG=${DEBUG:-0}
+      AWS_ACCESS_KEY_ID: "MY_ACCESS_KEY"
+      AWS_SECRET_ACCESS_KEY: "CHUTCHUT"
```
dantb, line 163: Are these needed yet?

dantb (approved):

PR #4826, opened 2024-04-03
Do not validate schemas when importing with nexus ship

olivergrabinski:


ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/schemas/SchemaProcessor.scala
```diff
@@ -50,6 +50,9 @@ class SchemaProcessor private (schemas: Schemas, projectMapper: ProjectMapper, c
     {
       case a: ResourceAlreadyExists => logger.warn(a)("The schema already exists").as(ImportStatus.Dropped)
       case i: IncorrectRev          => logger.warn(i)("An incorrect revision has been provided").as(ImportStatus.Dropped)
+      case i: InvalidSchema         =>
```
imsdu, line 53: We don't need this as it won't happen if there is no validation

imsdu (approved):

PR #4832, opened 2024-04-03
Make `ValidatePriority` a trait

olivergrabinski:
Allows to provide a different implementation; used in the Nexus ship


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/wiring/ResolversModule.scala
```diff
@@ -33,10 +33,13 @@ object ResolversModule extends ModuleDef {
 
   implicit private val loader: ClasspathResourceLoader = ClasspathResourceLoader.withContext(getClass)
 
+  make[ValidatePriority].from { (xas: Transactors) => ValidatePriority.priorityAlreadyExists(xas) }
```
imsdu, line 36: It could only declared with the resolvers as it is not meant to used somewhere else

imsdu (approved):

PR #4836, opened 2024-04-04
Wire more contexts in nexus ship; use `source` instead of `value` in BG and ES processor

olivergrabinski:


ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/views/BlazegraphViewProcessor.scala
```diff
@@ -38,11 +39,20 @@ class BlazegraphViewProcessor private (
...
     implicit val s: Subject = event.subject
+    implicit val c: Caller  = Caller(s, Set.empty)
     val cRev                = event.rev - 1
     val project             = projectMapper.map(event.project)
     event match {
-      case e: BlazegraphViewCreated      => views(event.uuid).flatMap(_.create(e.id, project, e.value))
-      case e: BlazegraphViewUpdated      => views(event.uuid).flatMap(_.update(e.id, project, cRev, e.value))
+      case e: BlazegraphViewCreated      =>
+        e.id match {
+          case id if id == defaultViewId => views(event.uuid).flatMap(_.create(e.id, project, e.value))
```
imsdu, line 48: Can be done later but in the default views we could drop the events and just recreate them ?
That would allow to be sure that they are aligned and that none is deprecated for example

imsdu (approved):

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/views/BlazegraphViewProcessor.scala
olivergrabinski: 👍  #4844

PR #4837, opened 2024-04-04
Use Jena and Topbraid versions

### LINKED ISSUE #4683: Update the version of the Jena library
**Motivation**
We have not upgraded the Jena library for a long time and there are bug and security fixes we could benefit from.
It will also be a first step before adopting Json-LD 1.1 

**Acceptance criteria**
1. Delta use the last stable version of Jena 4.x
2. Jena will still be configured to use Json-LD 1.0 (the latest ones use Json-LD 1.1 as the default one) 
3. Exclude the titanium library to be sure we don't use Json-LD 1.1

**Not in scope**
1. Migrating to Json-LD 1.1

imsdu:
Fixes #4683

Before:
```
[info] NeuroMorphologySampleSpec:
[info] Creating a morphology
[info] - should succeed provisionning the mandatory schemas (5 seconds, 863 milliseconds)
[info] - should succeed creating the morphology without a schema (174 milliseconds)
[info] - should succeed creating the morphology with a schema (3 seconds, 227 milliseconds)
[info] - should succeed updating the morphology with a schema (3 seconds, 24 milliseconds)
```

After:
```
[info] NeuroMorphologySampleSpec:
[info] Creating a morphology
[info] - should succeed provisionning the mandatory schemas (2 seconds, 990 milliseconds)
[info] - should succeed creating the morphology without a schema (105 milliseconds)
[info] - should succeed creating the morphology with a schema (2 seconds, 203 milliseconds)
[info] - should succeed updating the morphology with a schema (2 seconds, 187 milliseconds)
```


build.sbt
```diff
@@ -109,7 +109,7 @@ lazy val fs2Aws                        = "io.laserdisc"                 %% "fs2-
 lazy val fs2AwsS3                      = "io.laserdisc"                 %% "fs2-aws-s3"                         % fs2AwsVersion
 lazy val googleAuthClient              = "com.google.oauth-client"       % "google-oauth-client"                % googleAuthClientVersion
 lazy val handleBars                    = "com.github.jknack"             % "handlebars"                         % handleBarsVersion
-lazy val jenaArq                       = "org.apache.jena"               % "jena-arq"                           % jenaVersion
+lazy val jenaArq                       = "org.apache.jena"               % "jena-arq"                           % jenaVersion exclude ("com.apicatalog", "titanium-json-ld")
```
imsdu, line 112: To make sure we don't use Json-LD 1.1 for now

tests/src/test/resources/kg/schemas/bbp/sample-neuromorphology.json
crisely09, line 1: To me it would be great if you could also check the validation of something of type `Dataset`, as it is also a complex shape, and most of our models *will* inherit it, including neuron morphologies. So using for example [this resource](https://bbp.epfl.ch/nexus/web/bbp/mmb-point-neuron-framework-model/resources/https:%2F%2Fbbp.epfl.ch%2Fdata%2Fbbp%2Fmmb-point-neuron-framework-model%2Fe813daac-08d0-4603-9f97-c05a2f17a7fd) as dataset would do.

delta/app/src/test/resources/schemas/errors/invalid-schema.json
```diff
@@ -21,6 +21,6 @@
       },
       "value": "nxv:42"
     },
-    "targetedNodes": 417
+    "targetedNodes": 412
```
crisely09, line 24: why is this changing?

tests/src/test/resources/kg/schemas/bbp/sample-neuromorphology.json
imsdu: Ok, I will write a ticket about it as it can help defuse some regressions early.
If we integrate more of those in our tests, we will have to update those to follow the changes so some help will be appreciated :)

delta/app/src/test/resources/schemas/errors/invalid-schema.json
imsdu: Not totally sure but we don't load the dash/tosh extensions provided by Topbraid anymore so my guess there are fewer constraints and fewer nodes can get validated in the end.
`targetedNodes` is also a Nexus thing and not sure and a node validated by several constraints will be accounted several times.

olivergrabinski (approved):

build.sbt
```diff
@@ -57,7 +57,7 @@ val postgresJdbcVersion        = "42.7.3"
 val pureconfigVersion          = "0.17.6"
 val scalaTestVersion           = "3.2.18"
 val scalaXmlVersion            = "2.2.0"
-val topBraidVersion            = "1.3.2" // 1.4.1 fails to validate some test schemas
```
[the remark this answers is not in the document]
imsdu: Only when loading the extensions from topbraid where they added more constraints in 1.4.x and some were breaking.
As we don't rely on them,  I made the change to only rely on those from shacl core

PR #4859, opened 2024-04-09
Remove endpoint and region from the S3 Storage instances

shinyhappydan:


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageFields.scala
```diff
@@ -151,8 +141,6 @@ object StorageFields {
       description: Option[String],
       default: Boolean,
       bucket: String,
-      endpoint: Option[Uri],
-      region: Option[Region],
       readPermission: Option[Permission],
       writePermission: Option[Permission],
       maxFileSize: Option[Long]
```
imsdu, line 146: Maybe this should go too ?
We don't have a way to enforce this with multipart.
The maximum size for a single put operation is 5G so it could also be a constant ?

shinyhappydan: Good idea

shinyhappydan: We decided to do this another time

olivergrabinski (approved):

PR #4862, opened 2024-04-10
Improve message for 'ResourceShaclEngineRejection'

### LINKED ISSUE #4830: Better handling of `ResourceShaclEngineRejection`
**Motivation**

If a resource cannot be resolved, the error message during resource validation can look like:

```
ch.epfl.bluebrain.nexus.delta.sdk.resources.model.ResourceRejection$ResourceShaclEngineRejection: 
Resource 'http://api.brain-map.org/api/v2/data/Structure/1' failed to produce a SHACL engine for schema 'https://neuroshapes.org/dash/ontologyentity?rev=1'.
```

which is not helpful to understand the problem.

imsdu:
Fixes #4830 

In the end, the problem was not related to a resource not found but to a static remote context (the shacl one provided by Nexus) required to generate the compacted representation of the validation report.
In the end, the problem happened because the import batch did not inject this static context so this PR is:
- Enriching the error message to make it clearer
- Giving it a 5xx status as it is a problem of the application

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ValidateResource.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/ValidateResource.scala
@@ -84,7 +84,7 @@ object ValidateResource {
           schema.value.shapes,
           reportDetails = true
         ).adaptError { e =>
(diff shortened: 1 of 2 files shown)


shinyhappydan (commented):
If the resource isn't found shouldn't we end up with a 404? Feels like we could split the error into two: a shacl engine error and a resource not found

PR #4863, opened 2024-04-10
Add ability to read import file/directory from S3 in nexus ship

olivergrabinski:


ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/EventStreamer.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  def s3eventStreamer(client: S3AsyncClientOp[IO], bucket: BucketName): EventStreamer = new EventStreamer {
+
+    override def streamLines(path: Path): Stream[IO, String] =
+      S3.create(client)
+        .readFile(bucket, FileKey.apply(NonEmptyString.unsafeFrom(path.toString)))
+        .through(text.utf8.decode)
+        .through(text.lines)
+
+    override def fileList(path: Path): IO[List[Path]] =
```
imsdu, line 61: Non json files should be filtered out ?

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/EventStreamer.scala
```diff
@@ -0,0 +1,86 @@
...
+
+object EventStreamer {
+
+  private val logger = Logger[EventStreamer]
+
+  def s3eventStreamer(client: S3AsyncClientOp[IO], bucket: BucketName): EventStreamer = new EventStreamer {
+
+    override def streamLines(path: Path): Stream[IO, String] =
+      S3.create(client)
+        .readFile(bucket, FileKey.apply(NonEmptyString.unsafeFrom(path.toString)))
```
shinyhappydan, line 57: This probably won't result in a failed IO..?

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/EventStreamer.scala
olivergrabinski: Yep, I think the current test work only because the non JSON file I have used in tests is completely empty

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/Main.scala
```diff
@@ -55,9 +70,18 @@ object Main
       _      <- logger.info(config.root().render())
     } yield ()
 
+  private def s3client(config: ShipConfig) = Interpreter[IO]
```
imsdu, line 73: Could be done later but we could reuse the S3 client in storage ?

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/config/S3Config.scala
```diff
@@ -0,0 +1,14 @@
+package ch.epfl.bluebrain.nexus.ship.config
+
+import pureconfig.ConfigReader
+import pureconfig.generic.semiauto.deriveReader
+
+final case class S3Config(endpoint: String, importBucket: String)
```
imsdu, line 6: There may be a reader for URI so you could directly use it here
For the bucket, you may have one too but adding another library for refined is a bit overkill for now.
An alternative could be to use `Label` and add a way to translate it to a `NonEmptyString`

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/EventStreamer.scala
olivergrabinski: Yeah... I made some changes; I still don't like it but it's all within the implementation now

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/config/S3Config.scala
olivergrabinski: I made the change for URI. For the bucket I don't really see the value of using the NonEmptyString here is, vs. the change I made which encapsulates the refinement within the implementation. And even if we went for using the BucketName in the config, we still need to handle the FileKey somewhere else... so the way it is now at least covers both cases at once. And it's not like we are going to start running the tool and specifying an empty string for the bucket

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/Main.scala
olivergrabinski: I made the change but it's not exactly the same listing method, because in the ship I don't want to list everything in the bucket, but everything with a specific prefix ("folder")

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClient.scala
```diff
@@ -35,27 +37,32 @@ object S3StorageClient {
...
-        .S3AsyncClientOpResource(
-          S3AsyncClient
-            .builder()
-            .credentialsProvider(creds)
-            .endpointOverride(URI.create(cfg.defaultEndpoint.toString()))
-            .forcePathStyle(true)
-            .region(Region.US_EAST_1)
-        )
-        .map(new S3StorageClientImpl(_, cfg.defaultEndpoint))
+      resource(URI.create(cfg.defaultEndpoint.toString()), creds)
```
imsdu, line 40: You can just use URI in the S3StorageConfig and get rid a akka one.
(There may be something to write a decoder/encoder for Circe but it should be easy)

PR #4868, opened 2024-04-15
View initialization is nexus ship

olivergrabinski:
View events for default ES and BG views are ignored. Instead, these default views are created after project creation.


ship/src/test/resources/config/views.conf
```diff
@@ -0,0 +1,5 @@
+ship {
```
imsdu, line 1: To delete ?

olivergrabinski: Deleted

imsdu (approved):

PR #4870, opened 2024-04-16
Create S3 storage on project creation during nexus ship import

olivergrabinski:


ship/src/main/resources/ship-default.conf
```diff
@@ -37,6 +37,85 @@ ship {
 
     # the bucket which contains the import files
     import-bucket = "nexus-ship-import"
+
+    storages {
```
olivergrabinski, line 41: It's a bit annoying but in order to create `Storages` we need the full `StoragesConfig` and it requires us to have all of this even if we don't use the local and remote disk

imsdu: Yep when S3 is done and if we need to introduce some support for Azure Blob Storage, we will need to have a look at this.

imsdu: Maybe playing with imports, you could avoid duplication for now ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/StorageScopeInitialization.scala
```diff
@@ -20,40 +22,32 @@ import ch.epfl.bluebrain.nexus.delta.sourcing.model.{EntityType, Identity, Proje
...
+  *   the id to use for the default storage to be created
+  * @param defaultStorageFields
+  *   the default value for the storage fields
   */
 class StorageScopeInitialization(
     storages: Storages,
     serviceAccount: ServiceAccount,
-    defaults: Defaults
+    defaultStorageId: Iri,
+    defaultStorageFields: StorageFields
```
imsdu, line 34: Not in the scope of this PR, but we may need to add a way for new projects to have a storage initialized in the same uniform way

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/config/S3Config.scala
```diff
@@ -11,7 +12,7 @@ import pureconfig.generic.semiauto.deriveReader
 import java.net.URI
 import scala.annotation.nowarn
 
-final case class S3Config(endpoint: URI, importBucket: BucketName)
+final case class S3Config(endpoint: URI, importBucket: BucketName, storages: StoragesConfig, defaultBucket: BucketName)
```
imsdu, line 15: This config is technical to me (like the one about the database), the storage one which brings Delta related things would be better somewhere else

imsdu (approved):

PR #4871, opened 2024-04-16
Refactor file operations behind new interfaces

dantb:
This is an itch I've wanted to scratch for a while, sorry it's bigger than anticipated. File operations were constructed at the call site in `Files` with every call, meaning implementation details (remote client, S3 client, effect system stuff) needed to be threaded all the way through the storage data structures. This tangled up data with functionality. It made it difficult to understand which parameters of the operations were dynamic and which dependencies were needed by which storage type.

I started implementing the S3 register file operation but it seemed better to clean this up first and extend from there. Hopefully this makes working with files a bit simpler on the different storages. Key bits:
1. Interfaces for the different operations `DiskFileOperations`, `RemoteDiskFileOperations`, `S3FileOperations`. They generally share the implementation details too, so was able to delete a lot of code by structuring this way. Most of these are just a few lines (though saving is more complicated).
2. Another interface `FileOperations` - this just delegates based on the storage provided. There is no way to avoid at compile time the operation and the storage not being compatible since it's fetched dynamically, but we can at least keep it to one place (i.e. raising the `UnsupportedOperation` errors). Actually there probably is a way, but it would involve changing the whole delta API so 🙅 
3. Pass this interface into `Files` + remove the dependencies.
4. Unit test via interfaces, not implementations. Means we can hack away behind the scenes without tests detecting every change and tends to lead to simpler / less coupled modules. (There is still a lot to do on this point)

Hopefully this pattern helps with extensibility - both horizontally (new methods, like S3 registration) and vertically (new implementations / deleting ones we deprecate). I should also say these changes are intended to not change any functionality.


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/DiskStorageSaveFile.scala
```diff
@@ -24,17 +25,17 @@ import java.nio.file._
 import java.util.UUID
 import scala.concurrent.Future
 
-final class DiskStorageSaveFile(storage: DiskStorage)(implicit as: ActorSystem, uuidf: UUIDF) extends SaveFile {
+final class DiskStorageSaveFile(implicit as: ActorSystem, uuidf: UUIDF) {
```
imsdu, line 28: Any reason it is not included in `DiskFileOperations` ?

imsdu (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/disk/DiskStorageSaveFile.scala
dantb: Only that it's a bit more involved, happy to move that and `S3StorageSaveFile` to have it all in one place

PR #4872, opened 2024-04-16
Save summary at the end of the batch

### LINKED ISSUE #4847: Save a summary of the batch run in database
**Motivation**
To easily get some information about a batch run and compare it with previous runs, we need to save a summary in database.

**Acceptance criteria**
1. A summary should be run if the outcome is a success or an error
2. The summary can contain:
- The start and end time of the batch
- The input parameters
- Its outcome (success/error)
- The error if one has been met
- The report as a json blob
- The last processed offset if it is a success

imsdu:
Fixes #4847

### DIFF
--- a/ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/Main.scala
+++ b/ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/Main.scala
@@ -1,6 +1,6 @@
 package ch.epfl.bluebrain.nexus.ship
 
-import cats.effect.{ExitCode, IO}
+import cats.effect.{Clock, ExitCode, IO}
 import cats.syntax.all._
 import ch.epfl.bluebrain.nexus.delta.kernel.Logger
 import ch.epfl.bluebrain.nexus.delta.ship.BuildInfo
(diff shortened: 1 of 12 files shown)


ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/Main.scala
```diff
@@ -56,13 +56,23 @@ object Main
...
+    val clock = Clock[IO]
+    InitShip(r).use { case (config, eventsStream, xas) =>
+      for {
+        start         <- clock.realTimeInstant
+        reportOrError <- Transactors
+                           .init(config.database)
+                           .use { xas =>
+                             RunShip(eventsStream, config.input, xas)
+                           }
+                           .attempt
```
olivergrabinski, line 69: Why not use the existing transactors?

olivergrabinski (approved):

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/Main.scala
imsdu: Good catch, I missed to remove the previous one

PR #4878, opened 2024-04-17
Register S3 files in-place

dantb:
This conflicts with your PR @shinyhappydan, I'll wait until yours is landed :)

This API is a first pass and the integration test is basic.


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/S3StorageSpec.scala
```diff
@@ -145,4 +148,24 @@ class S3StorageSpec extends StorageSpec {
...
+      val id      = genId()
+      val path    = s"$id/nexus-logo.png"
+      uploadLogoFile(path)
+      val payload = Json.obj("path" -> Json.fromString(path))
+      deltaClient.put[Json](s"/files/$projectRef/register/$id?storage=nxv:$storageId", payload, Coyote) {
+        (_, response) => response.status shouldEqual StatusCodes.Created
+      } >>
+        deltaClient.get[Json](s"/files/$projectRef/$id", Coyote) { (json, response) =>
+          response.status shouldEqual StatusCodes.OK
+          json.findAllByKey("_mediaType").headOption shouldEqual Some("image/png".asJson)
```
shinyhappydan, line 163: There's an example in `ResourceMatchers`:
`json should haveField("_mediaType", "image/png")`

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -229,6 +229,33 @@ final class Files(
     } yield res
   }.span("updateFileMetadata")
 
+  def registerFile(
```
imsdu, line 232: Not sure about the register keyword, I don't know if there could be something clearer ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -229,6 +229,33 @@ final class Files(
...
+      tag: Option[UserTag]
+  )(implicit caller: Caller): IO[FileResource] = {
+    for {
+      (iri, pc)             <- id.expandIri(fetchContext.onCreate)
+      (storageRef, storage) <- fetchAndValidateActiveStorage(storageId, id.project, pc)
+      s3Metadata            <- fileOperations.register(storage, path)
+      filename              <- IO.fromOption(path.lastSegment)(InvalidFilePath)
+      attr                   = FileAttributes.from(FileDescription(filename, s3Metadata.contentType.some, metadata), s3Metadata.metadata)
+      res                   <- eval(
+                                 CreateFile(
```
imsdu, line 246: Could be done in another PR but there is the update case too

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3FileOperations.scala
```diff
@@ -63,6 +74,34 @@ object S3FileOperations {
 
     override def save(storage: S3Storage, filename: String, entity: BodyPartEntity): IO[FileStorageMetadata] =
       saveFile.apply(storage, filename, entity)
+
+    override def register(bucket: String, path: Uri.Path): IO[S3FileMetadata] = {
+      for {
+        _           <- log.info(s"Fetching attributes for S3 file. Bucket $bucket at path $path")
+        resp        <- client.headObject(bucket, path.toString())
```
imsdu, line 81: Why not using GetObjectAttributes:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAttributes.html

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
dantb: I thought "link" seemed clearer and more accurate than its current usage, but has too much historical baggage 😅

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3FileOperations.scala
```diff
@@ -63,6 +74,34 @@ object S3FileOperations {
...
+    private def parseContentType(raw: String): IO[ContentType] =
+      ContentType.parse(raw).map(_.pure[IO]).getOrElse(IO.raiseError(InvalidContentType(raw)))
+
+    private def mkS3Metadata(bucket: String, path: Uri.Path, resp: HeadObjectResponse, ct: ContentType) = {
+      S3FileMetadata(
+        ct,
+        FileStorageMetadata(
+          UUID.randomUUID(), // TODO unused field for this use case, but mandatory everywhere?
+          resp.contentLength(),
+          Digest.ComputedDigest(DigestAlgorithm.MD5, resp.eTag().filterNot(_ == '"')),
```
imsdu, line 98: It may not be MD5, it is better to check for the checksum part in:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectAttributes.html

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3FileOperations.scala
```diff
@@ -63,6 +74,34 @@ object S3FileOperations {
...
+      }
+
+    private def parseContentType(raw: String): IO[ContentType] =
+      ContentType.parse(raw).map(_.pure[IO]).getOrElse(IO.raiseError(InvalidContentType(raw)))
+
+    private def mkS3Metadata(bucket: String, path: Uri.Path, resp: HeadObjectResponse, ct: ContentType) = {
+      S3FileMetadata(
+        ct,
+        FileStorageMetadata(
+          UUID.randomUUID(), // TODO unused field for this use case, but mandatory everywhere?
```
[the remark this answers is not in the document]
imsdu: We can keep it as this for now

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClient.scala
```diff
@@ -69,20 +69,10 @@ object S3StorageClient {
     override def readFile(bucket: BucketName, fileKey: FileKey): Stream[IO, Byte] =
       s3.readFile(bucket, fileKey)
 
-    override def getFileAttributes(bucket: String, key: String): IO[GetObjectAttributesResponse] =
-      client
-        .getObjectAttributes(
```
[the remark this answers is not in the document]
imsdu: There is the object size no ?
It is not populated ?
Is it just a localstack problem ? Even with the latest version ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
dantb: Once it is registered, can it be treated like a regular S3 file with the existing update endpoint?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutes.scala
```diff
@@ -329,15 +355,18 @@ object FilesRoutes {
       metadata: Option[FileCustomMetadata]
   )
 
+  final case class RegisterFileRequest(path: Path, metadata: Option[FileCustomMetadata])
```
imsdu, line 358: like the link operation, it would be good to leave the ability to pass the media type so that it takes precedence on the detected one.
Because there are examples of some extensions like asc which does not mean the same in the BBP context.

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
olivergrabinski: In the context of the nexus ship, when we get an FileUpdated event, we will need to perform this new operation again because the file is already in the right location and we want to just register it

dantb: Could maybe do a noop in the case of `FileAttributesOrigin.External`? Since as you said the updated file will already be present after the FileCreated event was processed

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClient.scala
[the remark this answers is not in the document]
dantb: The response fields are not mandatory. In this case, only a timestamp and the etag was returned. I'll check against the real S3 and new localstack

olivergrabinski (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
imsdu: What about something like bind ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3FileOperations.scala
imsdu: Hum, I just thought that head did not return the required info but if it does, I am fine with it

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3FileOperations.scala
```diff
@@ -63,6 +75,48 @@ object S3FileOperations {
...
+          checksum,
+          FileAttributesOrigin.External,
+          client.baseEndpoint / bucket / path,
+          path
+        )
+      )
+    }
+
+    private def checksumFrom(response: HeadObjectResponse) = IO.fromOption {
+      Option(response.checksumSHA256())
```
imsdu, line 112: We should check at least for SHA-1 too

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3FileOperations.scala
```diff
@@ -63,6 +75,48 @@ object S3FileOperations {
...
+        .map { checksum =>
+          Digest.ComputedDigest(
+            DigestAlgorithm.default,
+            Hex.encodeHexString(Base64.getDecoder.decode(checksum))
```
imsdu, line 116: I saw it in the other S3 PR too, we could have a utility method for this ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3FileOperations.scala
```diff
@@ -63,6 +75,48 @@ object S3FileOperations {
...
+          )
+        }
+    }(new IllegalArgumentException("Missing checksum"))
```
imsdu, line 119: Can be done in another PR but could it be controlled by a configuration flag ?

PR #4880, opened 2024-04-18
Patch distributions in import batch

### LINKED ISSUE #4876: Patching distributions content urls in resources original payload
**Motivation**
Those links are dependent to the environments as they represent the endpoint to download files via the Nexus API

**Acceptance criteria**
1. The base uri should be updated to match the new one
2. The org and project segments should be updated when the project is renamed in the new environment

imsdu:
Fixes #4876

### DIFF
--- a/delta/sourcing-psql/src/main/resources/scripts/postgres/init/V1_10_M09_001__ship_original_project_context.ddl
+++ b/delta/sourcing-psql/src/main/resources/scripts/postgres/init/V1_10_M09_001__ship_original_project_context.ddl
@@ -1,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS public.ship_original_project_context(
+    org      text         NOT NULL,
+    project  text         NOT NULL,
+    context  JSONB        NOT NULL,
+    PRIMARY KEY(org, project)
+)
+
--- a/ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/RunShip.scala
+++ b/ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/RunShip.scala
@@ -35,8 +35,8 @@ object RunShip {
         val fetchContext           = FetchContext(ApiMappings.empty, xas, Quotas.disabled)
         val originalProjectContext = new OriginalProjectContext(xas)
         val eventLogConfig         = config.eventLog
-        val originalBase           = config.originalBase
-        val targetBase             = config.targetBase
+        val originalBaseUri        = config.originalBaseUri
+        val targetBaseUri          = config.targetBaseUri
         val projectMapper          = ProjectMapper(config.projectMapping)
         for {
           // Provision organizations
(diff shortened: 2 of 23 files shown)


delta/sourcing-psql/src/main/resources/scripts/postgres/init/V1_10_M09_001__ship_original_project_context.ddl
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS public.ship_original_project_context(
```
imsdu, line 1: To save the original project context to expand selfs when needed

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/RunShip.scala
```diff
@@ -28,12 +30,14 @@ object RunShip {
...
+        val orgProvider            =
           OrganizationProvider(config.eventLog, config.serviceAccount.value, xas, clock)(uuidF)
-        val fetchContext   = FetchContext(ApiMappings.empty, xas, Quotas.disabled)
-        val eventLogConfig = config.eventLog
-        val baseUri        = config.baseUri
-        val projectMapper  = ProjectMapper(config.projectMapping)
+        val fetchContext           = FetchContext(ApiMappings.empty, xas, Quotas.disabled)
+        val originalProjectContext = new OriginalProjectContext(xas)
+        val eventLogConfig         = config.eventLog
+        val originalBase           = config.originalBase
```
shinyhappydan, line 38: Including uri in the name is much clearer

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/projects/OriginalProjectContext.scala
```diff
@@ -0,0 +1,49 @@
...
+import doobie.implicits._
+import io.circe.generic.semiauto.deriveCodec
+import io.circe.syntax.EncoderOps
+import io.circe.{Codec, Json}
+
+/**
+  * Allows to keep track of the original project context which will change because of project renaming / patching
+  * configuration so as to be able to use them to expand iris in resource payload
+  */
+class OriginalProjectContext(xas: Transactors) extends FetchContext {
```
shinyhappydan, line 18: You don't need to implement FetchContext for this?

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/resources/DistributionPatcher.scala
```diff
@@ -0,0 +1,57 @@
...
+    * Distribution may be defined as an object or as an array in original payloads
+    */
+  def singleOrArray: Json => IO[Json] = root.distribution.json.modifyA { json =>
+    json.asArray match {
+      case Some(array) => array.traverse(single).map(Json.arr(_: _*))
+      case None        => single(json)
+    }
+  }(_)
+
+  def single: Json => IO[Json] = (json: Json) => fileContentUrl(json).map(toS3Location)
```
shinyhappydan, line 28: This method is never used in real code, the tests can easily call the other one

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/projects/OriginalProjectContext.scala
shinyhappydan: The only place it's needed is in FileSelf, so you could just split the interface. I think it's likely that a lot of places it would make sense to just pass `onRead` rather than all of these methods

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/projects/OriginalProjectContext.scala
```diff
@@ -0,0 +1,49 @@
...
+import ch.epfl.bluebrain.nexus.delta.sdk.projects.model.{ApiMappings, ProjectContext}
+import ch.epfl.bluebrain.nexus.delta.sourcing.Transactors
+import ch.epfl.bluebrain.nexus.delta.sourcing.implicits._
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.{Identity, ProjectRef}
+import doobie.implicits._
+import io.circe.generic.semiauto.deriveCodec
+import io.circe.syntax.EncoderOps
+import io.circe.{Codec, Json}
+
+/**
```
shinyhappydan, line 14: I guess this is going in postgres so we can import in multiple sessions?

ship/src/test/scala/ch/epfl/bluebrain/nexus/ship/resources/DistributionPatcherSuite.scala
```diff
@@ -0,0 +1,109 @@
...
+                }
+              }
+            }"""
+    val expected =
+      json"""{
+            "atLocation": {
+              "store": {
+                "@id": "https://bluebrain.github.io/nexus/vocabulary/defaultS3Storage",
+                "@type": "S3Storage",
+                "_rev": 1
```
shinyhappydan, line 64: It might be worth explaining why these fields are changing, because as a reader I can't tell currently

shinyhappydan (approved):

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/projects/OriginalProjectContext.scala
imsdu: Most of the methods come from `FetchContext` and the import batch is a special use case, if we are to refactor it, it is outside the scope of this PR

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/resources/DistributionPatcher.scala
imsdu: It would still be more complicated as an extra level would be required

PR #4882, opened 2024-04-22
Remove s3 md5

shinyhappydan:


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClient.scala
```diff
@@ -72,7 +88,75 @@ object S3StorageClient {
...
+            }
+          }
+        }
+    }
+
+    private def checksumFromResponse(response: PutObjectResponse, algorithm: DigestAlgorithm): String = {
+      algorithm.value match {
+        case "SHA-256" => Hex.encodeHexString(Base64.getDecoder.decode(response.checksumSHA256()))
+        case "SHA-1"   => Hex.encodeHexString(Base64.getDecoder.decode(response.checksumSHA1()))
+        case _         => throw new IllegalArgumentException(s"Unsupported algorithm for S3: ${algorithm.value}")
```
dantb, line 148: I know we're not consistent but would ideally be in `IO` and can be deduplicated with below

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClient.scala
```diff
@@ -28,12 +34,22 @@ trait S3StorageClient {
...
 
-  def underlyingClient: S3AsyncClientOp[IO]
+  def uploadFile(
+      fileData: Stream[IO, Byte],
+      bucket: String,
+      key: String,
+      algorithm: DigestAlgorithm
+  ): IO[UploadMetadata]
+
+  def objectExists(bucket: String, key: String): IO[Boolean]
```
dantb, line 44: Cleaner 👍  could also be a `final def` and written in terms of `headObject` on the interface

dantb (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClient.scala
shinyhappydan: I would rather not because there's no I/O happening (also, the exception will never get hit due to config validation)

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClient.scala
```diff
@@ -72,7 +88,75 @@ object S3StorageClient {
...
+    override def uploadFile(
+        fileData: Stream[IO, Byte],
+        bucket: String,
+        key: String,
+        algorithm: DigestAlgorithm
+    ): IO[UploadMetadata] = {
+      for {
+        fileSizeAcc <- Ref.of[IO, Long](0L)
+        digest      <- fileData
+                         .evalTap(_ => fileSizeAcc.update(_ + 1))
```
imsdu, line 111: The size could be retrieved by querying S3.
If it is not returned by the put operation, then we can follow the put by a headObject like the registering operation

PR #4893, opened 2024-04-23
Add endpoint to update previously registered S3 files

dantb:
@olivergrabinski is this enough to unblock you? Still need to verify all the file attribute fields in the E2E test response and the API design could be better.

Some points of note
1. We change the `FileAttributesOrigin` from `External` to `Client`.
2. The S3 path isn't provided by the client - it comes from the file state.
3. Generate another UUID that's stored in the event but isn't used.
4. Content type and filename now come from the request, other attributes are populated in the S3 client (same as for an initial save).

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
@@ -245,6 +245,37 @@ final class Files(
     } yield res
   }.span("registerFile")
 
+  def updateRegisteredFile(
+      id: FileId,
+      storageId: Option[IdSegment],
+      rev: Int,
+      entity: HttpEntity,
+      tag: Option[UserTag],
+      metadata: Option[FileCustomMetadata]
+  )(implicit caller: Caller): IO[FileResource] = {
+    for {
+      (iri, pc)             <- id.expandIri(fetchContext.onModify)
+      _                     <- test(UpdateFile(iri, id.project, testStorageRef, testStorageType, testAttributes, rev, caller.subject, tag))
+      (storageRef, storage) <- fetchAndValidateActiveStorage(storageId, id.project, pc)
+      uploadedInfo          <- formDataExtractor(iri, entity, storage.storageValue.maxFileSize)
+      path                  <- fetchState(id, iri).map(_.attributes.path)
+      storageMetadata       <- fileOperations.registerUpdate(storage, path, uploadedInfo.contents)
+      attr                   = FileAttributes.from(FileDescription.from(uploadedInfo, metadata), storageMetadata)
+      res                   <- eval(
+                                 UpdateFile(
+                                   iri,
+                                   id.project,
+                                   storageRef,
+                                   storage.tpe,
+                                   attr,
+                                   rev,
+                                   caller.subject,
+                                   tag
+                                 )
+                               )
+    } yield res
+  }.span("updateRegisteredFile")
+
   /**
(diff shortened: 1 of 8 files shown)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutes.scala
```diff
@@ -290,6 +290,32 @@ final class FilesRoutes(
                     }
                   }
                 }
+              },
+              pathPrefix("register-update") {
+                (idSegment & indexingMode) { (id, mode) =>
+                  pathEndOrSingleSlash {
+                    operationName(s"$prefixSegment/files/{org}/{project}/register-update/{id}") {
```
imsdu, line 297: Not important but we don't need this `operationName` anymore as the Kamon feature is disabled because it made Prometheus crash by sending data with a lot of distinct values (which Prometheus is not fond about)

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/Files.scala
```diff
@@ -245,6 +245,37 @@ final class Files(
     } yield res
   }.span("registerFile")
 
+  def updateRegisteredFile(
+      id: FileId,
+      storageId: Option[IdSegment],
+      rev: Int,
+      entity: HttpEntity,
```
imsdu, line 252: Why do we have a http entity here ?
It should be a path like the `registerFile` no ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutes.scala
```diff
@@ -290,6 +290,32 @@ final class FilesRoutes(
                     }
                   }
                 }
+              },
+              pathPrefix("register-update") {
```
imsdu, line 294: Can be fixed in a later PR but we have a problem, it clashes with this endpoint:
https://bluebrainnexus.io/docs/delta/api/files-api.html#create-using-put

imsdu: Not in the update operation but for the create one, we will need to provide a way to "register" a file where the user leaves Delta generating the id

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/StorageFileRejection.scala
```diff
@@ -138,6 +137,9 @@ object StorageFileRejection {
           s"File cannot be saved because it already exists on path '$path'."
         )
 
+    final case class FileNotFound(path: String)
```
imsdu, line 140: I don't know if we can rename it in a way that makes it different from the `FileNotFound` from `FileRejection` ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/routes/FilesRoutes.scala
dantb: Ah yeah this was intentional - can we do without the version where we generate the Id? Fewer endpoints => smaller API surface => less maintenance (tests, docs, code, etc)

PR #4903, opened 2024-04-24
Use update registered file in nexus ship; log import report

olivergrabinski:


ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/files/FileProcessor.scala
```diff
@@ -72,6 +74,14 @@ class FileProcessor private (
     {
       case a: ResourceAlreadyExists => logger.warn(a)("The resource already exists").as(ImportStatus.Dropped)
       case i: IncorrectRev          => logger.warn(i)("An incorrect revision has been provided").as(ImportStatus.Dropped)
+      case n: NoSuchKeyException    =>
```
imsdu, line 77: The file copier could handle the error itself (but it is not very important)

olivergrabinski: I will change it a bit later

PR #4907, opened 2024-04-25
Patch target projects for views within an aggregated blazegraph view

dantb:
Started with blazegraph only to check the approach is reasonable?


ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/views/ViewPatcher.scala
```diff
@@ -0,0 +1,19 @@
...
+import ch.epfl.bluebrain.nexus.ship.config.InputConfig.ProjectMapping
+import io.circe.Json
+import io.circe.optics.JsonPath.root
+
+class ViewPatcher(mapping: ProjectMapping) {
+
+  def patchAggregateViewSource(input: Json): Json =
+    root.views.each.project.string.modify { p =>
+      val projectRef = parseProjectRef(p)
+      mapping.getOrElse(projectRef, projectRef).toString
```
dantb, line 13: Continuing to default to the source project if not present in the mapping file

shinyhappydan (approved):

PR #4917, opened 2024-04-30
Add configured default bucket on S3 storages

### LINKED ISSUE #4886: Define a default bucket for S3 storages
**Motivation**
So far, the client needs to provide a bucket when creating a storage.
Defining a default one by configuration would allow to make it optional in the api endpoint

**Acceptance criteria**
1. If no bucket is defined in the request, we fallback to the one defined in the configuration

dantb:
Fixes #4886 

Will need a config change in AWS before upgrading the image to this version (S3 is disabled in all the kubernetes deployments)

### DIFF
--- a/tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/S3StorageSpec.scala
+++ b/tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/S3StorageSpec.scala
@@ -74,9 +74,12 @@ class S3StorageSpec extends StorageSpec {
 
   override def beforeAll(): Unit = {
     super.beforeAll()
-    (s3Client.createBucket(CreateBucketRequest.builder.bucket(bucket).build) >> uploadLogoFileToS3(logoKey)).accepted
+    (createBucket(bucket) >> uploadLogoFileToS3(logoKey)).accepted
     ()
   }
+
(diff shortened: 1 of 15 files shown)


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/S3StorageSpec.scala
```diff
@@ -164,6 +173,23 @@ class S3StorageSpec extends StorageSpec {
...
+        "kg/storages/s3.json",
+        "storageId" -> s"https://bluebrain.github.io/nexus/vocabulary/$id"
+      ).mapObject(_.remove("bucket"))
+
+      val expectedStorage = storageResponse(projectRef, id, defaultBucket, "resources/read", "files/write")
+
+      for {
+        _ <- createBucket(defaultBucket)
+        _ <- storagesDsl.createStorage(payload, projectRef)
+        _ <- storagesDsl.checkStorageMetadata(projectRef, id, expectedStorage)
```
shinyhappydan, line 190: Not a big deal but it would be clearer if the assertion was just on the bucket name

shinyhappydan (approved):

olivergrabinski:
The config for AWS is still in the `aws/infra` folder, you can edit it already if you want

imsdu (approved):

Issue #4896, opened 2024-04-24
Schema validation failing when importing an ontology


crisely09:
**Description**
Schema importing an ontology is failing validation for a "valid" resource.

**What part of Nexus is affected**
- [ ] Fusion
- [x] Delta
- [ ] CLI
- [ ] Documentation

**To Reproduce**
Steps to reproduce the behavior, use the trial endpoint on staging:
1. Use as body:
```
{
    "schema": "https://neuroshapes.org/dash/entity",
    "resource": {
        "@context": "https://bbp.neuroshapes.org",
        "@type": [
            "Entity"
        ],
        "brainLocation": {
            "@type": "BrainLocation",
            "brainRegion": {
                "@id": "mba:778",
                "label": "VISp5"
            }
        },
        "description": "This is an e-model with detailed morphology. See https://bbpteam.epfl.ch/project/issues/browse/BBPP152-12 for details",
        "name": "test entity",
        "@id": "https://bbp.epfl.ch/data/dke/kgforge/78bd293d-3ca4-4495-bba1-25dbf406a83d"
    }
}
```
2. Do a POST request to `https://staging.nise.bbp.epfl.ch/nexus/v1/trial/resources/neurosciencegraph/datamodels`
3. See error:
```
"error": {
        "@context": [
            "https://bluebrain.github.io/nexus/contexts/shacl-20170720.json",
            "https://bluebrain.github.io/nexus/contexts/error.json"
        ],
        "@type": "InvalidResource",
        "reason": "Resource 'https://bbp.epfl.ch/data/dke/kgforge/78bd293d-3ca4-4495-bba1-25dbf406a83d' failed to validate against the constraints defined in schema 'https://neuroshapes.org/dash/entity?rev=165'",
        "details": {
            "@type": "sh:ValidationReport",
            "conforms": false,
            "result": {
                "@type": "sh:ValidationResult",
                "detail": [
                    {
                        "@type": "sh:ValidationResult",
                        "focusNode": "https://bbp.epfl.ch/data/dke/kgforge/78bd293d-3ca4-4495-bba1-25dbf406a83d",
                        "resultMessage": "Value must have all of the following shapes: _:bfab0a149b19c29ece2f70ca48910d04, _:c3cf849ab8c9e8c5a667f911733a4497",
                        "resultSeverity": "sh:Violation",
                        "sourceConstraintComponent": "sh:AndConstraintComponent",
[...]

PR #4918, opened 2024-05-01
Add test to validate a brain region against an ontology

imsdu:
Related to #4896


tests/src/test/resources/kg/schemas/bbp/brain-region-ontology.json
```diff
@@ -0,0 +1,86 @@
...
+    {
+      "@id": "http://api.brain-map.org/api/v2/data/Structure/943",
+      "@type": "Class",
+      "label": "Primary motor area, layer 2/3",
+      "subClassOf": [
+        "https://neuroshapes.org/BrainRegion"
+      ]
+    }
+  ],
+  "distribution": [
```
olivergrabinski, line 23: I don't understand why we need the distribution here? It's not important but it should work without it?

olivergrabinski (approved):

tests/src/test/resources/kg/schemas/bbp/brain-region-ontology.json
imsdu: You are right, I could have get rid of it here (it was with the original ontology which is way larger)

PR #4923, opened 2024-05-02
Configure a global path prefix used when saving S3 files

### LINKED ISSUE #4885: Allow to define a prefix in a S3 storage
**Motivation**
The current S3 storage pushes files directly under the`/org/project` prefix while:
* Legacy jobs will want to mount the S3 bucket under a similar path
* The bucket may be used for other purposes

**Acceptance criteria**
1. The prefix is defined by configuration and is the same for all storages (like ES indices and BG namespaces)
* the resulting object key will be /prefix/org/project/....
* the path in file attributes would be the same value
2. The prefix can be empty
3. The prefix is used to compute the object key when pushing new files to the bucket

dantb:
Fixes #4885

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesConfig.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesConfig.scala
@@ -195,6 +195,10 @@ object StoragesConfig {
     *   flag to decide whether or not to show the absolute location of the files in the metadata response
     * @param defaultMaxFileSize
     *   the default maximum allowed file size (in bytes) for uploaded files
+    * @param defaultBucket
+    *   the bucket used for S3 storages if not specified during storage creation
+    * @param prefix
+    *   global prefix prepended to the path of all saved S3 files
     */
   final case class S3StorageConfig(
       digestAlgorithm: DigestAlgorithm,
@@ -206,7 +210,8 @@ object StoragesConfig {
       defaultWritePermission: Permission,
       showLocation: Boolean,
       defaultMaxFileSize: Long,
-      defaultBucket: String
+      defaultBucket: String,
+      prefix: Uri
   ) extends StorageTypeEntryConfig
 
   /**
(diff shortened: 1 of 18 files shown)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClient.scala
```diff
@@ -57,6 +57,8 @@ trait S3StorageClient {
   def bucketExists(bucket: String): IO[Boolean]
 
   def baseEndpoint: Uri
+
+  def prefix: Uri
```
dantb, line 61: I don't like this but it at least keeps the optionality of S3 storages to one place (otherwise you'd need to inject optionally to the call site)

infra/aws.conf
```diff
@@ -172,6 +172,7 @@ plugins {
         enabled = true
         use-default-credential-provider = true
         default-bucket = "nexus-delta-production"
+        prefix = ""
```
dantb, line 175: Do we know the value for this yet?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesConfig.scala
```diff
@@ -206,7 +210,8 @@ object StoragesConfig {
       defaultWritePermission: Permission,
       showLocation: Boolean,
       defaultMaxFileSize: Long,
-      defaultBucket: String
+      defaultBucket: String,
+      prefix: Uri
```
imsdu, line 214: It would be better to have an option here

infra/aws.conf
imsdu: It should be the same as the one defined for the storage service

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesConfig.scala
dantb: Agreed

imsdu (approved):

PR #4928, opened 2024-05-02
Deal with S3 download failures by streaming files

shinyhappydan:
The diff here is big but there are not many actual changes. The changes are:

- Usage of `AsyncResponseTransformer` (and some processing of what is returned in the `readFile` method) to actually stream when fetching a file
- Removal of `multiPartRead` since we can stream the file normally now
- Removal of `BucketName` / `FileKey` as they are no longer used
- Splitting `S3StorageClient` as the file was much too big. Privacy rules are used to ensure users cannot construct impls directly (this was actually happening before, my understanding is we prefer not to do this)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/Fs2StreamAsyncResponseTransformer.scala
```diff
@@ -0,0 +1,27 @@
...
+  override def onStream(publisher: SdkPublisher[ByteBuffer]): Unit = {
+    cf.complete(publisher)
+    ()
+  }
+
+  override def exceptionOccurred(error: Throwable): Unit = {
+    cf.completeExceptionally(error)
+    ()
+  }
+}
```
shinyhappydan, line 27: This is the main part of the functionality

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClientImpl.scala
```diff
@@ -0,0 +1,161 @@
...
+  override def listObjectsV2(bucket: String): IO[ListObjectsV2Response] =
+    client.listObjectsV2(ListObjectsV2Request.builder().bucket(bucket).build())
+
+  override def listObjectsV2(bucket: String, prefix: String): IO[ListObjectsV2Response] =
+    client.listObjectsV2(ListObjectsV2Request.builder().bucket(bucket).prefix(prefix).build())
+
+  override def readFile(bucket: String, fileKey: String): Stream[IO, Byte] = {
+    Stream
+      .eval(client.getObject(getObjectRequest(bucket, fileKey), new Fs2StreamAsyncResponseTransformer))
+      .flatMap(_.toStreamBuffered[IO](2).flatMap(bb => Stream.chunk(Chunk.byteBuffer(bb))))
```
shinyhappydan, line 31: This is where the streaming is happening

olivergrabinski (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClientDisabled.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import akka.http.scaladsl.model.Uri
+import cats.effect.IO
+import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.DigestAlgorithm
+import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.s3.client.S3StorageClient.{HeadObject, UploadMetadata}
+import ch.epfl.bluebrain.nexus.delta.sdk.error.ServiceError.FeatureDisabled
+import fs2.Stream
+import software.amazon.awssdk.services.s3.model._
+
+private[client] object S3StorageClientDisabled extends S3StorageClient {
```
imsdu, line 11: Not to address in this PR and not a priority but we may want to get rid of this one.
As long as you can't create a S3 storage when the feature is disabled, it might be enough ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClientImpl.scala
imsdu: What is the 2 for ?

ship/src/test/scala/ch/epfl/bluebrain/nexus/ship/RunShipSuite.scala
```diff
@@ -106,20 +105,20 @@ object RunShipSuite {
...
 
     override def headObject(bucket: String, key: String): IO[HeadObject] =
       IO.raiseError(new NotImplementedError("headObject is not implemented"))
 
     override def copyObject(
-        sourceBucket: Models.BucketName,
-        sourceKey: Models.FileKey,
-        destinationBucket: Models.BucketName,
-        destinationKey: Models.FileKey,
+        sourceBucket: String,
```
imsdu, line 118: It would still be nice to reintroduce something more refined than strings at some point

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClientDisabled.scala
shinyhappydan: I agree that this probably shouldn't be needed

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3FileOperations.scala
```diff
@@ -67,7 +67,7 @@ object S3FileOperations {
                 )
                 .compile
```
imsdu, line 68: Should not this go away no ?

ship/src/test/scala/ch/epfl/bluebrain/nexus/ship/RunShipSuite.scala
shinyhappydan: I think the problem with these types specifically is when you needed to use them as a string you had to call '.value.value' which is both annoying, and if you forget to do it in string interpolation for example you get a bad result. There might be other options which don't have these issues

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClientImpl.scala
shinyhappydan: It's the buffer size. It means that it'll ask for 2 bytebuffers at a time

shinyhappydan: (It was taken from the example you gave, it's not a value I chose)

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3FileOperations.scala
shinyhappydan: I've removed the compile, test will arrive soon

imsdu (approved):

PR #4929, opened 2024-05-03
Make S3 file prefix optional in type

dantb:


infra/aws.conf
```diff
@@ -172,7 +172,7 @@ plugins {
         enabled = true
         use-default-credential-provider = true
         default-bucket = "nexus-delta-production"
-        prefix = ""
+        prefix = "/sbo/data/project/"
```
dantb, line 175: Won't this break something? @imsdu @olivergrabinski 
It's what we have on prem but the folders do not yet exist in the `nexus-delta-production` bucket

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/StoragesConfig.scala
```diff
@@ -211,8 +211,10 @@ object StoragesConfig {
       showLocation: Boolean,
       defaultMaxFileSize: Long,
       defaultBucket: String,
-      prefix: Uri
-  ) extends StorageTypeEntryConfig
+      prefix: Option[Uri]
+  ) extends StorageTypeEntryConfig {
+    val prefixUri: Uri = prefix.getOrElse(Uri.Empty)
```
shinyhappydan, line 216: I'm not sure what the motivation for this is, but it feels weird that we have two fields available in the config, but we would only ever use one of them. Perhaps it's possible to specify the default to be an empty string, and have the parser deal with that correctly?

infra/aws.conf
imsdu: It is ok, I think

imsdu (approved):

PR #4933, opened 2024-05-06
Add S3 multipart copy and use it in Nexus ship

olivergrabinski:


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClientImpl.scala
```diff
@@ -76,6 +70,56 @@ final private[client] class S3StorageClientImpl(client: S3AsyncClientOp[IO], val
         .build()
     )
 
+  def copyObjectMultiPart(
+      sourceBucket: String,
+      sourceKey: String,
+      destinationBucket: String,
+      destinationKey: String
+  ): IO[CompleteMultipartUploadResponse] = {
+    val partSize = 5L * 1024L * 1024L // 5MB
```
imsdu, line 79: That may be a little low no ?

imsdu: Also some files are empty, we have to make sure it works, maybe for files lower than the part size we should not use multipart ?

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClientImpl.scala
```diff
@@ -76,6 +70,56 @@ final private[client] class S3StorageClientImpl(client: S3AsyncClientOp[IO], val
...
+        val uploadParts = (0L until objectSize by partSize).zipWithIndex.map { case (start, partNumber) =>
+          val lastByte              = Math.min(start + partSize - 1, objectSize - 1)
+          val uploadPartCopyRequest = UploadPartCopyRequest.builder
+            .sourceBucket(sourceBucket)
+            .sourceKey(sourceKey)
+            .destinationBucket(destinationBucket)
+            .destinationKey(destinationKey)
+            .uploadId(createMultipartUploadResponse.uploadId)
+            .partNumber(partNumber + 1)
+            .copySourceRange(s"bytes=$start-$lastByte")
```
dantb, line 97: This is fiddly, will S3 complain if the parts are not mutually consistent? I suppose the tests are good and pass it should be fine!

imsdu (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClientImpl.scala
olivergrabinski: You're right, but as long as we don't use the high-level api (TransferManager) then I'm failing to see a non-fiddly way, but it's possible I'm wrong

Issue #4943, opened 2024-05-06
Only support SHA-256 for S3 storages


imsdu:
**Motivation**
To make it easier to manage and operate the S3 bucket, only the SHA-256 algorithm should be supported

**Acceptance criteria**
1. Other algorithms should not be supported
2. The algorithm is removed from the configuration and the storage value
3. It is set a global variable for S3 storages

PR #4950, opened 2024-05-07
Make sha256 default for s3 storages

olivergrabinski:
#4943

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageValue.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageValue.scala
@@ -123,7 +123,7 @@ object StorageValue {
 
     override val tpe: StorageType = StorageType.S3Storage
 
-    override val algorithm: DigestAlgorithm  = DigestAlgorithm.default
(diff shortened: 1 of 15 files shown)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageValue.scala
```diff
@@ -116,11 +115,11 @@ object StorageValue {
       name: Option[String],
       description: Option[String],
       default: Boolean,
-      algorithm: DigestAlgorithm,
       bucket: String,
       readPermission: Permission,
       writePermission: Permission,
-      maxFileSize: Long
+      maxFileSize: Long,
+      algorithm: DigestAlgorithm = DigestAlgorithm.default
```
imsdu, line 122: I would even put it as a constant like the storage type

imsdu (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/model/StorageValue.scala
olivergrabinski: When I put it as a constant it didn't serialize correctly. I will look into it

PR #4951, opened 2024-05-07
Push files to a different location in import batch + add files suffix

### LINKED ISSUE #4890: Patch file paths when importing them
**Motivation**
With #4840, files are pushed in the same location as they were previously.
However the files can come from different types of storages with different types of path.

**Acceptance criteria**
1. The prefixes from files from local storage are replaced with the common S3 prefix
 * /opt/binaries/org/proj/... would become /new-prefix/org/proj/files
2. The prefixes from files from remote storage are replaced with the common S3 prefix
 * /gpfs/bbp.cscs.ch/data/project/projXX/nexus/org/proj would become /new-prefix/org/proj/files
3. Some old paths contains the project uuid and not `org/proj`, they should be replaced with the org and project handles
* /opt/binaries/project_uuid/... should become /new-prefix/org/proj/files
4. If a project is renamed in the target env, this should be reflected in the file path

**Not in scope**
1. Patching paths in distributions

### LINKED ISSUE #4925: Add a files suffix to file paths in S3
**Motivation**
In #4924, we will introduce file sets which will be a distinct entity, it would be valuable to easily make the difference between them at the project level in the S3 bucket.

Having a clear distinction between them will help to use tools like AWS S3 Inventory

**Acceptance criteria**
1. The file prefix was `/[prefix]/[org]/[project]/[uuid]/[filename]` and will now be `/[prefix]/[org]/[project]/files/[uuid]/[filename]`

imsdu:
Fixes #4890 
Fixes #4925

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3LocationGenerator.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3LocationGenerator.scala
@@ -1,0 +1,17 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.s3
+
+import akka.http.scaladsl.model.Uri
+import ch.epfl.bluebrain.nexus.delta.rdf.syntax.uriSyntax
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.ProjectRef
+
+import java.util.UUID
+
+class S3LocationGenerator(prefix: Uri) {
+
+  private def uuidToPath(uuid: UUID) =
+    Uri.Path(uuid.toString.toLowerCase.takeWhile(_ != '-').mkString("/"))
+
+  def file(project: ProjectRef, uuid: UUID, filename: String): Uri =
+    prefix / project.organization.toString / project.project.toString / "files" / uuidToPath(uuid) / filename
+
+}
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3StorageSaveFile.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3StorageSaveFile.scala
@@ -13,7 +13,6 @@ import ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model.FileAttributes.
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.files.model.{Digest, FileStorageMetadata}
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.DigestAlgorithm
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.model.Storage.S3Storage
-import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.FileOperations.intermediateFolders
 import ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.StorageFileRejection.SaveFileRejection._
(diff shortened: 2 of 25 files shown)


delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3LocationGenerator.scala
```diff
@@ -0,0 +1,17 @@
+package ch.epfl.bluebrain.nexus.delta.plugins.storage.storages.operations.s3
+
+import akka.http.scaladsl.model.Uri
+import ch.epfl.bluebrain.nexus.delta.rdf.syntax.uriSyntax
+import ch.epfl.bluebrain.nexus.delta.sourcing.model.ProjectRef
+
+import java.util.UUID
+
+class S3LocationGenerator(prefix: Uri) {
```
imsdu, line 9: So it is used for upload and the batch and also later for the delegate registration where we will have to generate a path

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClient.scala
```diff
@@ -49,15 +48,11 @@ trait S3StorageClient {
 
   def objectExists(bucket: String, key: String): IO[Boolean]
   def bucketExists(bucket: String): IO[Boolean]
-
-  def baseEndpoint: Uri
-
-  def prefix: Uri
```
imsdu, line 55: Pushed it to a higher level

ship/src/main/resources/ship-default.conf
```diff
@@ -163,7 +153,17 @@ ship {
...
       }
+    }
 
+    files {
+      # prepended to all file paths
+      #prefix = "/some-prefix"
+      import-bucket = ${ship.s-3.import-bucket}
+      # The bucket to which the files will be copied by the Nexus Ship
+      target-bucket = "nexus-delta-production"
+      enable-target-rewrite = false
```
imsdu, line 164: Temporary feature flag

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3StorageSaveFile.scala
```diff
@@ -33,49 +32,45 @@ final class S3StorageSaveFile(s3StorageClient: S3StorageClient)(implicit
...
     FileStorageMetadata(
       uuid = uuid,
       bytes = uploadMetadata.fileSize,
-      digest = Digest.ComputedDigest(algorithm, uploadMetadata.checksum),
+      digest = Digest.ComputedDigest(DigestAlgorithm.SHA256, uploadMetadata.checksum),
       origin = Client,
-      location = uploadMetadata.location,
-      path = Uri.Path(key)
+      location = location,
+      path = location.path
```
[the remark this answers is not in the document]
olivergrabinski: If it becomes the same then yeah it's not needed but an `https://...` URI could still be used by a client that does not want to mount the bucket but has access to it

olivergrabinski (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/S3StorageSaveFile.scala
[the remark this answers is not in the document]
imsdu: Well, he can fetch the bucket from the storage and with both he should be able to fetch the file via S3.

ship/src/main/resources/ship-default.conf
imsdu: Yes, probably to delete when the current import is finished and we start another one

PR #4954, opened 2024-05-10
Handle S3 access error gracefully with 403

### LINKED ISSUE #4887: Handle gracefully read-only access on buckets
**Motivation**
Delta can have access to different buckets where it can only have read-only access or write access to only a subset of it (ex: the sponsored bucket).

**Acceptance criteria**
1. On write operations, if those fail because of a lack of a permission, a 403 is returned with an appropriate message

dantb:
Fixes #4887 

To test this using localstack would require adding IAM to our container setup, then creating and setting up users with the desired permissions via the SDK. With all that setup, it might be inconsistent with the real S3 anyway. I thought it best to test manually instead by:
1. Creating user in sandbox with S3 read permissions
2. Pointing integration tests to sandbox S3 with access keys from my admin user
3. Pointing the dockerised delta to sandbox S3 with access keys from the restricted user
4. Run `S3StorageSpec`, see what file upload failures look like

In the end we now return 403 with
```json
   {
     "@type" : "BucketAccessDenied",
     "reason" : "File 'http://delta:8080/v1/resources/6k0mab3nc23xhan/ykab7aepbgphlt0/_/attachment.json' could not be saved using storage 'https://bluebrain.github.io/nexus/vocabulary/mys3storage'",
     "details" : "Access denied to bucket da7goncyy5irq8u at key myprefix/6k0mab3nc23xhan/ykab7aepbgphlt0/files/6/1/6/a/d/7/0/a/attachment.json"
   }
```

### DIFF
--- a/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileRejection.scala
+++ b/delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/files/model/FileRejection.scala
@@ -269,20 +269,21 @@ object FileRejection {
 
   implicit final val fileRejectionHttpResponseFields: HttpResponseFields[FileRejection] =
     HttpResponseFields.fromStatusAndHeaders {
-      case RevisionNotFound(_, _)                                          => (StatusCodes.NotFound, Seq.empty)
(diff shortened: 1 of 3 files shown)


shinyhappydan (approved):
Is forbidden the correct code to return here? The error is really a delta (or perhaps just the storage) setup issue rather than an issue with the request. I suppose since the story says 403 then it's correct

Issue #4941, opened 2024-05-06
Make sure that the filename is correctly encoded in the S3 bucket


imsdu:
**Motivation**
Filenames can sometimes create special characters which are url-encoded.
Akka urls .toString encode them by default and the S3 sdk may encode them a second time

**Acceptance criteria**
1. Make sure that filenames are not url-encoded twice

shinyhappydan:
In order to play this story I would need an example. What is the scenario where this could occur? What should the filename look like on S3?

For example, if the user supplies a filename which is `my file.txt`

What should this look like in nexus? 
- `my file.txt`
- `my%20file.txt` 

Similarly, what should this look like in S3? My current understanding is that only url-encoded strings are allowed, so our options are:
- `my%20file.txt` 
- `urlencode(my%20file.txt)`

It seems as if this ticket is saying that the first option is the correct one, but it's not clear. Perhaps as long as we tell the user the correct path of the file, it doesn't actually matter if the name is double encoded.

PR #4955, opened 2024-05-13
Add test to ensure filenames are encoded correctly in S3

shinyhappydan:
Covers #4941

### DIFF
--- a/tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/S3StorageSpec.scala
+++ b/tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/S3StorageSpec.scala
@@ -210,33 +210,42 @@ class S3StorageSpec extends StorageSpec {
 
   "Filenames with url-encodable characters" should {
     "have an appropriate filename in S3" in {
-      val id   = genId()
+
       val name = "name with spaces.txt"
 
-      for {
-        _ <- deltaClient.uploadFile[Json](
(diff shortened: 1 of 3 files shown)


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/files/S3StorageSpec.scala
```diff
@@ -206,6 +208,36 @@ class S3StorageSpec extends StorageSpec {
...
+      for {
+        _ <- deltaClient.uploadFile[Json](
+               s"/files/$projectRef/$id?storage=nxv:$storageId",
+               "file contents",
+               ContentTypes.`text/plain(UTF-8)`,
+               name,
+               Coyote
+             )((json, response) => {
+               response.status shouldEqual StatusCodes.Created
+               json should have(filenameField(name))
```
imsdu, line 225: Can you also check if the location field is ok ?
The `Uri` from akka could bring surprises

shinyhappydan: I eyeballed it and it's fine. Do you want that in the test also?

imsdu: Users use more location that the filename so if you can have an extra assertion to make sure that location == key, it would be great

olivergrabinski (approved):

PR #4983, opened 2024-05-16
Fix bug while importing files with special characters in name

imsdu:


dantb (approved):
Maybe accepting a new type in the client could enforce that paths are decoded correctly, so we don't have to remember at every call site (could also have convenience syntax)

imsdu:
> Maybe accepting a new type in the client could enforce that paths are decoded correctly, so we don't have to remember at every call site (could also have convenience syntax)

That is what I am thinking too, I will open a ticket about it

PR #4985, opened 2024-05-16
Filter out directories from import batch

### LINKED ISSUE #4962: Ignore which are in fact directories
**Motivation**
Some files are in fact directories, we plan to make the distinction in the future so we must ignore them know.
However, we may want to import them later when directories are introduced so we must stash them temporarily somewhere

**Acceptance criteria**
1. When a file event is in fact a directory, it is ignored

imsdu:
Fixes #4962


ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/files/FileProcessor.scala
```diff
@@ -63,56 +62,51 @@ class FileProcessor private (
...
+      case _: FileCancelledEvent                                 => IO.pure(ImportStatus.Dropped) // Not present in the export anyway
     }
-  }.redeemWith(
-    {
-      case a: ResourceAlreadyExists => logger.warn(a)("The resource already exists").as(ImportStatus.Dropped)
-      case i: IncorrectRev          => logger.warn(i)("An incorrect revision has been provided").as(ImportStatus.Dropped)
-      case f: FileNotFound          =>
-        // TODO: Remove this redemption when empty filenames are handled correctly
-        logger.warn(f)(s"The file ${f.id} in project ${f.project} does not exist.").as(ImportStatus.Dropped)
-      case n: NoSuchKeyException    =>
```
imsdu, line 104: This part is handled in the `FileCopier` now

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/files/FileCopier.scala
```diff
@@ -14,7 +14,7 @@ import scala.concurrent.duration.DurationInt
 
 trait FileCopier {
 
-  def copyFile(project: ProjectRef, attributes: FileAttributes): IO[Uri.Path]
+  def copyFile(project: ProjectRef, attributes: FileAttributes): IO[Option[Uri.Path]]
```
shinyhappydan, line 17: It's not very intuitive what the option represents. It would be clearer with an ADT

shinyhappydan (approved):

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/files/FileCopier.scala
imsdu: You are right, I will handle it in my next PR which also implies changes on this part

PR #4987, opened 2024-05-17
Setting content type when copying over the file in import batch

### LINKED ISSUE #4968: Set the content type when copying a file in S3
**Motivation**
When copying a file between 2 buckets, we should be able to provide an optional content type to be set on the new object

**Acceptance criteria**
1. The provided content type is set on the new object if provided
2. The original content type is kept otherwise

imsdu:
Fixes #4968 

Bonus: do not copy the file again if it exists at the given key


shinyhappydan (approved):

delta/plugins/storage/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/storage/storages/operations/s3/client/S3StorageClientImpl.scala
```diff
@@ -57,34 +58,56 @@ final private[client] class S3StorageClientImpl(client: S3AsyncClientOp[IO]) ext
...
+        val requestWithOptions = options.newContentType.fold(requestBuilder) { contentType =>
+          requestBuilder
+            .contentType(contentType.value)
+            .metadataDirective(MetadataDirective.REPLACE)
+        }
+        client.copyObject(requestWithOptions.build()).void
+      }
+    }
 
   def copyObjectMultiPart(
```
imsdu, line 82: No test so far here as Oliver is fixing it

PR #4993, opened 2024-05-22
Add slow query logger based on Doobie

imsdu:
Example of log:
```
2024-05-22 09:48:14 WARN  c.e.b.n.d.sourcing.QueryLogHandler - [StreamingPool] Slow Statement Execution:

 ((SELECT 'newState', type, id, org, project, value, instant, ordering, rev
  FROM public.scoped_states
  WHERE org = ? and project = ?  AND ordering > ?  AND tag = ? 
  ORDER BY ordering
  LIMIT ?)
  UNION ALL
  (SELECT 'tombstone', type, id, org, project, null, instant, ordering, -1
  FROM public.scoped_tombstones
  WHERE org = ? and project = ?  AND ordering > ?  AND tag = ?  AND (cause->>'deleted' = 'true' ) 
  ORDER BY ordering
  LIMIT ?)
  ORDER BY ordering)
  LIMIT ?

 arguments = [mbodytmoxbxk0do, n5rspxio3rlet3w, 184, latest, 30, mbodytmoxbxk0do, n5rspxio3rlet3w, 184, latest, 30, 30]
 label     = unlabeled
 elapsed = 11 ms exec + 0 ms processing (11 ms total)
```
So we have the arguments now to make it easier

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -912,7 +912,6 @@ lazy val kamonSettings = Seq(
     "io.kamon" %% "kamon-core"           % kamonVersion,
     "io.kamon" %% "kamon-executors"      % kamonVersion,
     "io.kamon" %% "kamon-jaeger"         % kamonVersion,
-    "io.kamon" %% "kamon-jdbc"           % kamonVersion,
     "io.kamon" %% "kamon-logback"        % kamonVersion,
     "io.kamon" %% "kamon-prometheus"     % kamonVersion,
     "io.kamon" %% "kamon-scala-future"   % kamonVersion,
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -912,7 +912,6 @@ lazy val kamonSettings = Seq(
     "io.kamon" %% "kamon-core"           % kamonVersion,
     "io.kamon" %% "kamon-executors"      % kamonVersion,
     "io.kamon" %% "kamon-jaeger"         % kamonVersion,
-    "io.kamon" %% "kamon-jdbc"           % kamonVersion,
```
imsdu, line 915: Slow query logging was the only feature we used from it so we can just get rid of it now

olivergrabinski (approved):

PR #5000, opened 2024-05-28
Do not update resource if the payload and the provided schema have not changed

### LINKED ISSUE #4970: Improve how diff is performed in the resource update
**Motivation**
When performing an update and there is no detected diff, a new event is still created when a schema id is provided.
We want to change this depending on how the schema is resolved

**Acceptance criteria**
1. If the schema is resolved to the same schema with the same or newer revision , no update event is created
2. If the schema is resolved to another schema, a updated schema event is created
3. If a tag is provided, the tag is always created (it could be tied to a different event though)
4. There is no change of behaviour when a change is detected on the resource

imsdu:
Fixes #4970

### DIFF
--- a/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/model/ResourceEvent.scala
+++ b/delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/model/ResourceEvent.scala
@@ -205,6 +205,26 @@ object ResourceEvent {
 
   }
 
+  /**
+    * Event representing a change of schema for a resource
+    *
+    * @param id
+    *   the resource identifier
+    * @param project
+    *   the project where the resource belongs
+    * @param schema
+    *   the schema used to constrain the resource
+    * @param schemaProject
+    *   the project where the schema belongs
+    * @param types
+    *   the collection of known resource types
+    * @param rev
+    *   the resource revision
+    * @param instant
+    *   the instant when this event was created
+    * @param subject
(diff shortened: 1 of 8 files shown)


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/resources/model/ResourceEvent.scala
```diff
@@ -213,7 +233,8 @@ object ResourceEvent {
       types: Set[Iri],
       rev: Int,
       instant: Instant,
-      subject: Subject
+      subject: Subject,
+      tag: Option[UserTag]
```
imsdu, line 237: To handle the option tag when an update attempt fallback to a update schema event

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/ResourcesSpec.scala
```diff
@@ -373,19 +372,11 @@ class ResourcesSpec extends BaseIntegrationSpec {
...
 
     "succeed" in {
-      for {
-        _ <-
-          deltaClient.put[Json](s"/resources/$project1/test-schema/test-resource:1?rev=1", payload, Rick) {
-            (json, response) =>
-              response.status shouldEqual StatusCodes.OK
-              _rev.getOption(json).value shouldEqual 2
-          }
-        // Sending the same update should not create a new revision
```
imsdu, line 383: Tested in the other suite

PR #5009, opened 2024-06-04
Always rewrite file paths in import batch

imsdu:


ship/src/test/scala/ch/epfl/bluebrain/nexus/ship/S3RunShipSuite.scala
```diff
@@ -36,33 +34,6 @@ class S3RunShipSuite
     } yield ()
   }
 
-  test("Run import with file events") {
```
imsdu, line 39: Tested in another test suite

olivergrabinski (approved):

PR #5018, opened 2024-06-05
Use the write db pool when fetching project context for write operations

### LINKED ISSUE #5016: Fetching the project context during write operations should be done on the write pool
**Describe the bug**
When we create a project, we automatically create some resources (storage/resolver/view)
However when Nexus is using a PostgreSQL cluster, there may be some replication delays and fetching the context of the new project can fail and the creation of those default resources can't be completed.

**Expected behavior**
When fetching the context for a write operation, this should be done from the primary node

imsdu:
Fixes #5016


delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/FetchContext.scala
```diff
@@ -53,42 +53,51 @@ abstract class FetchContext { self =>
...
+      fetchProject: (ProjectRef, Boolean) => IO[Option[ProjectResource]],
       quotas: Quotas
   ): FetchContext =
     new FetchContext {
 
       override def defaultApiMappings: ApiMappings = dam
 
       override def onRead(ref: ProjectRef): IO[ProjectContext] =
-        fetchProject(ref).flatMap {
+        fetchProject(ref, false).flatMap {
```
dantb, line 93: Named parameter may be a bit clearer for the flag. Same below

dantb (approved):

delta/sdk/src/main/scala/ch/epfl/bluebrain/nexus/delta/sdk/projects/FetchContext.scala
imsdu: We can't use them here with a function passed as a parameter

PR #5031, opened 2024-06-19
Add purge routine for tombstones

### LINKED ISSUE #4986: Add a routine to purge state tombstones
**Describe the bug**
State tombstones allows for indexing to delete resources from a view if their type changed and did not fit the type restriction on the view anymore.

This table is meant to be purged after the tombstone reaches a given ttl.
The routine responsible for purging the tombstone is missing and needs to be implemented as the table is growing and the indexing query is getting slower where 

**Expected behavior**
* Tombstones should be purged when they reached the ttl

This should be handled in the same way as how we purge views restart requests for example

imsdu:
Fixes #4986 

As it is another purge routine added to the ones we already had, it includes a refactoring to align them and make it easier to declare them to the supervisor


delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/stream/PurgeProjectionCoordinator.scala
```diff
@@ -0,0 +1,37 @@
+package ch.epfl.bluebrain.nexus.delta.sourcing.stream
+
+import cats.syntax.all._
+import cats.effect.{Clock, IO}
+import ch.epfl.bluebrain.nexus.delta.sourcing.config.PurgeConfig
+import fs2.Stream
+import java.time.Instant
+
+sealed trait PurgeProjectionCoordinator
```
imsdu, line 9: Same pattern as for views

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/DeleteExpired.scala
```diff
@@ -32,18 +34,12 @@ object DeleteExpired {
...
-
-    val stream = Stream
-      .awakeEvery[IO](config.deleteExpiredEvery)
-      .evalTap(_ => deleteExpired())
-      .drain
-
-    val deleteExpiredProjection =
-      CompiledProjection.fromStream(metadata, ExecutionStrategy.TransientSingleNode, _ => stream)
-    supervisor.run(deleteExpiredProjection).as(deleteExpired)
+  def apply(deleteExpiredEvery: FiniteDuration, xas: Transactors): PurgeProjection = {
```
imsdu, line 39: We just have to define a purge projection now without having to care about the supervisor part

delta/sourcing-psql/src/main/scala/ch/epfl/bluebrain/nexus/delta/sourcing/PurgeElemFailures.scala
```diff
@@ -1,29 +1,28 @@
...
 
-final class PurgeElemFailures private[sourcing] (xas: Transactors, ttl: FiniteDuration, clock: Clock[IO]) {
+final class PurgeElemFailures private[sourcing] (xas: Transactors) {
 
   /**
-    * Deletes the projection errors that are older than the given `ttl`.
+    * Deletes the projection errors that are older than the given instant.
     */
-  def apply(): IO[Unit] =
+  def apply(instant: Instant): IO[Unit] =
```
imsdu, line 19: The threshold instant is already computed from the info from the `PurgeProjection`

olivergrabinski (approved):
👍

PR #5036, opened 2024-06-21
Update Elasticsearch and Localstack versions for integration tests

imsdu:


docs/src/main/paradox/docs/getting-started/running-nexus/docker/docker-compose.yaml
```diff
@@ -1,11 +1,10 @@
-version: "3.3"
```
imsdu, line 1: obsolete with the latest version of Docker

olivergrabinski (approved):

PR #5046, opened 2024-06-28
Prohibit from using Nexus vocab types in generic resources

### LINKED ISSUE #5043: Prevent a generic resource  to be created with a type from the Nexus vocabulary
**Motivation**
The resource endpoint is very permissive as it can accept almost any valid json-ld payload to create a resource.
This can cause some problems as someone can create what should be a schema or a view with this endpoint: the resource then has the correct json-ld type but is still considered a generic resource and extra validation is not applied.

Ex:
* Import resolution for schemas
* Creation of the index and validation of the mapping/settings for Elasticsearch views

**Acceptance criteria**
1. When attempting to create a generic resource when one of the types belongs to the Nexus vocabulary (so a type starting with https://bluebrain.github.io/nexus/vocabulary/), an error is returned

**Implementation notes**
This error must be translated as an Akka rejection in `ResourcesRoutes`

imsdu:
Fixes #5043

### DIFF
--- a/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
+++ b/delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
@@ -59,6 +59,10 @@ final class ResourcesRoutes(
   implicit private def resourceFAJsonLdEncoder[A: JsonLdEncoder]: JsonLdEncoder[ResourceF[A]] =
     ResourceF.resourceFAJsonLdEncoder(ContextValue.empty)
 
+  private val rejectPredicateOnWrite: PartialFunction[ResourceRejection, Boolean] = {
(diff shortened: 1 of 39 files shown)


delta/app/src/main/scala/ch/epfl/bluebrain/nexus/delta/routes/ResourcesRoutes.scala
```diff
@@ -93,7 +98,7 @@ final class ResourcesRoutes(
                             .flatTap(index(project, _, mode))
                             .map(_.void)
                             .attemptNarrow[ResourceRejection]
-                            .rejectOn[InvalidSchemaRejection]
+                            .rejectWhen(rejectPredicateOnWrite)
```
shinyhappydan, line 101: "reject when reject predicate on write" doesn't communicate what is happening here

shinyhappydan (approved):

PR #5047, opened 2024-06-28
Make projects fields a list in suite json-ld context

imsdu:


delta/plugins/search/src/test/scala/ch/epfl/bluebrain/nexus/delta/plugins/search/SearchRoutesSpec.scala
```diff
@@ -32,10 +32,11 @@ class SearchRoutesSpec extends BaseRouteSpec {
 
   private val fields = Json.obj("fields" := true)
 
-  private val publicProjects = Set(ProjectRef.unsafe("org", "project"), ProjectRef.unsafe("org2", "project2"))
-  private val suites         = Map(
+  private val publicProjects  = Set(ProjectRef.unsafe("org", "project"), ProjectRef.unsafe("org2", "project2"))
+  private val privateProjects = Set(ProjectRef.unsafe("org3", "project3"))
```
Dinika, line 36: Nitipicking - The name `privateProjects` is a bit confusing to me because in theory we could have a suite of one public project also (even though it doesn't make much sense) and the endpoint would return the same response as it would if the project was private.

Dinika (approved):
Thanks for the quick fix :slightly_smiling_face: 
I've added a minor comment that can be ignored.

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigIndexingSpec.scala
```diff
@@ -1097,7 +1097,7 @@ class SearchConfigIndexingSpec extends BaseIntegrationSpec {
                 }"""
 
     assertOneSource(queryField(singleNeuronSimulationId, "singleNeuronSimulation")) { json =>
-      json shouldEqual expected
+      json should equalIgnoreArrayOrder(expected)
```
imsdu, line 1100: Fixing the flaky test as recordingLocation elements can have different orders

PR #5064, opened 2024-07-10
Patch properly distribution including rev or tag

imsdu:


ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/resources/DistributionPatcher.scala
```diff
@@ -47,8 +48,7 @@ final class DistributionPatcher(
   private def modificationsForFile(originalProject: ProjectRef, resourceRef: ResourceRef): IO[Json => Json] = {
     val targetProject                                = projectMapper.map(originalProject)
     val patchedResourceRef                           = iriPatcher(resourceRef)
-    val newContentUrl                                =
-      ResourceUris("files", targetProject, patchedResourceRef.original).accessUri(targetBase).toString()
+    val newContentUrl                                = createContentUrl(targetProject, patchedResourceRef)
```
shinyhappydan, line 51: If `ResourceUris` accepted a `ResourceRef`, it would prevent this mistake in the future. It feels like if you have a `ResourceRef` you would always want this behaviour

shinyhappydan (approved):

ship/src/main/scala/ch/epfl/bluebrain/nexus/ship/resources/DistributionPatcher.scala
imsdu: The thing is that with `ResourceUris`, it is possible to build other uris than the `accessUri` one.
And some of them would not be relevant for older revisions like incoming and outgoing.
It could maybe achieved with a refactoring but it would be outside the scope of this bug fix.

PR #5099, opened 2024-08-13
Add analysisSuitable annotation for search index

crisely09:
Same idea as the previous PR, just one more annotation.


imsdu (approved):

crisely09:
It should work now.

PR #5100, opened 2024-08-13
fix SynaptomeSimulation type

bilalesi:
fix `SingleNeuronSynaptomeSimulation` to `SynaptomeSimulation` type to witelist of types for search

### DIFF
--- a/tests/docker/config/resource-types.json
+++ b/tests/docker/config/resource-types.json
@@ -55,7 +55,7 @@
   "https://neuroshapes.org/SingleCellSimulationTrace",
   "https://bbp.epfl.ch/ontologies/core/bmo/SingleNeuronSimulation",
   "https://bbp.epfl.ch/ontologies/core/bmo/SingleNeuronSynaptome",
-  "https://bbp.epfl.ch/ontologies/core/bmo/SingleNeuronSynaptomeSimulation",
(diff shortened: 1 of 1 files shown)


crisely09:
I am not sure I agree with this change, are we sure there is no other SynaptomeSimulation that is not Single Neuron?

imsdu (approved):

bilalesi:
> I am not sure I agree with this change, are we sure there is no other SynaptomeSimulation that is not Single Neuron?

I totally agree and that was my first impression, but I changed after I check the schema in this [MR ](https://bbpgitlab.epfl.ch/dke/apps/brain-modeling-ontology/-/merge_requests/319/diffs#2ade5113e5aaeb2b1722c2312e3a4e25ad245994_0_4)

so may be the schema need to be updated

crisely09:
> > I am not sure I agree with this change, are we sure there is no other SynaptomeSimulation that is not Single Neuron?
> 
> I totally agree and that was my first impression, but I changed after I check the schema in this [MR ](https://bbpgitlab.epfl.ch/dke/apps/brain-modeling-ontology/-/merge_requests/319/diffs#2ade5113e5aaeb2b1722c2312e3a4e25ad245994_0_4)
> 
> so may be the schema need to be updated

I see, yes, I think here I am suggesting using the objectOfStudy field to specify this is single cell, OK.

crisely09 (approved):

PR #5115, opened 2024-08-22
Make simulationReady property a construction from a quality metric annotation

crisely09:
The goal here is to avoid the need of updating manually the annotations inside the morphology, and instead use the existing quality measurement annotations to construct this field when indexing.


crisely09:
OK, seems like it's good now.

imsdu (approved):

PR #5119, opened 2024-08-23
Add synaptome properties, examples and tests

crisely09:
I added the properties requested, just be aware that they are indexed with different names:
    SingleNeuronSynaptome:
        used -> .singleNeuronSynaptome.memodel, 
        seed -> .singleNeuronSynaptome.seed
    SynaptomeSimulation:
        used -> synaptomeSimulation.synaptome
        injectionLocation -> synaptomeSimulation.injectionLocation
        recordingLocation  -> synaptomeSimulation.recordingLocation

### DIFF
--- a/tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigIndexingSpec.scala
+++ b/tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigIndexingSpec.scala
@@ -1144,8 +1144,8 @@ class SearchConfigIndexingSpec extends BaseIntegrationSpec {
       json"""{
                  "synaptomeSimulation": {
                     "injectionLocation": "soma[0]",
-                    "recordingLocation": ["soma[0]_0.5", "dend[38]_0.5", "dend[1]_0.4", "apic[53]_0.3"dendrite.01A", "dendrite.01B"],
-                    "synaptome": { "@id" : "https://bbp.epfl.ch/data/synapse/single-neuron-synaptome", "name" : "syanptome-model-04"}
+                    "recordingLocation": ["soma[0]_0.5", "dend[38]_0.5", "dend[1]_0.4", "apic[53]_0.3", "dendrite.01A", "dendrite.01B"],
+                    "synaptome": { "@id" : "https://bbp.epfl.ch/data/synapse/single-neuron-synaptome", "name" : "synaptome-model-04"}
                   }
                 }"""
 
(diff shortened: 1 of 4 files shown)


tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigIndexingSpec.scala
```diff
@@ -1121,6 +1125,35 @@ class SearchConfigIndexingSpec extends BaseIntegrationSpec {
...
+
+  "have the correct synaptome simulation information" in {
+    val expected =
+      json"""{
+                 "synaptomeSimulation": {
+                    "injectionLocation": "soma[0]",
+                    "recordingLocation": ["soma[0]_0.5", "dend[38]_0.5", "dend[1]_0.4", "apic[53]_0.3"],
+                    "synaptome": { "@id" : "https://bbp.epfl.ch/data/synapse/single-neuron-synaptome", "name" : "synaptome-model-04"}
+                  }
+                }"""
```
bilalesi, line 1150: so basically request to the index endpoint will return this format not "used" property along with different other generic properties?
if so, the response resource already of type `SynaptomeSimulation` , it seems that will have duplicated  context here, no ?

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigIndexingSpec.scala
```diff
@@ -1121,6 +1125,35 @@ class SearchConfigIndexingSpec extends BaseIntegrationSpec {
...
   }
 
+  "have the correct single neuron synaptome information" in {
+    val expected =
+      json"""{
+                 "singleNeuronSynaptome": {
+                    "seed": 100,
+                    "memodel": { "@id" : "https://bbp.epfl.ch/data/memodel", "name" : "me-model with status"}
+                  }
+                }"""
```
bilalesi, line 1135: same as next comment

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigIndexingSpec.scala
crisely09: Yes, the issue is that without "binding" these properties to the "synaptomeSimulation" id, as other resources of other types may have also the `used` property, then instead of getting one single result you get a list of all the triples found. This is why you can see in the same query we do this trick for emodels, memodels, etc.

bilalesi:
LGTM 👍

bilalesi (approved):

crisely09:
For the merging on gitlab, usually @imsdu does that part

PR #5132, opened 2024-09-09
Edit constraint on trace's encoding format

ssssarah:
The current filter will filter out any ExperimentalTrace that has one of its distribution of encoding format something other than `application/nwb`.
In other words, if an ExperimentalTrace has multiple distributions, one being `application/nwb` and another being `application/abf`, the ExperimentalTrace would be filtered out, which is not what we want. The goal is to ensure there is at least one distribution of encoding format `application/nwb`


crisely09 (approved):
Looks good to me.

PR #5237, opened 2024-11-19
Added simulation properties isDraft and status to search index

crisely09:
Properties to be queried for SynaptomeSimulation and SingleNeuronSimulation types.


tests/docker/config/search/construct-query.sparql
```diff
@@ -242,6 +242,8 @@ CONSTRUCT {
   ?alias :singleNeuronSimulation  ?singleNeuronSimulation  .
                                   ?singleNeuronSimulation  :injectionLocation                ?injectionLocation .
                                   ?singleNeuronSimulation  :recordingLocation                ?recordingLocation .
+                                  ?singleNeuronSimulation  :isDraft                          ?isDraft .
+                                  ?singleNeuronSimulation  :status                           ?status  .
```
imsdu, line 246: `status` is already used by simulation campaigns, are we sure they are not clashing each other ?

crisely09: are they indexed together? how is the batch constructed?

crisely09: is there a way to test this?

imsdu: They can as the batch takes the state log in its chronological order

crisely09: then isn't it the same for `Simulation` VS `SimulationCampaign`, they both use status

imsdu (approved):

PR #5249, opened 2024-11-25
Changing the patching strategy for contentUrl, atLocation.location and url

imsdu:
When `contentUrl` matches the defined prefix
* Strip the prefix from its value
* Apply the patched value to `contentUrl`, `atLocation.location` and `url`

If not. fallback to `atLocation.location`:
* Strip the prefix from its value
* Apply the patched value to `atLocation.location` and `url`


crisely09 (approved):
looks good

Issue #5102, opened 2024-08-13
When entries in the context are substrings of one another, wrong match for URIs


ssssarah:
**Describe the bug**

I am unsure how to describe this in general terms but take the example of the following resource: 
https://bbp.epfl.ch/nexus/web/bbp/ionchannel/resources/https:%2F%2Fbbp.epfl.ch%2Fneurosciencegraph%2Fdata%2F000119dd-37fd-45ae-add9-9e5ac85a81a7

The last edit I made to it was on distribution.contentUrl. 

- However, Fusion adds into its diff that some stimulus types were changed, but they weren't. 
The URI of the stimulus type has an entry in the context, but there is an entry before it in the context that is a substring of it:
Fusion believes that "http://bbp.epfl.ch/neurosciencegraph/ontologies/stimulustypes/Activation" is A:ctivation because it matches with the "http://bbp.epfl.ch/neurosciencegraph/ontologies/stimulustypes/A" entry in the context.

- Additionally if one tries to traverse the link it generates an error: 
Resource 'A%3Activation' not found

- Finally, this is also happening when fetching this resource without the original payload. 
GET https://bbp.epfl.ch/nexus/v1/resources/bbp/ionchannel/_/https:%2F%2Fbbp.epfl.ch%2Fneurosciencegraph%2Fdata%2F000119dd-37fd-45ae-add9-9e5ac85a81a7

**What part of Nexus is affected**
- [x] Fusion
- [x] Delta

PR #5280, opened 2025-02-17
Update to Jena 5.2 and Json-LD 1.1

imsdu:


delta/app/src/main/resources/app.conf
```diff
@@ -116,23 +116,6 @@ app {
     resolve-base = "http://localhost:8081"
   }
 
-  # Json-LD Api configuration
-  json-ld-api {
```
imsdu, line 120: The ability to have to configure the severity of the parser was introduced a while back to avoid problems down the line with Blazegraph for example because some `@id` would be invalid.

Since then it was always used as strict and we never had problems so we can just get rid of it

delta/app/src/test/resources/schemas/errors/invalid-schema.json
```diff
@@ -11,15 +11,15 @@
     "result": {
       "@type": "sh:ValidationResult",
       "focusNode": "nxv:MyShape",
-      "result:Path": {
```
imsdu, line 14: This was a bug from the previous Json-LD library.
Same as #5102

delta/plugins/blazegraph/src/main/scala/ch/epfl/bluebrain/nexus/delta/plugins/blazegraph/model/SparqlLink.scala
```diff
@@ -71,21 +71,21 @@ object SparqlLink {
       for {
         link         <- SparqlExternalLink(bindings)
         project      <-
-          bindings.get(nxv.project.prefix).map(_.value).flatMap(Iri.absolute(_).toOption).flatMap(projectRefFromId)
```
imsdu, line 74: This was renamed to follow a naming change in Jena which now follows a different RFC for IRIs.

delta/plugins/storage/src/main/resources/contexts/files.json
```diff
@@ -13,16 +12,24 @@
...
+        "@base": "https://bluebrain.github.io/nexus/vocabulary/"
+      },
       "@id": "https://bluebrain.github.io/nexus/vocabulary/origin",
       "@type": "@id"
     },
     "_algorithm": "https://bluebrain.github.io/nexus/vocabulary/algorithm",
     "_attributes": "https://bluebrain.github.io/nexus/vocabulary/attributes",
-    "_keywords": "https://bluebrain.github.io/nexus/vocabulary/keywords",
+    "_keywords": {
+      "@context": {
```
imsdu, line 24: Use  a scoped context

delta/rdf/src/main/scala/io/circe/jakartajson/package.scala
```diff
@@ -0,0 +1,95 @@
...
+import jakarta.json.{Json => JakartaJson, JsonNumber => JakartaJsonNumber, JsonObject => JakartaJsonObject, JsonString, JsonValue => JakartaJsonValue}
+
+import scala.jdk.CollectionConverters._
+import java.math.{BigDecimal => JBigDecimal}
+
+/**
+  * Support for converting between Jakarta Json and Circe Based on
+  * https://github.com/circe/circe-jackson/blob/master/shared/src/main/scala/io/circe/jackson/package.scala
+  */
+package object jakartajson {
```
imsdu, line 14: Nexus and Titanium use different json libraries so this is needed to do a conversion between those.
Before that Nexus was just getting back to a raw string and parse things again with the target library. This had a performance overhead which is now eliminated.

delta/rdf/src/test/resources/context.json
```diff
@@ -4,9 +4,7 @@
     "@vocab": "http://example.com/",
     "@base": "http://nexus.example.com/",
     "xsd": "http://www.w3.org/2001/XMLSchema#",
-    "xsd2": "http://www.w3.org/2001/XMLSchema#",
     "Person": "http://schema.org/Person",
-    "Person2": "http://schema.org/Person",
```
imsdu, line 9: I still need to check that again but it does not take the same prefix when 2 prefixes point to the same target.
It should not happen in real life though.

delta/rdf/src/test/scala/ch/epfl/bluebrain/nexus/delta/rdf/jsonld/ExpandedJsonLdSpec.scala
```diff
@@ -153,7 +153,7 @@ class ExpandedJsonLdSpec extends CatsEffectSpec with Fixtures with GraphHelpers
 
       val input =
         json"""{"@context": ["http://localhost/c", {"a": "http://localhost/a"} ], "a": "A", "c": "C", "d": "D"}"""
-      ExpandedJsonLd(input).rejected shouldEqual RemoteContextCircularDependency(iri"http://localhost/c")
+      ExpandedJsonLd(input).rejectedWith[ConversionError].getMessage contains "Too many contexts"
```
imsdu, line 156: With Json-LD 1.1 and imported contexts, the circular dependency check is replaced by a maximum number of loaded contexts

tests/docker/config/search/search-context.json
```diff
@@ -46,15 +46,6 @@
   "project": {
     "@type": "@id"
   },
-  "valueX": {
```
imsdu, line 49: For some reason, putting the type meant that the property was not compacted anymore.
The Json-LD playground has the same behaviour.

tests/src/test/resources/kg/context/bbp-neuroshapes.json
```diff
@@ -2,7 +2,6 @@
   "@context": [
     "https://neuroshapes.org",
     {
-      "@base": "https://bbp.epfl.ch/neurosciencegraph/data/",
```
imsdu, line 5: When setting a base, it attempts to compact the ids relatively to this value by default.
It was enabled in the previous code but did not work for some reason.

tests/src/test/resources/kg/context/neuroshapes.json
```diff
@@ -4329,7 +4329,7 @@
     "uHillCoefficient": {
       "@id": "https://bbp.epfl.ch/ontologies/core/parameters/uHillCoefficient"
     },
-    "uberon": "http://purl.obolibrary.org/obo/UBERON_",
```
imsdu, line 4332: targets ending this way do not work anymore.
Consistent with the Json-Ld playground

tests/src/test/resources/kg/files/list.json
```diff
@@ -19,6 +19,7 @@
       "_filename": "attachment.json",
       "_incoming": "{{fileSelfPrefix}}attachment.json/incoming",
       "_mediaType": "application/json",
+      "_origin" : "Client",
```
imsdu, line 22: Bug from the previous library which is now fixed

tests/src/test/resources/kg/search/data/bouton-density.json
```diff
@@ -10,7 +10,7 @@
   "annotation": {
     "@type": [
       "Annotation",
-      "MType:Annotation"
+      "MTypeAnnotation"
```
imsdu, line 13: Same

tests/src/test/resources/kg/search/data/neuron-morphology.json
```diff
@@ -17,7 +17,7 @@
         "@id": "http://uri.interlex.org/base/ilx_0383207",
         "@type": [
           "MType",
-          "Annotation:Body"
+          "AnnotationBody"
```
imsdu, line 20: Same

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigIndexingSpec.scala
```diff
@@ -572,10 +572,10 @@ class SearchConfigIndexingSpec extends BaseIntegrationSpec {
         json"""
         {
           "neuronDensity" : {
-            "label" : "35200.0 neurons/mm³ (N = 5)",
+            "label" : "35200 neurons/mm³ (N = 5)",
             "nValue" : 5,
             "unit" : "neurons/mm³",
-            "value" : 35200.0
+            "value" : 35200
```
imsdu, line 578: Conforms to the spec if the decimal value is 0, it gets dropped.

tests/src/test/scala/ch/epfl/bluebrain/nexus/tests/kg/SearchConfigIndexingSpec.scala
```diff
@@ -1039,7 +1039,7 @@ class SearchConfigIndexingSpec extends BaseIntegrationSpec {
       val expectedFeature =
         json"""
                {
-                 "value" : 0.21388888888888888,
+                 "value" : 0.2138888888888889,
```
imsdu, line 1042: Rounding is different now.
It is consistent with the Json-LD playground.

tests/src/test/resources/kg/context/neuroshapes.json
crisely09: you mean in underscore?

imsdu: Without a `/` or a `#`

